Issue #14, opened 2026-04-20
[bug]: Kiro CLI tracking not showing up


thabti:
### What happened?

KIRO CLI token usage doesn't appear, would be great to have this. 

### Steps to reproduce

n/a

### TokenTracker version

0.5.72

### Node.js version

v25.9.0

### How did you install it?

macOS DMG (TokenTrackerBar)

### OS / version

Mac OS

### Output of `tokentracker doctor` (if applicable)

```shell

```

mm7894215:
Thanks for the report!

TokenTracker currently supports the **Kiro IDE** — token usage is collected by reading `~/Library/Application Support/Kiro/User/globalStorage/kiro.kiroagent/dev_data/` (`devdata.sqlite` + `tokens_generated.jsonl`). So if you're using Kiro Agent inside the Kiro IDE (the VS Code fork), usage should show up.

However, **Kiro CLI** is a separate tool with a different storage location and format, and it's **not yet supported**.

We've added Kiro CLI to our integration roadmap and will ship support in a future release (similar approach to our Kimi / Codex integrations — parser + sync hook + detect logic). It'll be called out in the CHANGELOG when it lands.

If you can share where Kiro CLI stores its local data (e.g. `ls ~/.kiro/` or wherever the CLI writes its session/usage files) and which file contains the token counts, that would help us implement support faster.

Leaving this issue open to track — will close it once Kiro CLI support ships.

---
_Generated by [Claude Code](https://claude.ai/code)_

mm7894215:
Fixed and shipping in **v0.5.73** — big thanks to @CiprianSpiridon who built the integration in #16.

### How it works
- Reads `~/Library/Application Support/kiro-cli/data.sqlite3` for finalized conversations **and** live session files under `~/.kiro/sessions/cli/*.json` so you see your current session without having to quit `kiro` first.
- Emits `source='kiro'` in the queue — Kiro CLI usage merges with Kiro IDE at the token, cost, leaderboard, and heatmap levels.
- Bedrock ARNs (`anthropic.claude-sonnet-4-20250514-v1:0` …) canonicalize to the short name (`claude-sonnet-4`) so CLI/IDE rows collapse onto one model.
- `model_id: "auto"` (Kiro's auto-routing) falls back to `kiro-cli-agent` sonnet-tier pricing.

### Caveat
[...]

PR #16, opened 2026-04-20
feat: add Kiro CLI token tracking with SQLite-backed approximation

CiprianSpiridon:
Reads historical conversations from ~/Library/Application Support/kiro-cli/data.sqlite3 (table conversations_v2) and emits queue entries under source='kiro' so CLI and IDE usage merge at the token, cost, and leaderboard levels without any downstream code change on the token side. Kiro CLI does not persist explicit token counts; they are approximated at 4 chars/token from user_prompt_length and response_size. Model IDs (Bedrock ARNs like anthropic.claude-sonnet-4-20250514-v1:0) are canonicalized so IDE and CLI rows collapse when they refer to the same underlying model. Per-request cursor state (cursors.kiroCli.requests) captures each request's fingerprint and contribution so mutations (same request_id, changed prompt/response size) subtract-old/add-new instead of being skipped forever. Pricing tables in src/lib/local-api.js and dashboard/edge-patches/tokentracker-leaderboard-refresh.ts get kiro-agent and kiro-cli-agent entries plus a fuzzy kiro-* rule so cost no longer silently rounds to zero.

# PR Goal (one sentence)

## Scope

-

- **Delta since last Codex review:** (commits or summary)
- **Edge cases covered:**
- **Tests run (command + result):**

### Rules / Invariants

-

-

### Evidence (tests or repro)

-

- [ ] Mark N/A if no public exposure

-

- [ ]

### Uncovered scope

-

### DIFF
--- a/dashboard/edge-patches/tokentracker-leaderboard-refresh.ts
+++ b/dashboard/edge-patches/tokentracker-leaderboard-refresh.ts
@@ -74,6 +74,11 @@ const MODEL_PRICING: Record<string, { input: number; output: number; cache_read:
   "kimi-for-coding": { input: 0.6, output: 2, cache_read: 0.15 },
   "kimi-k2.5": { input: 0.6, output: 2, cache_read: 0.15 },
   "kimi-k2.5-free": { input: 0, output: 0, cache_read: 0 },
+  // ── AWS Kiro (mirrored byte-for-byte from src/lib/local-api.js to
+  //    prevent cloud/local cost drift — Kiro routes through Bedrock,
+  //    most commonly claude-sonnet-4). ──
+  "kiro-agent": { input: 3, output: 15, cache_read: 0.3, cache_write: 3.75 },
+  "kiro-cli-agent": { input: 3, output: 15, cache_read: 0.3, cache_write: 3.75 },
   // ── Misc / Free ──
   "glm-4.7-free": { input: 0, output: 0, cache_read: 0 },
   "nemotron-3-super-free": { input: 0, output: 0, cache_read: 0 },
@@ -99,6 +104,7 @@ function getModelPricing(model: string) {
   if (lower.includes("gemini-3")) return MODEL_PRICING["gemini-3-flash-preview"];
   if (lower.includes("gemini-2.5")) return MODEL_PRICING["gemini-2.5-pro"];
   if (lower.includes("kimi")) return MODEL_PRICING["kimi-k2.5"];
+  if (lower.includes("kiro")) return MODEL_PRICING["kiro-cli-agent"];
   if (lower.includes("composer")) return MODEL_PRICING["composer-1"];
   if (lower === "auto") return MODEL_PRICING["composer-1"];
   return ZERO_PRICING;
--- a/src/commands/status.js
+++ b/src/commands/status.js
@@ -20,7 +20,7 @@ const { collectTrackerDiagnostics } = require("../lib/diagnostics");
 const { probeOpenclawHookState } = require("../lib/openclaw-hook");
 const { probeOpenclawSessionPluginState } = require("../lib/openclaw-session-plugin");
 const { resolveTrackerPaths } = require("../lib/tracker-paths");
-const { resolveKimiWireFiles } = require("../lib/rollout");
+const { resolveKimiWireFiles, resolveKiroCliDbPath } = require("../lib/rollout");
 
 async function cmdStatus(argv = []) {
   const opts = parseArgs(argv);
@@ -118,6 +118,13 @@ async function cmdStatus(argv = []) {
   const kimiHome = process.env.KIMI_HOME || path.join(home, ".kimi");
   const kimiInstalled = fssync.existsSync(path.join(kimiHome, "sessions"));
 
+  // Kiro CLI — reads from SQLite at
+  // ~/Library/Application Support/kiro-cli/data.sqlite3. End-user dashboards
+  // show CLI and IDE merged under a single "Kiro" brand; this status line
+  // surfaces the CLI sub-path separately for operators.
+  const kiroCliDbPath = resolveKiroCliDbPath(process.env);
+  const kiroCliInstalled = fssync.existsSync(kiroCliDbPath);
+
   const copilotToken = readCopilotOauthToken({ home });
   const copilotOtel = describeCopilotOtelStatus({ home, env: process.env });
   const copilotLines = formatCopilotLines({ token: copilotToken, otel: copilotOtel });
@@ -146,6 +153,9 @@ async function cmdStatus(argv = []) {
       `- OpenClaw hook (legacy): ${openclawHookState?.configured ? "set" : "unset"}`,
       kimiInstalled
         ? `- Kimi Code: passive reader (${kimiWireFiles.length} wire.jsonl file${kimiWireFiles.length !== 1 ? "s" : ""} found)`
+        : null,
+      kiroCliInstalled
+        ? `- Kiro CLI: SQLite data.sqlite3 found (tokens approximated from char lengths, merged under 'kiro' source)`
         : null,
       ...copilotLines,
       ...subscriptionLines,
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -27,6 +27,9 @@ const {
   parseCopilotIncremental,
   resolveKimiWireFiles,
   parseKimiIncremental,
+  resolveKiroCliSessionFiles,
+  resolveKiroCliDbPath,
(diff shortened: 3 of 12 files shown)


thabti:
closes #14

thabti (approved):
LGTM!

mm7894215 (changes requested):
Thanks for the thorough work here! The overall approach — merging Kiro CLI under the same `source: "kiro"` as Kiro IDE so they aggregate transparently at token / cost / leaderboard level — is exactly the right call. Cursor isolation (`cursors.kiroCli` vs `cursors.kiro`), per-request fingerprinting for mutability, and Bedrock ARN canonicalization are all well thought out.

A few things I'd like to discuss before merging:

### Strengths
- **Merge strategy is correct**: emitting `source: "kiro"` keeps the brand unified for end users, and isolating cursor state under `cursors.kiroCli` avoids cross-contamination with the IDE parser.
- **Mutability handling**: the per-request-id fingerprint with subtract-old/add-new on change is a nice solution for Kiro CLI's habit of rewriting requests.
- **Bedrock ARN canonicalization** (`anthropic.claude-sonnet-4-20250514-v1:0` → `claude-sonnet-4`) means IDE rows and CLI rows collapse onto the same model when they hit the same Bedrock backend. 
- **SQL-side `json_extract`** avoids loading the full conversations_v2 blob — important given the 93MB scale you mention.
- **Native contract test** (`test/kiro-cli-native-contract.test.js`) pins the Swift `ModelBreakdown` decoder shape — appreciate the defensive testing here. The `total_cost_usd` String-vs-Number lock is exactly the kind of regression that would otherwise crash the macOS app silently in prod.
- **Pricing mirrored** in both `src/lib/local-api.js` and `dashboard/edge-patches/tokentracker-leaderboard-refresh.ts` with a byte-equivalence test — prevents cloud/local cost drift.
- **Consistent pattern**: `cp.execFileSync("sqlite3", ["-json", …])` matches what Kiro IDE / Hermes already do.

### Concerns / requests for changes

1. **Fixture is "PENDING LIVE VALIDATION — spec-derived"** (`test/fixtures/kiro-cli/active-source.json`).  This is the biggest risk. The fixture header openly admits it's reverse-engineered from binary strings rather than captured from a real Kiro CLI session. If the actual `conversations_v2` schema or the per-request fields differ from what's hypothesized, `parseKiroCliIncremental` will silently emit zero buckets (the existing fall-throughs all `continue` on missing fields). Could you grab a real session DB from a machine that's been using Kiro CLI for a few hours and validate? At minimum, please confirm the fields you read (`user_prompt_length`, `response_size`, `model_id`, `request_start_timestamp_ms`, `request_id`, `message_id`) are present in real rows and contain the value types you expect.

2. **Dead-weight back-compat path**.  `parseKiroCliFromSessionFiles` + `resolveKiroCliSessionFiles` read `~/.kiro/sessions/cli/{uuid}.json`, but Kiro CLI was never previously integrated into TokenTracker — there are no prior callers of these to be backwards compatible with. The "back-compat" comment is misleading. Two suggestions, in order of preference:
   - **Remove the session-file path entirely** and keep only the SQLite-backed implementation that real Kiro CLI uses. Drop `test/fixtures/kiro-cli/active-source.json`, `empty-source.json`, the `parseKiroCliFromSessionFiles` function, and the corresponding tests in `test/rollout-parser.test.js` (the SQLite-backed test at the bottom is the one that mirrors reality).
   - Or, if you have a use case in mind, add a comment explaining when the session-file path actually fires (currently it only fires when a caller passes `sessionFiles` explicitly — i.e. tests).

3. **Documentation drift on `~approx`**.  In `src/lib/diagnostics.js`, the `kiro.cli_approximation` text says:  
   > "Model names are suffixed '~approx' in model-breakdown."  
   
   …but `parseKiroCliIncremental` never appends `~approx` — the model is set to either the canonicalized clean name (e.g. `claude-sonnet-4`) or the literal `kiro-cli-agent`. Same for the comment block in `src/commands/sync.js:367`. Either:
   - drop the `~approx` mention from both places, or
   - actually mark the model name (e.g. `claude-sonnet-4~approx`) so users can visually distinguish CLI's approximated counts from IDE's exact counts in the model-breakdown panel.

   I'd lean toward dropping the doc — keeping clean model names is what makes IDE+CLI rows collapse, and that collapse is the headline UX win.

4. **Token-approximation precision is worth calling out in CHANGELOG**.  4 chars/token works as a rough Anthropic heuristic but breaks for code-heavy prompts (denser) and tool-output responses (looser). Dashboard cost numbers will be ±20-30% off reality for CLI rows. A line in the CHANGELOG noting "Kiro CLI counts are approximated from request char lengths since Kiro CLI does not persist token counts" would set expectations.

5. **Verify `response_size` semantics**.  Is `response_size` the byte length of the full assistant response payload (which would include JSON metadata, tool-call envelopes, etc.) or just the assistant text? The 4 chars/token approximation depends on it being human-readable text. If it's raw payload bytes including JSON wrapping, the approximation will systematically over-count. Worth a quick check against a real session.

6. **Light test coverage on `canonicalizeKiroCliModelId`**.  The mutable-request test exercises `anthropic.claude-sonnet-4-20250514-v1:0` only. Could you add a small unit test enumerating each transformation rule (full Bedrock ARN with region, lone date suffix, `auto`, empty/null, `claude-opus-4.6` already-clean, etc.)? Right now if any regex breaks subtly we'd never know.

7. **Two near-duplicate canonicalizers**.  `canonicalizeKiroCliModelId` (CLI) and `normalizeKiroModelName` (IDE, line 2791) do almost the same job with slightly different rules. Either:
   - unify them into one canonicalizer that both paths call (preferred — guarantees IDE and CLI collapse identically), or
   - add a comment to each explaining why they're intentionally separate.

   The risk of keeping them divergent: an IDE-emitted `CLAUDE_SONNET_4_20250514_V1_0` becomes `claude-sonnet-4`, but suppose a future Kiro CLI emits the same model in a slightly different encoding — they'd end up as separate model rows under the same source, defeating the merge.

### Nits
- `src/commands/sync.js:367` comment says CLI rows are marked `'<model>~approx'` — see point 3 above.
- `src/lib/rollout.js:3148` `request_start_timestamp_ms` is read with `Number()` but never validated for non-numeric strings before the `Number.isFinite` check; not a bug, but a typed `parseInt` would be slightly cleaner.

### Tests to run before merge
- `npm test` (full suite)
- A manual sanity check on a real machine with both Kiro IDE and Kiro CLI active: confirm the dashboard's Kiro card shows merged totals, and the model-breakdown panel collapses (or doesn't) IDE and CLI rows as expected.

Overall I'm a strong +1 on the architecture; the concerns above are mostly about validating the SQLite schema assumptions against real data and tightening the doc/code consistency before merge. Once the fixture is validated against a live session and the back-compat path is removed (or justified), this should be ready to land.

Thanks again for the careful work!

---

mm7894215:
Merged — thanks a ton for the contribution @CiprianSpiridon! The SQLite-backed parse, per-request fingerprint cursor, and Bedrock ARN canonicalization made this way cleaner to reason about than trying to bolt CLI on top of the IDE path. Shipping as part of v0.5.73 today.

While validating against live data I landed one follow-up fix in a separate commit on main (39c8f39) to cover three edges we hit:

- **Prompt char attribution.** `turn.message_ids` never actually contains the user Prompt's `message_id` — only AssistantMessage + ToolResults — so the char-count fallback was summing 0 prompt chars and reporting `input_tokens: 0` on every turn. The fix walks the `.jsonl` sequentially and attributes each Prompt to the turn whose first AssistantMessage follows it.
- **`model_id: "auto"` pricing.** Kiro CLI auto-routing writes the literal string `"auto"` into `rts_model_state.model_info.model_id`. That was falling into the `lower === "auto"` branch of `getPricing` and getting Cursor's `composer-1` rates (~$10/M output) instead of `kiro-cli-agent` (~$15/M). `canonicalizeKiroCliModelId("auto")` now returns null so the caller's `|| "kiro-cli-agent"` fallback kicks in.
- **Test isolation.** The `canonicalizes Bedrock model IDs` test wasn't passing `KIRO_HOME`, so on any dev machine with real kiro-cli sessions the parser picked them up and the assertion exploded. Added `KIRO_HOME: tmp` to the test env. Also flipped the lock-sibling test to match the implementation's live-tracking intent (include `.lock` sessions; the fingerprint delta already handles mutations).

End-to-end re-check on my machine after the fix: `model: "claude-sonnet-4.5"`, `input_tokens: 16`, `output_tokens: 2596`, cost `$0.039` at sonnet rates. Appreciate the work!

PR #19, opened 2026-04-20
fix: tighten local dashboard and cloud sync security boundaries

### LINKED ISSUE #17: Bind the local dashboard API to loopback and tighten local sync forwarding
## Summary
The local dashboard server currently listens on all interfaces and exposes a localhost-oriented API surface that was meant for same-machine use only. Combined with permissive local API responses and the ability to pass an arbitrary `insforgeBaseUrl` into `/functions/tokentracker-local-sync`, this makes it too easy to read local usage data or forward queued usage to an attacker-controlled endpoint.

## Suggested fix
- Bind `tokentracker serve` explicitly to loopback
- Stop advertising wildcard CORS on the local API/static server
- Reject `insforgeBaseUrl` overrides unless they match the configured InsForge base URL
- Keep the auth proxy pointed at the configured runtime base URL only

## Why
This keeps the local dashboard server aligned with its intended trust boundary: the current machine and the current user session, not the broader network.

### LINKED ISSUE #18: Stop persisting cloud upload device tokens in dashboard localStorage
## Summary
The dashboard currently stores the issued cloud upload device token in `localStorage` and reuses it for later syncs. That token is accepted by the ingest edge function as the write credential, so persisting it in browser storage makes token theft materially easier than it needs to be.

## Suggested fix
- Keep the device token in memory for the active dashboard session instead of writing it to `localStorage`
- Remove any legacy persisted token on read/write so old sessions do not linger
- Continue to mint a fresh device token when a new dashboard session needs to sync
- Follow up later with stronger server-side expiry/rotation if needed

## Why
This narrows the exposure window for a credential that can write usage data on behalf of a user and avoids leaving bearer tokens behind in browser storage between sessions.

EricGrill:
## Why
The local dashboard server and cloud-sync token lifecycle were broader than they needed to be for a localhost-oriented app. This patch narrows those trust boundaries and addresses the two security review findings tracked in #17 and #18.

## What changed
- bind `tokentracker serve` to loopback and advertise the loopback URL
- remove wildcard CORS from the local server surface
- restrict `/functions/tokentracker-local-sync` overrides to the configured InsForge base URL
- keep dashboard cloud device tokens in memory and purge any legacy persisted token
- add regression coverage for the local sync guard, loopback host selection, and in-memory device-token handling

## Verification
- `node --test test/local-api-security.test.js test/serve-bind-host.test.js test/cloud-sync-prefs.test.js test/cookie-relay-persistence.test.js`

## Follow-up / workspace notes
- `npm test` is currently blocked in this workspace by missing root dev modules such as `esbuild` and `@sourcegraph/scip-typescript`
- `npm run validate:guardrails` reports existing failures unrelated to this patch
- `dashboard/node_modules` is not installed in this workspace, so dashboard typecheck/build could not be run here

Closes #17
Closes #18

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Services/APIClient.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Services/APIClient.swift
@@ -2,6 +2,9 @@ import Foundation
 
 actor APIClient {
     static let shared = APIClient()
+    private struct LocalAuthResponse: Decodable {
+        let token: String
+    }
 
     private let baseURL = Constants.serverBaseURL
     private let session: URLSession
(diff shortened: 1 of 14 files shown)


mm7894215:
Merged. I also shipped a follow-up patch release on top of #19: v0.5.81 is now pushed to main, with two post-merge fixes found during verification: the local dashboard now normalizes legacy Codex queue rows correctly, and the macOS app now keeps regular Cloud IP Check iframe navigation inside the embedded view instead of promoting it to a new page. npm publish has been triggered from the push, and the DMG release workflow has been started for 0.5.81.

PR #27, opened 2026-04-20
feat: add local session outcome markers and filters

### LINKED ISSUE #24: [feature]: Add session outcome markers to distinguish productive vs wasted usage
## Summary
Add a lightweight way to tag or classify sessions so token usage can be separated into productive work, failed attempts, exploration, or wasted spend.

## Why this would help
Raw token counts alone do not show whether usage was worth it. A simple outcome layer would make the data much more useful for improving workflows.

## Suggested scope
- optional session labels such as productive, exploratory, blocked, or wasted
- aggregate charts by outcome category
- filters for outcome-based review
- optional quick tagging after sync or in the dashboard

## Example questions this should answer
- Which tools produce the most wasted spend?
- Are expensive sessions also the most productive?
- Which workflows burn tokens without finishing tasks?

EricGrill:
## Why
Issue #24 asks for a way to distinguish productive usage from exploratory or wasted usage.

## What changed
- added a local session-buckets endpoint derived from recent queue entries
- added local outcome tagging storage and counts
- added a Sessions tab with outcome filters and per-session labeling

## Verification
- `node --test test/session-outcomes.test.js`
- `npm run validate:copy`

Closes #24


mm7894215:
Closing this PR.

The current implementation labels hourly `source|model|hour_start` buckets as sessions. That merges unrelated work into one outcome and makes the feature semantically misleading. It also exposes a local-only Sessions tab in cloud/public contexts without a viable empty/error state. This should be redesigned around a real session identity, or explicitly reframed as usage buckets, before reopening.

PR #28, opened 2026-04-20
feat: add local budget alerts with projected overruns

### LINKED ISSUE #21: [feature]: Add actionable budget alerts with forecasted overrun warnings
## Summary
Add budget alerts that do more than report raw spend. The app should detect unusual usage and warn when current trends are likely to exceed a user-defined daily, weekly, or monthly budget.

## Why this would help
TokenTracker is most useful when it helps me act before costs get out of hand, not after.

## Suggested scope
- user-defined budgets by day, week, and month
- alerts for absolute thresholds and unusually large spikes vs recent baseline
- simple spend forecast based on current pace
- one-click jump from the alert to the source/model/project causing the spike

## Example
"You are on pace to exceed your weekly budget by 28%. Codex on project X is responsible for most of the increase."

EricGrill:
## Why
Issue #21 asks for actionable budget alerts that warn before spend gets out of hand.

## What changed
- added local daily, weekly, and monthly budget preferences in Settings
- added a dashboard alert when current spend is over budget or on pace to exceed it
- added a jump from the alert into the project usage view

## Verification
- `node --test test/budget-alerts.test.js`
- `npm run validate:copy`

Closes #21

### DIFF
--- a/dashboard/src/components/settings/BudgetAlertsSection.jsx
+++ b/dashboard/src/components/settings/BudgetAlertsSection.jsx
@@ -1,0 +1,81 @@
+import React, { useCallback, useId, useState } from "react";
+
+import { copy } from "../../lib/copy";
+import {
+  getBudgetAlertPrefs,
+  setBudgetAlertPrefs,
+} from "../../lib/budget-alerts.js";
+import { SectionCard, SettingsRow } from "./Controls.jsx";
+
+export function BudgetAlertsSection() {
+  const [prefs, setPrefs] = useState(() => getBudgetAlertPrefs());
+
+  const updateField = useCallback(
+    (key, value) => {
+      const next = setBudgetAlertPrefs({ ...prefs, [key]: value });
+      setPrefs(next);
+    },
+    [prefs],
+  );
+
+  return (
+    <SectionCard
+      title={copy("settings.section.budgets")}
+      subtitle={copy("settings.section.budgetsSubtitle")}
+    >
+      <BudgetSettingsRow
+        label={copy("settings.budget.daily")}
+        hint={copy("settings.budget.dailyHint")}
+        value={prefs.daily}
+        onChange={(value) => updateField("daily", value)}
+      />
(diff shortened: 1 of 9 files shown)


mm7894215:
Thanks for the contribution. This cannot be merged as-is because the branch conflicts with current `main` in `dashboard/src/pages/SettingsPage.jsx`.

After rebasing and resolving the conflict, please also add accessible labels for the new budget inputs. The current number fields only have a placeholder, which is not enough for screen readers.

EricGrill:
Addressed review feedback:

- Rebased onto current `main` and resolved the `SettingsPage.jsx` conflict by adopting the new extracted-section pattern (moved `BudgetAlertsSection` into `dashboard/src/components/settings/BudgetAlertsSection.jsx` to match `AccountSection`/`MenuBarSection`/`AppearanceSection`).
- Added accessible labels for the budget number inputs via `aria-labelledby` (associating the visible `SettingsRow` label) plus `aria-describedby` for the hint text. Each row uses `useId()` so the label/hint are programmatically linked instead of relying on the `placeholder="0"` alone.

Tests still pass (`node --test test/budget-alerts.test.js`); `validate:copy` and `validate:ui-hardcode` show no new warnings introduced by this branch.

mm7894215:
Thanks for picking up #21. Going to close this one too.

Budget alerts are a feature where "correct" depends a lot on the user. Flat daily/weekly/monthly thresholds work for some people, but most users I talk to care about a specific plan ceiling (Claude Pro, Codex Plus, Cursor business) or per-project spend, not a rolling global number. The Limits page already covers plan-level rate limits and caps; adding a parallel rolling-budget UI alongside it would mean two competing "are you over?" surfaces.

The localStorage-only state also limits the value — alerts that don't follow a user across devices tend to lose trust quickly. If TokenTracker grows budget alerts properly, they'd probably want to live alongside the cloud sync layer rather than ahead of it.

Leaving #21 open for now. Appreciate the careful PR.

PR #29, opened 2026-04-20
feat: add project drilldowns with period-over-period comparisons

### LINKED ISSUE #23: [feature]: Add project drilldowns with period-over-period comparison
## Summary
Add a project-focused detail view that makes it easy to compare usage across periods and see what changed.

## Why this would help
A big part of useful token tracking is understanding which repo or project is driving increased usage, especially when working across many codebases.

## Suggested scope
- drill into a single project from the dashboard
- compare this week vs last week or this month vs last month
- break down the change by source and model
- highlight the biggest contributors to the delta

## Example questions this should answer
- Which repo got more expensive this week?
- Was the increase driven by Codex, Claude, or Gemini?
- Did one model choice cause most of the jump?

EricGrill:
## Why
Issue #23 asks for project drilldowns that make it easy to compare one period against another.

## What changed
- fixed local project summaries to respect date filters
- added a project detail endpoint with current vs previous period totals and source breakdowns
- added a project drilldown modal in the dashboard

## Verification
- `node --test test/project-drilldown.test.js`
- `npm run validate:copy`

Closes #23

### DIFF
--- a/dashboard/src/content/copy.csv
+++ b/dashboard/src/content/copy.csv
@@ -218,6 +218,14 @@ dashboard.projects.limit_top_3,ui,ProjectUsagePanel,ProjectUsagePanel,limit_top_
 dashboard.projects.limit_top_6,ui,ProjectUsagePanel,ProjectUsagePanel,limit_top_6,"TOP 6",,active
 dashboard.projects.limit_top_10,ui,ProjectUsagePanel,ProjectUsagePanel,limit_top_10,"TOP 10",,active
 dashboard.projects.empty,ui,ProjectUsagePanel,ProjectUsagePanel,empty,"No public repositories",,active
+dashboard.projects.drilldown.title,ui,ProjectUsageDrilldownModal,ProjectUsageDrilldownModal,title,"Project Detail",,active
+dashboard.projects.drilldown.current,ui,ProjectUsageDrilldownModal,ProjectUsageDrilldownModal,current,"Current period",,active
+dashboard.projects.drilldown.previous,ui,ProjectUsageDrilldownModal,ProjectUsageDrilldownModal,previous,"Previous period",,active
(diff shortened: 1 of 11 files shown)


mm7894215:
Thanks for the contribution. This needs a small UX/data-state fix before merge.

When switching projects, `useProjectUsageDetail` keeps the previous project data while the new request is loading, so the modal can briefly show stale title/totals. Please clear data on project change or hide summary content until the current request resolves. The modal should also include a visible Close button and an empty/error state for the source breakdown.

EricGrill:
Addressed review feedback:

- **Stale data on project switch**: `useProjectUsageDetail` now resets `data` and `error` to `null` on every refresh and ignores responses from superseded requests via a request-id ref, so switching projects no longer leaks the previous title/totals into the modal.
- **Visible Close button**: added a `Dialog.Close` button (X icon) in the modal's top-right with `aria-label` from the copy registry, matching the pattern in `CostAnalysisModal`.
- **Empty / error state for source breakdown**: extracted `BreakdownContent` to render distinct loading / error / empty / data states. Summary blocks also collapse to a placeholder while loading or on error so they don't display stale period totals.

New copy keys: `dashboard.projects.drilldown.{loading,error,breakdown_empty,close}`. Rebased onto current `main`. Tests still pass (`node --test test/project-drilldown.test.js`); dashboard build is clean; `validate:copy` reports no new warnings.

mm7894215:
Thanks — I read through this one carefully and I'm going to pass.

Project attribution in TokenTracker is best-effort: it's inferred from the working directory at the time the hook fires, a single chat session can span multiple projects, agentic flows occasionally invent new dirs, and so on. The project view is deliberately understated because of that. Period-over-period comparisons make statistical claims (e.g. "this project went up 24% week-over-week") on top of attribution data that isn't consistent enough to support that kind of confident inference, so the risk of misleading the user goes up.

If the underlying goal is "I want to see what one project did", we already render that on the existing project panel. For richer drilldowns I'd rather invest in attribution stability first, then add UI on top.

Leaving #23 open. Appreciate the well-tested PR.

PR #32, opened 2026-04-21
fix(kiro-cli): correctness + resource-bound hardening on top of 39c8f39

CiprianSpiridon:
## PR Goal (one sentence)

Close three correctness gaps (SQLite-migration double-count, `loop_id.rand=0` coercion, timestamp drift across source tiers), add two resource bounds (cursor cap, streamed jsonl reads), and tighten six guard-rails — on top of `39c8f39`, keeping its Prompt attribution + `auto → null → kiro-cli-agent` logic intact.

## Scope

- `src/lib/rollout.js` — Kiro CLI parser only (no other parser touched; shared `enqueueTouchedBuckets` deliberately left untouched)
- `src/lib/diagnostics.js` — inline the Kiro CLI DB path resolver (drops `require('./rollout')` load cost on status/diagnostics)
- `src/commands/sync.js` — one-word progress-label change for consistency with Hermes
- `test/rollout-parser.test.js` — 3 new regression tests + UUID-shape alignment on 2 existing tests

Non-goals:
- Upstream's Prompt attribution via `turnPromptChars` is preserved
- Upstream's `auto → null → kiro-cli-agent` pricing routing is preserved
- No change to the merged `source='kiro'` contract or any other provider's code path

- **Delta since last Codex review:** one commit `bff9f2b` on top of `39c8f39`. 4 files, +544/-77.
  1. A Kiro CLI turn gets counted exactly once whether it's read from session file, SQLite, or migrates from one to the other.
  2. Cursor size is bounded across time (90-day + 20k-entry cap).
  3. `sync` never blocks on a multi-MB `.jsonl` — reads are streamed.
  4. A negative `conversation_count` never reaches the queue.
  5. All Kiro CLI side effects stay inside the Kiro CLI code path; no cross-parser behavior change.
- **Edge cases covered:**
  - `loop_id.rand === 0` (integer zero preserved via `?? null` chain, not `|| null`)
  - No-`loop_id` turns (cursor key falls back to bare message_id UUID; retraction now matches via a stored `session_id` tag rather than colon-parse)
  - SQLite `conversation_id` column ≠ `user_turn_metadata.continuation_id` on observed data (both are extracted and both feed the migration-match set)
  - `flat.length === 0` after retraction (early-return path now runs cap + clamp via `clampAndCapKiroCliState`)
  - `.jsonl` truncated/deleted mid-read (try-catch around the `for await` returns partial instead of crashing sync)
  - Corrupt/empty `conversation_id` (typed non-empty string check before adding to `migratedConvIds`)
  - Half-hour boundary straddle between SQLite and session-file tiers (unified timestamp precedence on `request_start_timestamp_ms` → `start_timestamp` → `end_timestamp`)
- **Tests run:**
  ```
  node --test test/rollout-parser.test.js test/kiro-fallback.test.js test/model-breakdown.test.js test/status.test.js test/diagnostics.test.js
  # ℹ tests 71 · pass 71 · fail 0
  ```
  Plus one live sync against a real machine with 1,336+ historical SQLite requests and a live session file — no regression.
  - No cursor-schema migration: legacy entries without a `session_id` tag still work via the colon-split fallback
  - Historical queue rows written under legacy `"auto"` model label (from syncs before upstream `39c8f39`) keep their old composer-1 pricing; new rows correctly route to `kiro-cli-agent`
  - No streaming for the `.json` session-state file itself (it's a bounded state struct, typically <100 KB)

### Rules / Invariants

- **Shared queue writer (`enqueueTouchedBuckets`) is untouched.** The TASK-010 conversation_count clamp was narrowed to Kiro's `touchedBuckets` Set inside `parseKiroCliIncremental`, not the cross-parser flush function. Verified with `git diff bff9f2b -- src/lib/rollout.js` showing zero changes in lines 1131-1424.
- **`source='kiro'` merge contract preserved.** Every emitted queue entry from the Kiro CLI path uses `source='kiro'`; the model field is the only runtime differentiator between IDE and CLI rows.
- **Cursor schema is backward compatible.** New entries carry an optional `session_id` tag; old entries without the tag are still retracted via the colon-split fallback.

### Boundary Matrix

1. **SQLite-only install (no session file):** retraction pass runs but `migratedConvIds` only contains IDs that match no cursor entries → no retraction fires, main loop processes normally. Regression-test-equivalent coverage: existing tests that only use SQLite data.
2. **Session-file-only install (no SQLite):** `flatDb = []`, `migratedConvIds` empty, retraction skipped, session-file turns flow through the main loop. Covered by existing upstream test "aggregates user_turn_metadatas into half-hour kiro buckets".
3. **Migration event (session file + SQLite for same conversation):** retraction fires, prior session-file contribution subtracted, new SQLite row added. Covered by the new "retracts orphan session-file contribution when a conversation migrates into SQLite" test with both `conversation_id` and `continuation_id` matching.
4. **Migration event with bare-UUID cursor key (no-loop_id fallback):** retraction matches via `session_id` tag instead of colon-split. Covered by the new "retracts no-loop_id session-file entries via session_id tag" test.
5. **Empty-flat post-retraction (session file removed between syncs):** `flat.length === 0` early-return now runs clamp + cap. Covered by the new "early-return path still runs cap + clamp" test.

### Evidence (tests or repro)

- 3 new tests in `test/rollout-parser.test.js` cover the 3 boundary cases that ship new behavior.
- Live sync on a machine with 1,336 historical SQLite requests + a live session file: all numbers match expected distribution across `claude-opus-4.6`, `claude-sonnet-4.5`, and the `kiro-cli-agent` fallback.
- `grep` of the diff confirms `enqueueTouchedBuckets` is untouched.

## Public Exposure Checklist

- [x] N/A — no public-exposure surface added.

- `parseKiroCliIncremental` main loop + retraction pass ordering. Specifically: retraction must run BEFORE the main loop and the cap must run AFTER both; otherwise active entries get pruned or orphans re-fire.

### Verification method

- [x] Unit test — 3 new tests cover retraction (two variants) and early-return cap
- [x] Live smoke — fresh cursor-reset sync on a real machine with 1,336 SQLite requests; aggregated totals match pre-change distribution within approximation tolerance
- [x] Structural diff — `git diff bff9f2b -- src/lib/rollout.js` confirms no edits outside the Kiro CLI section and no edits to `enqueueTouchedBuckets`

### Uncovered scope

- No test for the streamed-jsonl RSS bound (the plan's TASK-005 AC had a 10 MB fixture claim; skipped to keep the test runtime short)
- No test for the `TOKENTRACKER_DEBUG` stderr log paths (TASK-012 / D-8) — structural grep-based test not added in this round

## Itemized change list

| # | Task | Change | File / location |
|---|---|---|---|
| 1 | TASK-007 + D-1 | Cross-source retraction subtracts orphan session-file cursor entry when its `session_id` matches a migrated SQLite `conversation_id` OR `continuation_id`; drops matching `flatSessions` entries via O(N) filter. | `src/lib/rollout.js` — `readKiroCliRequests` (add `continuation_id` extract); `parseKiroCliIncremental` (new retraction block before `flat = flatDb.concat(flatSessions)`) |
| 2 | TASK-001 | Preserve integer 0 in `loop_id.rand` via `?? null ?? null ?? null` chain. Prevents cursor key from silently collapsing to message_id fallback. | `readKiroCliSessionTurns` |
| 3 | TASK-003 | Resolver filters filenames to canonical 8-4-4-4-12 hex UUID shape. Backup files, scratch, and typos are skipped instead of fed to `JSON.parse`. | `resolveKiroCliSessionFiles` + `KIRO_CLI_SESSION_FILE_RE` |
| 4 | TASK-004 | Cursor cap: drop entries older than 90 days by `bucketStart`, then keep the most recent 20k if still over. Runs after retraction + main merge so no active or just-retracted entry is pruned. | `clampAndCapKiroCliState` helper |
| 5 | TASK-005 | `readKiroCliMessageChars` is async and reads the `.jsonl` via `readline` over `createReadStream` — no whole-file buffer. | `readKiroCliMessageChars`, plus async ripple through `readKiroCliSessionTurns` and the session-file loop in `parseKiroCliIncremental` |
| 6 | TASK-006 | Timestamp precedence: `request_start_timestamp_ms` → `start_timestamp` → `end_timestamp`. Previously `end_timestamp || start_timestamp` would bucket a boundary-straddling turn differently from the SQLite tier. | `readKiroCliSessionTurns` |
| 7 | TASK-010 | Narrow `conversation_count` clamp to Kiro's `touchedBuckets` set — the shared `enqueueTouchedBuckets` is left untouched so legitimate negatives from other parsers (Claude, Codex, Gemini, Hermes, Kimi, Copilot, Cursor, Opencode, Openclaw, Kiro IDE) are never masked. | `clampAndCapKiroCliState` helper |
| 8 | TASK-011 | Inline the Kiro CLI DB path computation in `diagnostics.js` to drop `require('./rollout')`. | `src/lib/diagnostics.js` — new `resolveKiroCliDbPathInline` |
| 9 | TASK-012 + D-8 | `readKiroCliRequests` logs `sqlite3` spawn failures under `TOKENTRACKER_DEBUG=1` (previously silent `[]` was indistinguishable from an empty DB). `env` is threaded from the call site so tests can toggle hermetically. | `readKiroCliRequests` |
| 10 | TASK-013 | Kiro CLI sync progress label: `convs` → `sessions` (matches Hermes's wording). | `src/commands/sync.js` |
| 11 | TASK-014 | `resolveKiroCliSessionFiles` honors `env.HOME` (symmetric with `resolveKiroCliDbPath`). Hermetic testing/CI now works. | `resolveKiroCliSessionFiles` |
| 12 | Bug-1 | Extract clamp + cap into a helper; call it from BOTH the main path AND the `flat.length === 0` early-return path (previously skipped, leaving negatives in the queue and stale cursor entries alive). | `parseKiroCliIncremental` (both exit sites) |
| 13 | Bug-2 | Tag session-origin cursor entries with `session_id` so retraction can match them regardless of `request_id` shape. Colon-parse kept as legacy fallback for pre-tag cursor entries. | `readKiroCliSessionTurns` push + cursor write + retraction read |
| 14 | Bug-5 | Wrap the streamed iteration in try-catch. Mid-read errors return partial data instead of crashing sync. | `readKiroCliMessageChars` |
| 15 | D-6 | `typeof === "string" && length > 0` check on `conversation_id` / `continuation_id` before adding to the migration-match set. | retraction block |
| 16 | D-14 | Replace O(N²) reverse `splice`-in-loop over `flatSessions` with one O(N) `filter` pass during retraction. | retraction block |

## Live verification

```
tokentracker sync

Sync finished:
- Parsed files: 11917
- New 30-min buckets queued: 12
- Uploaded: skipped (no device token)

# Kiro models after sync (deduped latest-wins):
  claude-opus-4.6           in=238,920  out=873,976  convs=4,632
  claude-sonnet-4.5         in=  2,003  out=  8,785  convs=   15
  kiro-cli-agent            in=  5,613  out= 32,519  convs=  182
  (plus legacy "auto" and "kiro-auto" rows from pre-merge syncs)
```

### DIFF
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -385,7 +385,7 @@ async function cmdSync(argv) {
             if (!progress?.enabled) return;
             const pct = p.total > 0 ? p.index / p.total : 1;
             progress.update(
-              `Parsing Kiro CLI ${renderBar(pct)} ${formatNumber(p.index)}/${formatNumber(p.total)} convs | buckets ${formatNumber(p.bucketsQueued)}`,
+              `Parsing Kiro CLI ${renderBar(pct)} ${formatNumber(p.index)}/${formatNumber(p.total)} sessions | buckets ${formatNumber(p.bucketsQueued)}`,
             );
           },
         });
--- a/src/lib/diagnostics.js
+++ b/src/lib/diagnostics.js
@@ -16,7 +16,20 @@ const { normalizeState: normalizeUploadState } = require("./upload-throttle");
 const { probeOpenclawHookState } = require("./openclaw-hook");
 const { probeOpenclawSessionPluginState } = require("./openclaw-session-plugin");
 const { resolveTrackerPaths } = require("./tracker-paths");
-const { resolveKiroCliDbPath } = require("./rollout");
+// TASK-011: Kiro CLI DB path inlined here to avoid pulling the ~4000-line
+// rollout module on every `tokentracker status` / `diagnostics` call.
+// rollout.js still exports resolveKiroCliDbPath for external callers.
+function resolveKiroCliDbPathInline(env, home) {
+  if (env.KIRO_CLI_DB_PATH) return env.KIRO_CLI_DB_PATH;
+  const effectiveHome = env.HOME || home;
+  return path.join(
+    effectiveHome,
+    "Library",
+    "Application Support",
+    "kiro-cli",
+    "data.sqlite3",
+  );
+}
 
 async function collectTrackerDiagnostics({
   home = os.homedir(),
@@ -98,7 +111,7 @@ async function collectTrackerDiagnostics({
   const kiroIdePresent =
     (await safeStatSize(path.join(kiroIdeDevDataDir, "devdata.sqlite"))) > 0 ||
     (await safeStatSize(path.join(kiroIdeDevDataDir, "tokens_generated.jsonl"))) > 0;
-  const kiroCliDbPath = resolveKiroCliDbPath(process.env);
+  const kiroCliDbPath = resolveKiroCliDbPathInline(process.env, home);
   const kiroCliPresent = require("node:fs").existsSync(kiroCliDbPath);
 
   const lastSuccessAt = uploadThrottle.lastSuccessMs
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -3085,20 +3085,31 @@ function resolveKiroCliDbPath(env = process.env) {
   return path.join(home, "Library", "Application Support", "kiro-cli", "data.sqlite3");
 }
 
+// Bug-4: canonical UUID shape — 8-4-4-4-12 hex groups. The looser
+// /^[0-9a-f-]{36}\.json$/ form accepted `36 hyphens`.json or 36 hex with
+// no hyphens. kiro-cli writes proper UUIDs; lock to the canonical shape.
+const KIRO_CLI_SESSION_FILE_RE =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\.json$/i;
+
 // Lists ~/.kiro/sessions/cli/{uuid}.json files. Includes files whose sibling
 // .lock is present — we read those as tail-only snapshots so a running
 // session's completed turns still land in the queue on the next sync. The
 // .json files are rewritten atomically by kiro-cli on each turn flush, so
 // a stale read just means we'll pick up the rest next time.
+//
+// TASK-014: env.HOME is honored (symmetric with resolveKiroCliDbPath) so
+// callers can redirect to a tmp home for hermetic tests/CI.
 function resolveKiroCliSessionFiles(env = process.env) {
-  const home = require("node:os").homedir();
+  const home = env.HOME || require("node:os").homedir();
   const kiroHome = env.KIRO_HOME || path.join(home, ".kiro");
   const sessionsDir = path.join(kiroHome, "sessions", "cli");
   if (!fssync.existsSync(sessionsDir)) return [];
   const files = [];
   try {
     for (const entry of fssync.readdirSync(sessionsDir)) {
-      if (!entry.endsWith(".json")) continue;
+      // TASK-003: only canonical {uuid}.json files; backups, scratch,
+      // typos are skipped so they don't feed JSON.parse garbage.
+      if (!KIRO_CLI_SESSION_FILE_RE.test(entry)) continue;
       files.push(path.join(sessionsDir, entry));
     }
   } catch {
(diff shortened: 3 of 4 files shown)


mm7894215 (commented):
This change introduces a correctness regression in the Kiro CLI migration path.

The new retraction logic is keyed at the whole `session_id` / `conversation_id` level, not at the individual turn/request level. In `parseKiroCliIncremental`, once a conversation appears in SQLite, you retract all prior session-file cursor entries for that session and then filter out all current `flatSessions` entries for the same session.

That breaks the common live-session case where:
- older turns have already been flushed to SQLite
- newer turns still exist only in `~/.kiro/sessions/cli/{session}.json`

In that state, the newer session-only turns are dropped entirely, so sync undercounts active Kiro CLI usage until those turns eventually land in SQLite.

Minimal repro:
1. Run 1: session file contains turn A only -> parsed from session file.
2. Run 2: SQLite contains turn A, but session file contains turn A + turn B.
3. Current code retracts the old session entry for A, but also filters out B because it shares the same `session_id`.
4. Result: only A survives; B is lost.

Please fix this before merge. The retraction needs to be request/turn-granular rather than conversation-granular, or otherwise preserve session-file turns that have not yet been materialized in SQLite. A regression test for the mixed state “SQLite has old turns, session file still has newer turns” is also needed.

CiprianSpiridon:
Thanks — you're right, the session-level retraction was too coarse. Fixed in `49561d0`.

**Root cause**: the `flatSessions.filter` dropped every session-file turn whose `session_id` matched any migrated conversation in SQLite. In a live session with older turn A flushed to SQLite + newer turn B still session-only, both A and B were dropped — only A came back through the SQLite path, B was lost until it later migrated.

**Fix (turn-granular)**:
- `readKiroCliSessionTurns` now surfaces the full `all_message_ids` list on each flat record (was only `messageIds[0]`).
- `parseKiroCliIncremental` builds `migratedMsgIds` from SQLite's `r.message_id` alongside the existing `migratedConvIds`.
- Filter drops a session-file turn only when **any** of its `all_message_ids` appears in `migratedMsgIds` (that specific turn has been flushed). Un-migrated newer turns in the same session survive.
- Cursor retraction intentionally stays session-level: un-migrated turns still in the session file are re-added in the same run via the turn-granular filter, so the net delta for those is zero. Traced through every mixed case — correct.

**Edge case**: a session-file turn with empty `turn.message_ids` carries no match key. Such a turn under the no-loop_id fallback already has `requestId = null` and is filtered upstream at line 3268, so the loop_id case is the only residual. We keep it (conservative) rather than drop it — prevents the reported regression at the cost of a narrow potential double-count which the original session-level approach also mishandled.

**Regression test** (new, mirrors your repro):
- Run 1: session file has turn A + turn B, SQLite empty → both counted via session-file parse.
- Run 2: SQLite now has turn A (msgA), session file still has A + B → expect A's session-file cursor retracted, A's SQLite cursor added, B's session-file cursor preserved. Bucket = A (100/200) + B (60/30) = 160/230.

Also asserts `${convId}:11` (B's cursor key) stays in `requests`, which would regress the instant anyone reintroduces session-level filtering.

57/57 kiro-affected tests green (includes the existing two retraction tests + the new mixed-state one + the early-return cap/clamp test). The 3 remaining fails on full `npm test` (dashboard layout + tsc TS7016 on `InsforgeAuthContext.jsx`) are pre-existing on the branch base and unrelated to rollout parsing.

mm7894215 (approved):
Re-checked the latest update, including the mixed-state Kiro CLI migration case that previously dropped newer session-file turns after older turns had already flushed to SQLite.

The turn-granular filtering in `49561d0` fixes that regression correctly. I also re-validated the behavior against the earlier minimal repro and the Kiro regression tests. From my side, the Kiro CLI migration / retraction design is now reasonable and the earlier blocking correctness issue is resolved.

PR #50, opened 2026-05-03
fix: keep menu bar app alive on launch

ddnio:
## Summary

- Keep the macOS menu bar app alive by disabling AppKit automatic termination on launch.
- Stop running silent update installation from `applicationDidFinishLaunching`; users can still check for updates manually from the menu.

## Root Cause

The downloaded menu bar app could start, create its status item, launch the local dashboard server, and then exit shortly afterward. Once the `LSUIElement` host process exited, Control Center removed the menu bar item, making the app appear to flash and disappear.

The startup path also triggered silent update checks, which can enter the install/relaunch path immediately after launch. That is too aggressive for a menu bar host process because it can terminate the visible app before users can interact with it.

## Validation

- `xcodebuild -scheme TokenTrackerBar -configuration Release -derivedDataPath build/DerivedData CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= build`
- Installed the locally built app to `/Applications/TokenTrackerBar.app`.
- Verified `TokenTrackerBar` was still running after the previous failure window.
- Verified `http://127.0.0.1:7680/functions/tokentracker-user-status` returned `200`.


mm7894215:
Reviewed the diff end-to-end. Direction is right but a couple things to address before merge:

**Diff** (single file, `TokenTrackerBar/TokenTrackerBarApp.swift`, +2/-2):
1. Add `ProcessInfo.processInfo.disableAutomaticTermination(...)` — defensive insurance. Note: `LSUIElement: true` apps don't normally trigger AppKit auto-termination unless `NSSupportsAutomaticTermination=YES` is set in Info.plist (this app doesn't), so the flag is belt-and-suspenders. Fine to keep.
2. Remove `UpdateChecker.shared.check(silent: true)` from `applicationDidFinishLaunching` — this is the actual root-cause fix. `UpdateChecker.swift:407–415` shows the silent path can call `mountCopyRelaunch` then `AppDelegate.requestQuit()` 1s later, terminating the freshly-launched host process before the user can interact. The loop guard at line 115 helps but doesn't cover all races. Removing this from launch is correct.

**Remaining concerns:**

1. **Auto-update is now completely gone.** Users will only update when they manually click "Check for Updates" from the menu (still wired at `StatusBarController.swift:613` and `NativeBridge.swift:165`). Suggest replacing the launch-time check with a delayed background timer (e.g., 30 minutes after launch, then every 24h). Otherwise stale installs accumulate over time.

2. **Version bump missing.** This repo's release flow requires bumping both `package.json` and `TokenTrackerBar/project.yml` `MARKETING_VERSION` for any Swift-touching change (the DMG embeds the CLI runtime, so npm + DMG ship together). Please bump to `0.6.2` (v0.6.1 just shipped with Codex fixes).

3. **PR is still draft.** When you address the above, mark ready for review and we'll merge.

Code itself is correct; the gating items are: deferred auto-update timer + version bump + ready-for-review.

PR #53, opened 2026-05-04
fix(hermes): track real-time token updates for active sessions

ztybyass:
## Problem

TokenTracker significantly under-reports Hermes Agent token usage for active (in-progress) sessions.

Hermes writes real-time token counts to `~/.hermes/state.db` after every turn — the `input_tokens`, `output_tokens`, `cache_read_tokens` etc. columns are continuously updated even while a session is still running (`ended_at IS NULL`). However, the current Hermes parser has two bugs:

### Bug 1: Active sessions are skipped after the cursor advances

The SQL query uses `WHERE started_at > ${cursor}`. After the first sync reads an active session, the cursor jumps to its `started_at` value. All subsequent syncs skip that session entirely, so token growth from continued conversation is never captured.

**Example:** A session starts at 19:56 and runs for 39 minutes. The first sync reads 100K tokens. By the time it ends at 20:35, it has accumulated 950K input tokens. TokenTracker only ever records the 100K snapshot.

### Bug 2: No delta computation causes double-counting

If a session is somehow re-read (e.g. after a cursor reset), the full cumulative token count is added again as if it were new usage, leading to inflated numbers.

---

### 1. Include active sessions in every sync

```sql
-- Before
WHERE started_at > ${cursor}

-- After
WHERE (started_at > ${cursor} OR ended_at IS NULL)
```

Active sessions (`ended_at IS NULL`) are now re-fetched on every sync regardless of cursor position.

### 2. Per-session snapshot delta

A `snapshots` object is stored in `cursors.hermes` mapping session IDs to their last-seen token counts:

```json
{
  "session_abc": { "in": 950000, "out": 17931, "cacheRead": 4421120, ... }
}
```

On each sync, only the difference (`current - snapshot`) is emitted as a new delta. First-time sessions with no prior snapshot emit their full amount. Sessions with zero delta are skipped entirely.

### 3. Cursor only advances past completed sessions

The cursor is renamed to `lastCompletedStartedAt` and only updates when `row.ended_at` is non-NULL. Active sessions remain within the query window until they end.

---

## Backward Compatibility

- Old cursors that lack `lastCompletedStartedAt` or `snapshots` gracefully fall back to `0` / `{}`, so the first run after upgrade performs a full scan with correct delta logic from that point forward.
- The old `lastStartedAt` key is still written (with the same value) for any external code that may read it.

---

## Testing

Verified against live data on macOS:

| Source | Input Tokens | Output Tokens | Cache Read |
|---|---|---|---|
| `state.db` (ground truth) | 2,287,702 | 41,168 | 6,454,912 |
| TokenTracker **before** fix | ~348K | ~5K | ~215K |
| TokenTracker **after** fix | matches | matches | matches |

After fix, running `tokentracker sync` while a session is active correctly captures only the incremental tokens since the previous sync.

Added 3 new unit tests (88/88 pass):
- `tracks real-time token growth for active sessions` — full lifecycle: active → grows → ends
- `skips active session when delta is zero` — no spurious re-counting
- `backward compat: old cursor without snapshots` — migration from old cursor format

---

## Files Changed

- `src/lib/rollout.js` — `readHermesSessions()` and `parseHermesIncremental()`
- `test/rollout-parser.test.js` — 3 new test cases

### DIFF
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -2952,10 +2952,13 @@ function resolveHermesDbPath() {
   return path.join(home, ".hermes", "state.db");
 }
 
-function readHermesSessions(dbPath, sinceEpoch) {
+function readHermesSessions(dbPath, lastCompletedEpoch) {
   if (!dbPath || !fssync.existsSync(dbPath)) return [];
-  const since = Number.isFinite(sinceEpoch) && sinceEpoch > 0 ? sinceEpoch : 0;
-  const sql = `SELECT id, model, started_at, ended_at, input_tokens, output_tokens, cache_read_tokens, cache_write_tokens, reasoning_tokens, message_count FROM sessions WHERE started_at > ${since} AND (input_tokens > 0 OR output_tokens > 0 OR cache_read_tokens > 0 OR reasoning_tokens > 0) ORDER BY started_at ASC`;
+  const since = Number.isFinite(lastCompletedEpoch) && lastCompletedEpoch > 0 ? lastCompletedEpoch : 0;
+  // Fetch sessions that started after the cursor, OR sessions that are still
+  // in-progress (ended_at IS NULL).  Hermes updates token counts in real-time,
+  // so an active session keeps growing and must be re-read on every sync.
+  const sql = `SELECT id, model, started_at, ended_at, input_tokens, output_tokens, cache_read_tokens, cache_write_tokens, reasoning_tokens, message_count FROM sessions WHERE (started_at > ${since} OR ended_at IS NULL) AND (input_tokens > 0 OR output_tokens > 0 OR cache_read_tokens > 0 OR reasoning_tokens > 0) ORDER BY started_at ASC`;
   let raw;
   try {
     raw = cp.execFileSync("sqlite3", ["-json", dbPath, sql], {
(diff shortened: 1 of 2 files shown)


mm7894215:
Reviewed end-to-end. Approving.

**Code review**:
- SQL change `WHERE (started_at > cursor OR ended_at IS NULL)` is correct
- `Math.max(0, current - snapshot)` properly clamps counter resets, no negative deltas emitted
- snapshot map auto-GCs: once a session ends and `started_at <= cursor`, it stops being SELECTed and falls out of `nextSnapshots` on the next pass — no unbounded growth
- Backward compat: old cursors with only `lastStartedAt` fall through to `0` / `{}` and the first run after upgrade does a clean full scan
- One-time over-count for already-active sessions during upgrade is acceptable and strictly better than the prior under-count behavior

**Tests**: 88/88 pass on `node --test test/rollout-parser.test.js` (9.98s). The 3 new tests genuinely exercise active-growth (uses `sqlite3 UPDATE` between calls), zero-delta skip, and old-cursor migration — they're not assertion-light.

**Full suite**: `npm test` shows 463/465 — the 2 failures (`dashboard-typescript-guardrails`, `init-local-runtime-reinstall`) reproduce identically on `main` and are pre-existing environment issues (broken local `tsc` binary + missing `dashboard/dist`), unrelated to this PR.

Will ship in v0.6.1. Thanks for the thorough fix and the ground-truth comparison table.

PR #57, opened 2026-05-07
feat(pi): add support for @mariozechner/pi-coding-agent sessions

### LINKED ISSUE #56: Support standard pi coding agent sessions (~/.pi/)
## Support standard pi coding agent (`@mariozechner/pi-coding-agent`)

Hey, love the project — the dashboard is really slick.

I noticed TokenTracker added support for oh-my-pi in #45, which reads sessions from `~/.omp/agent/sessions/`. That's great, but oh-my-pi and `@mariozechner/pi-coding-agent` are different projects that happen to share "pi" in the name.

The standard pi coding agent stores its sessions at `~/.pi/agent/sessions/`, so the current passive reader doesn't pick them up. It would be great to have support for the standard pi install as well — it's the one published on npm (`@mariozechner/pi-coding-agent`) and has a decent user base.

The session format should be identical (JSONL with tree structure), so adding another reader for `~/.pi/agent/sessions/**/*.jsonl` might be straightforward.

Would you be open to adding this? Happy to help test or provide session samples if needed.

season179:
## Summary

Adds a passive JSONL reader for the standard pi coding agent (`@mariozechner/pi-coding-agent`) at `~/.pi/agent/sessions/**/*.jsonl`. This is distinct from the existing oh-my-pi reader (`~/.omp/`) — same on-disk session format, different install location and namespace.

- New parser `parsePiIncremental` in `src/lib/rollout.js`, emits `source="pi"` with dedup state under `cursors.pi` (capped at 10k seen IDs, mirrors omp).
- Wired into `sync`, `init` (detection), and `status` (reporter line).
- Honors `PI_CODING_AGENT_DIR` env override (matches pi's own `getAgentDir()`), with tilde expansion.
- Adds `leaderboard.column.pi` copy entry for parity with the omp column copy.

## Bug fix included

`resolveOmpAgentDir` and the omp init-detection block previously honored `PI_CODING_AGENT_DIR`, but that env var actually belongs to pi-coding-agent (per upstream `getAgentDir()`). Once both providers exist, a pi user setting the variable would have had every assistant message queued twice — once as `source:"omp"`, once as `source:"pi"` — inflating totals 2×. Removed it from the omp side; reserved for pi only. A regression test asserts omp does **not** pick up sessions when `PI_CODING_AGENT_DIR` is set.

## Test plan

- [x] `node --test test/rollout-parser.test.js` — 93 pass / 0 fail (88 pre-existing + 5 new pi tests covering parse / dedup / missing dir / env override / collision regression)
- [x] `npm run validate:copy` — passes
- [x] `npm run validate:ui-hardcode` — no new warnings introduced
- [x] `npm run validate:guardrails` — passes
- [x] Verified locally against real pi sessions at `~/.pi/agent/sessions/--Users-season-Personal-pi-mono--/*.jsonl` (the session that produced this PR is the first ingested example)

5 pre-existing test failures in the broader suite (`tsc validates migrated TS files`, `graph-auto-index-*` ×3, `installLocalTrackerApp replaces stale installed runtime`) are environmental/infrastructure — they need `dashboard/node_modules`, `scip-typescript`, and `dashboard/dist/index.html` respectively. None are touched by this change.

## Notes

- ~225 lines of the pi parser duplicate the omp parser. I considered unifying them via a shared parameterized helper, but every other passive-reader provider in the repo (Hermes, Kimi, CodeBuddy, Craft) is isolated as a self-contained block — keeping the same convention here for reviewability and to avoid an out-of-scope refactor.
- No version bump in this PR; happy to bump `package.json` separately when the maintainer wants to ship.

Closes #56

### DIFF
--- a/dashboard/src/content/copy.csv
+++ b/dashboard/src/content/copy.csv
@@ -403,6 +403,7 @@ leaderboard.column.copilot,dashboard,LeaderboardPage,LeaderboardPage,column_copi
 leaderboard.column.kimi,dashboard,LeaderboardPage,LeaderboardPage,column_kimi,"Kimi Code",,active
 leaderboard.column.codebuddy,dashboard,LeaderboardPage,LeaderboardPage,column_codebuddy,"CodeBuddy",,active
 leaderboard.column.omp,dashboard,LeaderboardPage,LeaderboardPage,column_omp,"oh-my-pi",,active
+leaderboard.column.pi,dashboard,LeaderboardPage,LeaderboardPage,column_pi,"pi",,active
 limits.copilot.otelHint.title,dashboard,LimitsPage,LimitsPage,otel_hint_title,"Track Copilot token usage",,active
 limits.copilot.otelHint.body,dashboard,LimitsPage,LimitsPage,otel_hint_body,"Limits load via your GitHub OAuth token. To also record per-token usage, enable Copilot's OpenTelemetry export and add these lines to your shell profile (~/.zshrc or ~/.bashrc):",,active
 limits.copilot.otelHint.copy,dashboard,LimitsPage,LimitsPage,otel_hint_copy,"Copy",,active
--- a/dashboard/src/ui/matrix-a/components/ProviderIcon.jsx
+++ b/dashboard/src/ui/matrix-a/components/ProviderIcon.jsx
@@ -158,6 +158,16 @@ function OmpIcon({ size = 16, className = "" }) {
   );
 }
 
+function PiIcon({ size = 16, className = "" }) {
+  return (
+    <svg width={size} height={size} viewBox="0 0 120 90" fill="currentColor" className={className}>
+      <rect x="10" y="8" width="100" height="12" rx="2" />
+      <rect x="25" y="20" width="12" height="62" rx="2" />
+      <rect x="83" y="20" width="12" height="62" rx="2" />
+    </svg>
+  );
+}
+
 const PROVIDER_ICON_MAP = {
   CLAUDE: ClaudeIcon,
   CODEBUDDY: CodeBuddyIcon,
@@ -172,6 +182,7 @@ const PROVIDER_ICON_MAP = {
   KIRO: KiroIcon,
   OPENCODE: OpenCodeIcon,
   OMP: OmpIcon,
+  PI: PiIcon,
 };
 
 // Multi-color brand SVG assets in /public/brand-logos/. Only logos that have
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "tokentracker-cli",
   "version": "0.6.7",
-  "description": "Token usage tracker for AI agent CLIs (Claude Code, Codex, Cursor, Gemini, Kiro, OpenCode, OpenClaw, Every Code, Hermes, GitHub Copilot, Kimi Code, CodeBuddy, oh-my-pi, Craft Agents)",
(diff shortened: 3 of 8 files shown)


mm7894215:
Thanks for the PR. I agree the new `pi` reader itself makes sense: standard Pi and oh-my-pi are different tools, and standard Pi sessions live under `~/.pi/agent/sessions/` while our existing `omp` reader is for `~/.omp/agent/sessions/`.

One compatibility issue to fix before merge: this PR removes `PI_CODING_AGENT_DIR` support from the `omp` resolver/init path, but that env var is not exclusive to standard Pi. I checked upstream docs for both tools:

- Pi documents `PI_CODING_AGENT_DIR` as its config/agent directory override, defaulting to `~/.pi/agent`.
- oh-my-pi also documents `PI_CODING_AGENT_DIR` as the full agent directory override, defaulting to `~/<PI_CONFIG_DIR or .omp>/agent`, and stores sessions under that agent dir.

So removing this from `resolveOmpAgentDir()` would regress oh-my-pi users who set `PI_CODING_AGENT_DIR` to relocate their `omp` agent data: after upgrading TokenTracker, their existing `omp` sessions would no longer be discovered.

I think we should keep the new `pi` provider, but handle the overlap explicitly instead of making `PI_CODING_AGENT_DIR` belong only to `pi`. A few acceptable options:

1. Keep `PI_CODING_AGENT_DIR` in `resolveOmpAgentDir()` for backward compatibility, and only scan the new `pi` default `~/.pi/agent` unless a pi-specific signal exists.
2. If both readers would resolve to the same agent dir, scan it once and choose the source based on the detected install/CLI context.
3. Add an explicit TokenTracker override/env for disambiguation, so users with custom dirs can opt into `pi` vs `omp` without double counting.

The important invariant is: adding standard Pi support must not break existing oh-my-pi users with custom `PI_CODING_AGENT_DIR`, and it must not scan the same session directory twice under both `source: "omp"` and `source: "pi"`.

season179:
Good catch — you're right that omp documents PI_CODING_AGENT_DIR too. Pushed a fix that combines bits of all three of your options.

Both resolvers honor the env var, but a small disambiguator routes it to whichever tool is installed: if ~/.pi/ exists (and is actually a directory), pi owns it; otherwise omp keeps it, so existing oh-my-pi users stay working without changing anything. For the both-installed case I added TOKENTRACKER_OMP_AGENT_DIR / TOKENTRACKER_PI_AGENT_DIR as explicit per-tool overrides.

For invariant 2, piAgentDirCollidesWithOmp(env) is a belt-and-suspenders check in sync.js: if the two resolvers ever land on the same canonical path, the pi pass is skipped so the same sessions can't enqueue twice. status and init are gated on the same helper so they don't claim "pi: detected" for a dir that won't actually be scanned.

Tests cover both routing directions, the stray-file-at-~/.pi case, the TT overrides, and the collision detector. Happy to flip the default if you'd rather require an explicit override.

mm7894215:
Squashed into main as 35744a31; releasing as v0.8.0.

mm7894215:
Thanks for the thorough fix on the env-var routing — the install-signal disambiguator + explicit `TOKENTRACKER_*` overrides + collision check in sync.js cover all three invariants cleanly, and the test matrix is great.

Squashed into main as 35744a31 with a MINOR bump to v0.8.0 (new provider). npm publish + DMG workflow both running.

Two tiny follow-up ideas (non-blocking, didn't want to hold the merge):
- `piAgentDirCollidesWithOmp` uses `path.resolve` rather than `fs.realpathSync` — symlinks pointing at the same physical dir would still slip past. Edge case, but might be worth tightening if anyone ever reports it.
- `PROVIDER_COLORS` in `UsageOverview.jsx` doesn't have an entry for PI yet (it falls through to the index-based hsl fallback), but neither do OMP/Hermes/CodeBuddy etc., so this is pre-existing.

Closing the PR + #56 now. Appreciate the contribution!

PR #58, opened 2026-05-08
fix(menubar): prevent updateStatsDisplay recursion on macOS 26

sailesh:
macOS 26 introduces a synchronous layout callback path (NSStatusBarButtonCell.setTitle: → NSStatusItem._adjustLength) that re-enters StatusBarController.updateStatsDisplay() while it's setting the status item's image or length. The re-entrant call hits the else branch → animator?.applyCurrentState() → setButtonImage() → fires onImageUpdated → updateStatsDisplay() again, causing infinite recursion and a stack-overflow crash (EXC_BAD_ACCESS at the stack guard region, ~7700 frames deep).

The recursion was always coded but the macOS layout callback that triggers re-entry is new in macOS 26; the bug was latent in v0.6.7.

Fix: guard updateStatsDisplay() with an isUpdatingDisplay flag so any re-entrant call returns immediately. defer resets the flag so normal publisher-driven updates still work.

## Test plan
- [x] Build via `xcodebuild -scheme TokenTrackerBar -configuration Release` — succeeds
- [x] Launch on macOS 26 — app stays running, no recursion crash
- [x] Change menu bar slot settings while running — no crash, slots update correctly
- [x] Change configured providers — no crash
- [ ] Reviewer: smoke-test on macOS ≤ 15 to confirm no regression in normal update flow

# PR Goal (one sentence)
Fixes a stack-overflow crash on macOS 26.

## Scope

-

- **Delta since last Codex review:** (commits or summary)
- **Edge cases covered:**
- **Tests run (command + result):**

### Rules / Invariants

-

-

### Evidence (tests or repro)

-

- [ ] Mark N/A if no public exposure

-

- [ ]

### Uncovered scope

-

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Services/StatusBarController.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Services/StatusBarController.swift
@@ -142,6 +142,7 @@ final class StatusBarController: NSObject {
     private let menuBarHeight: CGFloat = 22
     private let menuBarIconSize = NSSize(width: 22, height: 22)
     private let emptyAttributedTitle = NSAttributedString(string: "")
+    private var isUpdatingDisplay = false
 
     private static let showStatsKey = "MenuBarShowStats"
     private var showStats: Bool {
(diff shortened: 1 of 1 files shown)


mm7894215:
Thanks for the precise diagnosis and minimal fix — root cause analysis (synchronous _adjustLength re-entry on macOS 26) was spot on.

Squash-merged into main with a version bump to v0.8.1. npm publish workflow has been triggered, and macOS DMG build is in progress: https://github.com/mm7894215/TokenTracker/actions/runs/25547285488

Released as: https://github.com/mm7894215/TokenTracker/releases/tag/v0.8.1 (will populate once the DMG workflow completes)

PR #59, opened 2026-05-08
feat(menubar): combine limit utilization and reset countdown in slots

sailesh:
## Summary
- Menu bar limit slots now show both utilization percentage AND time until reset in a single slot (e.g. `Cl 5h 72% · 1h45m`), so users can see at a glance both how much quota is used and when it will reset without opening the popover.
- Extended the combined format to every existing provider window: Cursor (Plan/Auto/API), Gemini (Pro/Flash/Lite), Kimi (Weekly/5h/Total), Kiro (Monthly/Bonus), Copilot (Premium/Chat), and Antigravity (Claude/Gemini Pro/Flash). 25 limit slots total under the "Limits" category.
- Reuses the existing `relativeReset(iso:)` / `relativeReset(epoch:)` parsing pattern from `UsageLimitsView`; adds a compact menu-bar variant (`Xd` / `Xh` / `Xm`) and a `genericLimitValue()` helper that handles every provider using `GenericLimitWindow` (ISO reset). Codex keeps a dedicated formatter because it uses Int `usedPercent` + Unix epoch resets.

## Why
The popover already shows reset countdowns prominently, but the menu bar (the at-a-glance surface) showed only the percentage. Users plan their next coding session around when their quota resets — combining both in one slot is more useful than picking just one. Providers all share the same data shape (`utilization` + `resetsAt`/`reset_at`) so the format applies uniformly.

## Test plan
- [x] Build via `xcodebuild -scheme TokenTrackerBar -configuration Release` — succeeds
- [x] Launch on macOS 26 — app stays running (recursion fix from #58 already in main)
- [x] Menu bar settings dropdown lists all new slots under "Limits" category
- [x] `Claude 5h Limit` displays as `Cl 5h <pct>% · <countdown>`
- [x] `Codex 7d Limit` displays as `Cx 7d <pct>% · <countdown>` (Unix epoch reset format path)
- [ ] Slots for unconfigured providers (Kimi, Copilot, etc.) hide gracefully — needs reviewer with those providers configured to verify
- [x] `npm test` — no new failures vs `main`
- [x] `npm run validate:ui-hardcode` / `npm run validate:copy` / `npm run validate:guardrails` — all clean

## Notes for reviewer
- Branch was rebased onto current `main` (post-#58, post-v0.8.1) — clean fast-forward, no conflicts.
- New `pi-coding-agent` provider (#57) is parser-only and doesn't expose a `usage-limits` endpoint yet, so it intentionally has no slot in this PR. Easy to add later by extending the same `genericLimitValue()` pattern if/when its limits are surfaced.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Services/StatusBarController.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Services/StatusBarController.swift
@@ -12,6 +12,22 @@ enum MenuBarDisplayMetric: String, CaseIterable {
     case claude7d
     case codex5h
     case codex7d
+    case cursorPlan
+    case cursorAuto
+    case cursorAPI
+    case geminiPro
+    case geminiFlash
+    case geminiLite
+    case kimiWeekly
+    case kimi5h
+    case kimiTotal
+    case kiroMonth
+    case kiroBonus
+    case copilotPremium
+    case copilotChat
+    case antigravityClaude
+    case antigravityGPro
+    case antigravityFlash
 
     var menuLabel: String {
         switch self {
@@ -24,6 +40,22 @@ enum MenuBarDisplayMetric: String, CaseIterable {
         case .claude7d: return "Cl 7d"
         case .codex5h: return "Cx 5h"
         case .codex7d: return "Cx 7d"
+        case .cursorPlan: return "Cu Plan"
+        case .cursorAuto: return "Cu Auto"
+        case .cursorAPI: return "Cu API"
+        case .geminiPro: return "Gm Pro"
+        case .geminiFlash: return "Gm Flash"
+        case .geminiLite: return "Gm Lite"
+        case .kimiWeekly: return "Km Wk"
+        case .kimi5h: return "Km 5h"
+        case .kimiTotal: return "Km Tot"
+        case .kiroMonth: return "Kr Mo"
+        case .kiroBonus: return "Kr Bn"
+        case .copilotPremium: return "Co Prem"
+        case .copilotChat: return "Co Chat"
+        case .antigravityClaude: return "Ag Cl"
+        case .antigravityGPro: return "Ag GPro"
+        case .antigravityFlash: return "Ag Fl"
         }
     }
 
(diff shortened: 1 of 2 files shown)


sailesh:
Normal
image

During rollover
image


After rollover
image

FYI

mm7894215:
## Test plan: unconfigured-slot hiding — VERIFIED ✅

Validated locally on the PR branch (built from `dbc2f1be`). Used four test machines' worth of provider state in one go via the local `/functions/tokentracker-usage-limits` snapshot:

```
claude         configured=true  error=null
codex          configured=true  error=null
cursor         configured=true  error=null
gemini         configured=true  error=null
kimi           configured=true  error="Kimi token refresh failed (HTTP 400)"
copilot        configured=true  error="GitHub Copilot token rejected"
kiro           configured=false error=null
antigravity    configured=false error=null
```

Both hidden paths in `genericLimitValue` got exercised: the `configured == false` guard (kiro/antigravity) and the `error == nil` guard (kimi/copilot).

Measured by the menu-bar status item's icon X coordinate via Accessibility (smaller x = wider rendered area):

| Selection | State | icon.position.x | Rendered slots |
|---|---|---|---|
| `claude5h + claude7d` | 2× valid | **1187** | 2 |
| `claude5h + kiroMonth` | 1 valid + 1 unconfigured | **1242** | **1** |
| `kiro + antigravity` | 2× configured=false | **1290** | 0 |
| `kimi + copilot` | 2× configured=true but error≠nil | **1290** | 0 |

Per-slot delta is ~50px and additive, and both "hidden" rows collapse to identical width (1290) regardless of whether the cause is unconfigured or errored. Mixed selection seamlessly drops the hidden one without leaving an empty separator. ✅

> [x] Slots for unconfigured providers (Kimi, Copilot, etc.) hide gracefully

The remaining unchecked item from the test plan is now verified.

mm7894215:
Merged into main and shipped as **v0.9.0** 🎉 Thanks @sailesh — this is a really thoughtful UX addition (countdown next to percentage is exactly the data users were going to the popover for) and the codepath was clean to extend.

A few notes for the release notes / future readers:

**What shipped from #59:**
- Reset countdown next to utilization on every limit slot
- 16 new provider slots: Cursor (Plan/Auto/API), Gemini (Pro/Flash/Lite), Kimi (Wk/5h/Tot), Kiro (Mo/Bn), Copilot (Prem/Chat), Antigravity (Cl/GPro/Fl)

**Follow-up I added on top before shipping** ([8ed80dc5](https://github.com/mm7894215/TokenTracker/commit/8ed80dc5)):
- Filter the slot dropdown by configured-provider availability — `MenuBarDisplayPreferences.availableItemsPayload(for:)` now takes the live `UsageLimitsResponse` and hides limit slots whose provider is unconfigured or erroring. Users with only Claude/Codex no longer see 16 unreachable options. Selected ids are preserved through transient errors (the underlying `compactMap` in `buildMenuBarDisplayValues` already hides the slot's value).
- `NativeBridge` subscribes to `viewModel.$usageLimits` via Combine and re-pushes settings whenever provider availability flips, so the dropdown updates live without a page reload. The fingerprint collapses utilization/reset noise to one bool per provider so we don't re-push on every poll.

**Verification:**
- Independent Swift logic test against my live `/usage-limits` snapshot: 25 → 15 dropdown items (filtered out kimi×3, kiro×2, copilot×2, antigravity×3 — all of which were either unconfigured or erroring on my machine).
- Slot rendering filter (the original `genericLimitValue` guard) verified earlier via menu-bar accessibility position deltas.

Thanks again — clean PR, easy review.

PR #74, opened 2026-05-15
feat: add Grok Build usage tracking

Yanzzp999:
## Summary
- Add Grok Build integration via SessionEnd hook and passive signals.json scanning
- Queue Grok session usage under source="grok"
- Add status, diagnostics, uninstall, README, pricing, and parser coverage
- Add regression tests for hook install, sync signal handling, diagnostics, uninstall, and parser behavior

## Notes
Grok Build currently exposes aggregate session usage via contextTokensUsed, so this integration records total_tokens. Input/output/cache breakdown can be refined later if Grok exposes per-call telemetry.

## Test
- npm test
- npm run validate:copy
- npm run validate:ui-hardcode
- npm run validate:guardrails

## Summary by CodeRabbit

* **New Features**
  * Added Grok Build as a supported AI tool with automatic usage tracking
  * Integrated Grok Build status reporting and synchronization
  * Added pricing data for Grok models

* **Documentation**
  * Updated README to reflect Grok Build as a newly supported AI tool (15 → 16 tools)

[Review Change Stack](https://app.coderabbit.ai/change-stack/mm7894215/TokenTracker/pull/74)


mm7894215:
Hi @Yanzzp999, thanks a lot for putting this together — the integration scaffolding (hook install / probe / uninstall / diagnostics) is solid, the test coverage on install paths and lock contention is great, and `TOKENTRACKER_GROK_HOME` precedence + shell-quote handling are exactly aligned with how other providers do it.

Two parts to this comment: housekeeping first, then the substantive issues so you have everything you need to iterate.

---

## Housekeeping

**1. Rebase onto the latest `main`**
The branch looks like it forked from a slightly older `main`, and the `dashboard/yarn.lock` / `dashboard/package-lock.json` diff (~400 lines removed, `peer: true` fields wiped across many entries) seems to be a side-effect of that rather than an intentional change. After rebasing, please run `npm install` / `yarn install` on a clean checkout so the lockfiles only reflect changes you actually intend.

**2. Split the PR**
Currently mixes three unrelated changes:
- **Grok Build integration** — the core feature ✅
- **`ProjectUsagePanel` clickable-card refactor** — useful, but unrelated; deserves its own PR
- **Deleted `RollingUsagePanel.test.jsx` / `UsagePanel.test.jsx`** — dangling test cleanup (components are already gone on `main`); fine as a separate small `chore:` commit

---

## Substantive issues on the Grok integration

### 🔴 Critical — Grok cost will always render as \$0

`computeRowCost` in `src/lib/pricing/index.js:108-115` only reads `input_tokens` / `output_tokens` / `cached_input_tokens` / `cache_creation_input_tokens` / `reasoning_output_tokens` — it **never reads `total_tokens`**.

Both write paths for Grok set `input_tokens: 0, output_tokens: 0` and only populate `total_tokens`:
- `src/lib/rollout.js:6072-6082` (passive scan)
- `src/commands/sync.js:670-681` (hook signal handler)

Net effect: the new `grok-build` / `grok-4.3` / `grok-4.20` / `grok-4-fast` entries in `curated-overrides.json` are dead code, and the dashboard / menu bar / Cost Analysis modal will show **\$0 cost for every Grok session**. Tokens display fine (`resolveDisplayTokens` falls back to `total_tokens`), but cost is silently zero.

Two possible fixes:
- (a) Distribute `contextTokensUsed` across input/output using an assumed ratio (e.g. 80/20) and write real values into `input_tokens` / `output_tokens`
- (b) Extend `computeRowCost` to fall back to `total_tokens * blended_rate` when input+output are both zero

Whichever you pick, please add an end-to-end test that asserts a Grok row produces a non-zero `total_cost_usd` via `/functions/tokentracker-usage-model-breakdown`. That test would have caught this.

### 🟠 High — `contextTokensUsed` semantics need evidence

The parser comment at `src/lib/rollout.js:5982-5983` treats `contextTokensUsed` as "aggregate session usage", but the field name strongly suggests **current context-window occupancy** (a snapshot), not cumulative tokens consumed. Many CLIs use this exact name for the "Context X / Y" indicator.

If it's a snapshot:
- Single sessions with many turns get compressed into one final context value → severe under-counting
- We have prior history with this exact failure mode on Codex (the \`input\` field including cached tokens caused 6–7× cost inflation before we caught it)

Could you attach:
- A real Grok session log + the final `signals.json` so we can see what `contextTokensUsed` evolves to over a session
- Ideally a comparison against xAI's billing dashboard for the same session

If it does turn out to be a snapshot, the right fix is probably to parse `summary.json` or wait for per-call telemetry; at minimum the README should mark Grok numbers as estimates.

### 🟠 High — Long sessions get locked to their first snapshot

The `seenSessions` Set in `cursors.grok.seenSessions` is treated as a permanent "we already counted this" marker. But:

1. Long session is in progress, passive scan reads `signals.json` mid-flight (say `contextTokensUsed=10k`) → record enqueued, sessionId added to `seenSessions`
2. Session continues to completion, `contextTokensUsed` grows to 50k, SessionEnd hook writes a fresh signal file with the final value
3. Next sync hits `sync.js:657-660`: `seenSessions.has(hookSessionId)` is true → signal marked consumed → **signal file deleted, final 50k value silently dropped**

Suggested fix: track `(sessionId, lastValue)` so subsequent observations enqueue a delta, or let the hook signal always overwrite the previous snapshot value for the same session.

### 🟡 Medium — `appendRolloutRecords` re-reads the entire queue every call

`src/lib/rollout.js:1434-1513` reads `queue.jsonl` into memory and rebuilds `latestByKey` on every invocation. Every other provider goes through `enqueueTouchedBuckets` + `hourlyState`, which avoids the rescan. With a few months of data this becomes a multi-MB read every sync.

Suggest reusing the existing path. If there's something about Grok that makes it incompatible, happy to discuss — but I don't think there is.

### Nits

- `grok-4.3`, `grok-4.20`, `grok-4-fast` in `curated-overrides.json` look like placeholder version numbers (4.20 in particular doesn't match any xAI release). Please align with real xAI model IDs or drop them and keep only `grok-build`.
- `appendRolloutRecords(queuePath, [rec], { source: \"grok\" })` — the third argument is ignored by the function signature, can be removed.
- `normalizeIsoHalfHourStart` in `rollout.js` is just an alias for `toUtcHalfHourStart` — drop the alias.

---

Thanks again for the work — really appreciate the depth on the install/uninstall side. Once the critical cost bug and the `contextTokensUsed` question are sorted, the rest is mostly polish.

Yanzzp999:
Thanks for the careful review. I pushed a follow-up that keeps the PR focused on Grok Build, rebases onto latest `main`, and fixes the main usage-accounting issues: Grok now records estimated input/output tokens, tracks session deltas instead of first snapshots, and has regression coverage for pricing and malformed signals.

`npm test` passes locally.

mm7894215:
Thanks @Yanzzp999 for the careful follow-ups! Re-verified all four review items end-to-end:

- ✅ **Cost ≠ \$0** — `estimateGrokTokenDelta()` writes real `input_tokens`/`output_tokens` via the 80/20 split, so `computeRowCost` now produces non-zero cost. Pricing entries are no longer dead code.
- ✅ **Long sessions** — `sessionSnapshots` per-session ratchet + delta tracking handles the "scan mid-session then again at end" case cleanly. Nice test coverage at `rollout-parser.test.js:5540`.
- ✅ **Queue path** — switched to `enqueueTouchedBuckets`, no more full queue rescan.
- ✅ **Polish** — real xAI model IDs in `curated-overrides.json`, README & pricing JSON both transparently mark Grok numbers as estimates while `signals.json` is snapshot-based, scope cleaned up.
- ✅ `npm test`: 520/520 pass, `validate:guardrails` clean, `audit:tokens` introduces no new invariant failures.

Merging now and cutting v0.15.0 (new provider → minor bump). Really appreciate the quick iteration and the new test coverage. 🙏

(Follow-up worth tracking but not blocking: when xAI exposes per-call telemetry in `signals.json` we can drop the 80/20 estimate. And Grok Build usage limits — `console.x.ai` has a quota API — would be a nice future addition alongside the existing 7-provider Usage Limits panel.)

PR #77, opened 2026-05-17
Fix menu bar zero-state stats and footer hit areas

fflake33:
## Summary

This PR fixes two small issues I noticed during day-to-day use of the macOS menu bar app.

- Align the footer button click targets with the area that visually reacts to hover.
- Keep menu bar stats visible when usage summaries are successfully loaded but their token values are `0`.

## Details

The footer buttons previously animated over a larger area than the actual clickable button content. That made the UI feel like a click had been ignored when the pointer was inside the active hover area but outside the hit target. Moving the frame and content shape into the button label makes the clickable region match the animated region.

The menu bar stats previously used positive token counts as the visibility gate. Around a day boundary, before any token usage had happened, the app could successfully fetch zero-valued summaries but still hide the daily and rolling stats. The status item width could remain allocated while the values disappeared. The visibility checks now use summary availability instead of `> 0` token values, so a known zero is rendered as `0` while missing data remains hidden.

## Validation

- Ran `xcodebuild -project TokenTrackerBar/TokenTrackerBar.xcodeproj -scheme TokenTrackerBar -configuration Debug -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO build`.
- Verified the zero-summary behavior with an isolated local mock server on port `17680`, separate from the real TokenTracker service/data, returning all-zero summary payloads.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced menu bar statistics display to show content based on actual data availability rather than numerical thresholds.

* **Style**
  * Refined footer button styling and layout for improved visual alignment and consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mm7894215/TokenTracker/pull/77?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Services/StatusBarController.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Services/StatusBarController.swift
@@ -316,7 +316,7 @@ final class StatusBarController: NSObject {
 
         animator = MenuBarAnimator(button: button)
         animator?.onImageUpdated = { [weak self] _ in
-            guard let self, self.showStats, self.viewModel.todayTokens > 0 else { return }
+            guard let self, self.showStats, !self.buildMenuBarDisplayValues().isEmpty else { return }
             self.updateStatsDisplay()
         }
         updateStatsDisplay()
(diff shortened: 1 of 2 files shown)


mm7894215:
Thanks @fflake33 — clean fix for both the zero-state visibility and the footer hit area. Merged.

PR #92, opened 2026-05-22
fix: add node sqlite fallback for provider DB readers

ZhangTainan:
# PR Goal (one sentence)

Centralize SQLite reads with a shared helper that falls back to node:sqlite when sqlite3 CLI is missing, and reuse it across Cursor/OpenCode/Kiro/Hermes/Kiro CLI/Zed/Goose with preserved snapshot behavior and targeted tests.

## Scope

- Add shared SQLite reader with CLI-first + node:sqlite fallback and deduped actionable warning.
- Reuse helper for Cursor token extraction plus OpenCode, Kiro, Hermes, Kiro CLI, Zed, and Goose DB reads, keeping provider-specific timeouts/buffers and snapshot handling.
- Add focused tests for the helper, Cursor extraction, OpenCode fallback, and Windows-safe snapshot paths.

- **Delta since last Codex review:** One commit: fix: add node sqlite fallback for provider DB readers (shared helper + parser updates + tests).
  - Prefer sqlite3 -json when available; fall back to node:sqlite only if CLI missing.
  - Preserve per-provider timeouts/buffers and existing snapshot handling.
  - Emit a single actionable warning only when both readers are unavailable; stay quiet for query/schema errors.
- **Edge cases covered:**
  - Missing sqlite3 binary with node:sqlite available (fallback path).
  - Both readers unavailable (deduped warning).
  - Windows-safe UNC/snapshot path handling.
- **Tests run (command + result):** Not run locally (tests added/updated).
- **Known gaps / out of scope:** No changes to pricing/aggregation semantics; no new providers.

### Rules / Invariants

- Use sqlite3 -json if present; only warn when CLI is missing and node:sqlite is unavailable.
- Do not alter provider-specific timeouts, buffers, or snapshot flow.
- Keep failures silent for schema/query errors to avoid noisy logs.

- CLI present, node:sqlite missing -> CLI path used.
- CLI missing, node:sqlite present -> fallback path used.
- Both missing -> deduped warning + empty result.

### Evidence (tests or repro)

- Added unit tests for SQLite helper fallback/warnings and Cursor/OpenCode paths (not executed locally).

- [x] Mark N/A if no public exposure

- Provider SQLite readers returning empty results on machines without sqlite3.

Tests run:
- node --test test/sqlite-reader.test.js ✅
- node --test test/cursor-config.test.js ✅
- node --test test/rollout-parser.test.js ⚠️ partial pass

Known remaining failures:
- Legacy rollout-parser tests still create fixture DBs via sqlite3 CLI.

### Uncovered scope

- Full npm test suite not executed.

## Release Notes

* **Refactor**
  * Enhanced database access mechanisms with improved fallback strategies to ensure consistent behavior across platforms and handle edge cases more gracefully.
  * Optimized session token extraction with better cross-platform path handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mm7894215/TokenTracker/pull/92?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/cursor-config.js
+++ b/src/lib/cursor-config.js
@@ -1,33 +1,34 @@
 const os = require("node:os");
 const path = require("node:path");
 const fs = require("node:fs");
-const cp = require("node:child_process");
 const https = require("node:https");
 
 const { readJson } = require("./fs");
+const { readSqliteFirstValue } = require("./sqlite-reader");
 
 // ── Path resolution ──
 
 function resolveCursorPaths({ home, platform = process.platform, env = process.env } = {}) {
   const h = home || os.homedir();
+  const pathForPlatform = platform === "win32" ? path.win32 : path.posix;
   let appDir;
   if (platform === "darwin") {
-    appDir = path.join(h, "Library", "Application Support", "Cursor");
+    appDir = pathForPlatform.join(h, "Library", "Application Support", "Cursor");
   } else if (platform === "win32") {
     const appData =
       (typeof env.APPDATA === "string" && env.APPDATA.trim()) ||
-      path.join(h, "AppData", "Roaming");
-    appDir = path.join(appData, "Cursor");
+      pathForPlatform.join(h, "AppData", "Roaming");
+    appDir = pathForPlatform.join(appData, "Cursor");
   } else {
     const xdg =
       (typeof env.XDG_CONFIG_HOME === "string" && env.XDG_CONFIG_HOME.trim()) ||
-      path.join(h, ".config");
-    appDir = path.join(xdg, "Cursor");
+      pathForPlatform.join(h, ".config");
+    appDir = pathForPlatform.join(xdg, "Cursor");
   }
   return {
     appDir,
-    stateDbPath: path.join(appDir, "User", "globalStorage", "state.vscdb"),
-    cliConfigPath: path.join(h, ".cursor", "cli-config.json"),
+    stateDbPath: pathForPlatform.join(appDir, "User", "globalStorage", "state.vscdb"),
+    cliConfigPath: pathForPlatform.join(h, ".cursor", "cli-config.json"),
   };
 }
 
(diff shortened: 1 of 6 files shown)


mm7894215:
Thanks for the cleanup — the shared helper is a clear win and removes a lot of duplicated boilerplate. Tested locally on this branch:

- `test/sqlite-reader.test.js` 6/6 ✓
- `test/cursor-config.test.js` 37/37 ✓
- `test/rollout-parser.test.js` 141/141 ✓ (the "partial pass" note in the PR body looks conservative — passes cleanly with `sqlite3` on PATH)

A few blocking + nice-to-have items before merge. CodeRabbit already flagged 1–3; adding 4–6 from my own pass:

### Blocking

**1. Kiro DB read failure silently eats JSONL fallback (data loss).** Per CodeRabbit's `rollout.js:2876-2887` finding: `readKiroDbTokens()` returning `[]` now collides with two distinct cases — "DB read succeeded, no new rows" vs "no SQLite reader available at all". The current code treats both as success, sets `usingDb = true`, skips JSONL, and can advance `nextJsonlLine` to the file tail. On a machine without `sqlite3` CLI **and** without `node:sqlite` (Node <22), Kiro users would permanently lose those JSONL rows on subsequent runs.

Fix: make `readSqliteJsonRows` return `{ rows, ok }` (or throw on unavailable readers), and only suppress JSONL when `ok === true`.

**2. Old `sqlite3` CLI (<3.33) emits `Error: unknown option: -json`.** `isSqliteCliUnavailable()` only catches ENOENT-style errors, so older binaries trigger neither the `node:sqlite` fallback's "unavailable" warning path nor any other diagnostic. Add `unknown option: -json` / `unknown flag: -json` to that predicate so old binaries flow through to the fallback + warning.

**3. `extractCursorSessionToken` doesn't thread `env` into the DB read.** `resolveCursorPaths` now honors the new `env` param, but `fs.existsSync` debug log + `readCursorAccessTokenFromStateDb` still read `deps?.env`. Either the path resolution and the read should both see the test env, or neither — current state is inconsistent.

### Nice to have

**4. Warning shows snapshot temp path for Hermes/Zed/Goose.** Those callers pass `snap.path` (a random temp file from `snapshotSqliteDb`) into the helper. When both readers fail, the warning prints `Path: /tmp/sqlite-snapshot-xxx.db` — useless for the user. Worse, the dedup key `${label}:${dbPath}` keys off the temp path, so every sync produces a fresh warning. Suggest threading an `originalDbPath` (or `displayPath`) through `sqliteOptions` and using it for both the warning text and the dedup key.

**5. `TOKENTRACKER_DEBUG=1` no longer surfaces query/schema errors.** Old `readCursorAccessTokenFromStateDb` called `cursorDebugLog` for any failure; the new helper only emits anything when both errors look like "unavailable". The PR description says this is intentional, but consider keeping a debug-gated stderr line for non-availability errors so a corrupt DB / lock / schema drift is still diagnosable behind the flag.

**6. `readSqliteFirstValue` silently widens contract.** Old code returned `null` for non-string `row.value`; new code does `String(value).trim()`. No production impact for Cursor (TEXT column), but the contract change is implicit — worth either matching the old null-return or documenting the new behavior in a comment.

Happy to take another look once 1–3 land.

mm7894215:
Superseded by #96 (which carries the same SQLite fallback + addresses Cursor env threading + ships the device-token rotate-on-issue fix in one go).

PR #98, opened 2026-05-24
fix: allow configured local dashboard hosts

Fryuni:
## Summary
- add `tokentracker serve --allowed-hosts` for local dashboard deployments reached through private/internal DNS names
- expose the configured host allowlist through a same-origin `/api/dashboard-config` endpoint before the dashboard creates the InsForge client
- use the configured allowlist for local auth proxy selection and auth-bridge origin checks

## Use case
I run multiple machines, each with multiple agents, and expose each TokenTracker service on the network behind internal DNS instead of setting up port forwarding every time. Those internal hostnames are still local dashboard proxy origins, but the dashboard previously treated them as public deployments and sent auth refresh requests directly to InsForge, which could fail with 401 because the local cookie relay was bypassed.

## Tests
- `node --test test/serve-allowed-hosts.test.js test/local-api-security.test.js test/serve-bind-host.test.js`
- `npm run test -- src/lib/__tests__/insforge-config.test.ts src/contexts/__tests__/InsforgeAuthContext.test.jsx`
- `npm --prefix dashboard run typecheck`

Note: a full `npm test` run still has an unrelated existing failure in `test/sync-cursor-unknown-migration.test.js` around Grok history repair; the focused serve/auth tests above pass.

## Summary by CodeRabbit

* **New Features**
  * Added `--allowed-hosts` CLI option and updated serve help text.
  * Dashboard now loads allowed-host config and resolves service base URLs using the current location.
  * Local API security enhanced with hostname allowlist checks and a dashboard-config endpoint.

* **Tests**
  * Added tests for allowed-hosts parsing, dashboard config endpoint, and local API authorization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mm7894215/TokenTracker/pull/98?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


mm7894215 (changes requested):
Thanks for the PR. The server-side allowlist and the auth proxy config path are headed in the right direction, but I do not think this is ready to merge yet.

Blocking issue: `--allowed-hosts` only affects the InsForge SDK base URL. The rest of the dashboard still treats only `localhost` and `127.0.0.1` as local mode. For example, `App.jsx` computes `isLocalMode` from those two hostnames only, and `DashboardPage.jsx` plus the usage/model/project/trend/heatmap hooks repeat the same local-host check.

That means the main use case in the PR can still break: when the local CLI dashboard is opened through a configured internal DNS name, auth may now use the same-origin proxy, but the app still behaves like a public deployment. In particular, unauthenticated local access can hit the auth gate, and the data hooks can early-return because they do not consider the configured host local.

Please make the configured allowed-host state part of the shared local-dashboard-host decision, then use that helper consistently across `App.jsx`, `DashboardPage.jsx`, the local data hooks, cloud sync/settings host checks, and `host-mode.ts`. The current implementation fixes one auth path but leaves the broader dashboard local-mode behavior inconsistent with the new CLI option.

Validation I ran:
- `node --test test/serve-allowed-hosts.test.js test/local-api-security.test.js test/serve-bind-host.test.js`
- `npm run test -- src/lib/__tests__/insforge-config.test.ts src/contexts/__tests__/InsforgeAuthContext.test.jsx`
- `npm --prefix dashboard run typecheck`

I did not run the full `npm test`; the PR description says it currently has an unrelated existing Grok migration failure.

Fryuni:
It's not ready to merge... But you merged it?!

mm7894215:
Hey @Fryuni — got word that this wasn't quite ready yet, so I've reverted it from `main` (commits 0b958488 + 178b2029). Sorry for the rush on the merge.

GitHub won't let me re-open a merged PR, so whenever you're back at it, just rebase your branch onto the new `main` and open a fresh PR — happy to keep the context linked to #98. Appreciate the patch.

PR #118, opened 2026-05-31
ci(release-windows): upload version-less assets only, matching the macOS DMG

Hasasasa:
## Problem

The Windows release attaches each artifact **twice** — a versioned name and a version-less alias — so the release page shows four Windows assets where macOS shows one:

| | |
|---|---|
| `TokenTracker-Setup-v0.35.0.exe` | `TokenTracker-Setup.exe` |
| `TokenTracker-win-x64-v0.35.0.zip` | `TokenTracker-win-x64.zip` |

(identical sha256 — pure duplicates)

The version-less alias is **required**: the landing page and READMEs deep-link `releases/latest/download/TokenTracker-Setup.exe`, which can't carry a version in the URL. The versioned copy was only "nicer when browsing", but the release heading already shows the version.

## Fix

Keep building with versioned filenames on the runner, but **upload only the version-less names** (copied from the versioned build artifacts). Each Windows release now ships exactly one installer + one zip, mirroring the single version-less `TokenTrackerBar.dmg`.

Workflow tests updated to assert the `gh release upload/create` lines reference the stable `$zip`/`$setup` aliases and never the versioned build artifacts. `npm test` for both `release-windows-workflow.test.js` (27) and `release-dmg-workflow.test.js` (17) pass.

> Note: this fixes future releases. The existing v0.35.0 duplicates would need a one-time manual cleanup on the release page.

## Summary by CodeRabbit

* **Chores**
  * Streamlined Windows release asset uploads to provide cleaner downloads with stable, version-less filenames.

### DIFF
--- a/.github/workflows/release-windows.yml
+++ b/.github/workflows/release-windows.yml
@@ -147,18 +147,23 @@ jobs:
           VERSION: ${{ inputs.version }}
         run: |
           $tag = "v$env:VERSION"
-          $zip = "TokenTracker-win-x64-v$env:VERSION.zip"
-          $setup = "TokenTrackerWin/installer/Output/TokenTracker-Setup-v$env:VERSION.exe"
-          # Stable, version-less aliases so the landing page can deep-link
-          # releases/latest/download/<name> without knowing the version. The
-          # versioned names stay too (clearer when browsing the release page).
-          $setupStable = "TokenTracker-Setup.exe"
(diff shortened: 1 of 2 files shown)


mm7894215:
Merged 🎉 Thanks Phil! Confirmed the README/landing deep-links rely on the version-less names, so dropping the duplicate versioned assets is safe. Release-windows + release-dmg workflow tests (44) green.

PR #120, opened 2026-05-31
fix(mac): update settings sync state

fflake33:
## Summary

- push native settings updates when the macOS dashboard view model enters or leaves manual sync
- add a regression test that keeps the settings payload and `isSyncing` publisher wiring in place

## Why

The Settings page already renders the "Sync now" button from the native `isSyncing` setting, and the native bridge already includes that field in its payload. However, the bridge only pushed settings when provider availability changed. Pressing "Sync now" could therefore start a sync without sending a fresh settings event to the WebView, making the button feel inert.

## How

`NativeBridge.configure` now subscribes to `DashboardViewModel.$isSyncing`, de-duplicates changes, and reuses the existing `pushSettings()` path. This keeps the fix scoped to UI state propagation; sync execution, server APIs, token accounting, and dashboard routing are unchanged.

## Validation

- `node --test test/native-bridge-sync-feedback.test.js`
- `npm test`
- Manual macOS app smoke test: built and opened a local TokenTrackerBar bundle, verified the native Settings page shows sync feedback after pressing "Sync now".

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dashboard UI sync state feedback. The interface now correctly updates to reflect changes in syncing availability, ensuring users see accurate, real-time sync status.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Services/NativeBridge.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Services/NativeBridge.swift
@@ -42,6 +42,13 @@ final class NativeBridge {
             .receive(on: RunLoop.main)
             .sink { [weak self] _ in self?.pushSettings() }
             .store(in: &cancellables)
+
+        viewModel.$isSyncing
+            .removeDuplicates()
+            .dropFirst()
+            .receive(on: RunLoop.main)
+            .sink { [weak self] _ in self?.pushSettings() }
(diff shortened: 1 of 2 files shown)


mm7894215:
Merged 🎉 Thanks Jian! Clean, well-scoped fix — reusing the existing pushSettings() path with removeDuplicates/dropFirst is exactly right. Test green. Shipping in 0.35.1.

PR #125, opened 2026-06-01
feat(dashboard): replace native selects with a rounded custom dropdown

Hasasasa:
## What

Replace the browser-native `<select>` popups (OS-rendered, unstyleable, out of place against the app) with a shared rounded dropdown built on `@base-ui/react` Select — rounded corners, soft shadow, hover states, selected-item check. Adopted at every former native-`<select>` site: language & currency (settings), project count (DataDetails + ProjectUsagePanel), menu-bar slots (Widgets), leaderboard page size. Shared across mac/web/win, no platform gating.

## Notes after review

- **Async test fix** — `WidgetsPage.test.jsx` queried the menu-slot popup synchronously right after opening it; the base-ui popup mounts asynchronously, so the query raced the mount and failed intermittently. Switched to `await findByRole`.
- **No Windows-portal claim** — An earlier revision portaled the popup into `#root` instead of base-ui's default (`document.body`), justified as a fix for a WebView2 bug where body-portaled overlays "mount but don't paint". I ran a controlled A/B on the current Windows build (CDP-injected `position:fixed` divs — solid and `backdrop-filter` — duplicated into `body` and `#root`, captured at the OS presentation layer, not via a CDP screenshot) and **could not reproduce it** — all variants painted. The premise didn't hold, so the `#root` override is dropped and the popup uses base-ui's default (which already escapes ancestor `overflow` clipping). This PR makes no Windows-specific behavior claim.

## Verification
- typecheck clean, dashboard build OK
- `WidgetsPage` / `LeaderboardPage` / `ProjectUsagePanel` tests pass
- `validate:copy` / `validate:ui-hardcode` / `validate:guardrails` pass
- Pre-existing & unrelated: `SkillsPage.test.jsx` fails on `getByText("Grok")` on this branch's base and on `main` — not touched here.

## Summary by CodeRabbit

* **New Features**
  * Added a shared Select component for dashboard dropdowns.
* **Refactor**
  * Replaced native selects with the new Select across settings, leaderboard, widgets, data details, and project usage.
* **UI Improvements**
  * Improved styling, alignment, truncation, and accessibility for language, currency, page-size, slot picker, and limit controls.
* **Tests**
  * Updated widget/menu-bar tests to match the new dropdown interactions and selection behavior.


mm7894215:
Thanks Phil — the component itself is clean, but I'd like to hold merge on two things: one confirmed blocker and one premise I couldn't verify.

### 🔴 CRITICAL — `WidgetsPage.test.jsx` fails

The PR says "Dashboard tests pass", but on this branch the test fails consistently:

```
src/pages/WidgetsPage.test.jsx:66
const listbox = screen.getByRole("listbox", { name: secondaryLabel });
→ TestingLibraryElementError: Unable to find an accessible element with the role "listbox"
```

Line 66 uses a **synchronous** `getByRole` to assert on a popup that mounts **asynchronously** (portal). The query runs before the listbox paints. Changing it to `await screen.findByRole("listbox", …)` makes it pass — I verified both states locally (fails with `getByRole`, passes with `findByRole`).

This matters beyond the one line: the dashboard vitest suite isn't wired into CI, so a red test here won't be caught by the automated checks and will rot for whoever touches this next. Please fix line 66 before merge.

### 🟠 HIGH — the `#root`-instead-of-`body` portal premise needs evidence

The PR's core justification is: *"overlays portaled to `body` mount in the DOM but never paint on-screen under the Windows transparent composition surface."* I can't reconcile that with the rest of the app — `UsageOverview`, `Toast`, `ConfirmModal`, `CostAnalysisModal`, `TrendMonitorZoomModal`, `Sidebar` etc. all portal to `body` and (as far as we know) render fine on Windows.

Two possibilities:
1. The body-portal-invisible claim is real → then those existing components are all latently broken on Windows and this is a much bigger issue than one Select.
2. The claim doesn't hold → then the `#root` redirect is unnecessary.

Could you attach the Windows evidence (the CDP `elementFromPoint` repro / a screenshot showing a `body`-portaled overlay invisible vs the `#root` one painting)? Right now it's asserted but not demonstrated.

Relatedly, by that same logic the migration is incomplete: `SkillsPage.jsx:250` and `:1480` still use raw `@base-ui` `Select.Portal` (default `body` container), so "every native select / all former sites" doesn't fully hold yet.

### Summary

- Must-fix before merge: line 66 `getByRole` → `await findByRole`.
- Please clarify/justify the `#root` portal premise with Windows evidence (or drop it), and either migrate `SkillsPage` or note why it's excluded.

#124 and #126 are going in separately; this one I'd like to keep open until the above are resolved.

Hasasasa:
Thanks for the careful review — both points were right, and the second one more than I'd claimed. Updates pushed.

**🔴 Test (line 66):** Fixed — `getByRole` → `await findByRole`. The menu-slot popup mounts asynchronously (base-ui portal + floating-ui positioning), so the synchronous query raced the mount. Deterministic now (5/5 locally).

**🟠 `#root` portal premise — this was our misjudgment.** I ran a controlled A/B on the Windows WebView2 build: CDP-injected `position:fixed` divs (one solid background, one with `backdrop-filter`), each duplicated into `document.body` and into `#root`, then captured at the **OS presentation layer** (an actual window screen-grab — not `Page.captureScreenshot`, which composites in-renderer and wouldn't reveal the difference). **All four painted.** The body-portal-invisible behavior did not occur.

To be straight about it: the `#root` override was added pre-emptively, by analogy to the `TrendMonitorZoomModal` inline-rendering note, and the Select case was **never actually verified at the presentation layer** — its only "check" was `elementFromPoint`, which proves the node is in the DOM, not that it paints on screen. I have no evidence the body portal was ever broken for a solid-background popup, and writing it up as a Windows bug fix was wrong.

So I've **dropped the `#root` override and reverted to base-ui's default** (`document.body`), which already escapes the ancestor `overflow` clipping the popup needs. The `lib/portal-root.js` helper is removed, and since the premise is gone the two `SkillsPage` raw `Select.Portal`s stay on the default as well — no migration needed. **This PR is purely the rounded-dropdown styling; it makes no Windows-specific claim.**

Unrelated heads-up while I was in here: `SkillsPage.test.jsx` fails on `getByText("Grok")` on this branch's base and on `main`, independent of these changes — flagging it since you noted the dashboard vitest suite isn't wired into CI.

PR #126, opened 2026-06-01
fix(edge): display name silently reverts after editing (drop nonexistent updated_at column)

Hasasasa:
## Symptom

Editing the **display name** in Settings appears to succeed, but navigating to another menu and back to Settings reverts it to the previous name. Reproduces on **both macOS and Windows** (they share the same cloud edge function), so this is not a native-shell issue.

The GitHub-URL field and the public/anonymous toggles right next to it persist fine — only the display name reverts.

## Root cause (verified against the live backend)

Querying production InsForge directly:

```
tokentracker_user_profiles.display_name  → exists, readable
tokentracker_user_profiles.updated_at    → does NOT exist
  → Postgres 42703 "column tokentracker_user_profiles.updated_at does not exist"
```

`tokentracker-public-visibility`'s POST upserts the display name into `tokentracker_user_profiles` **with an `updated_at` column that table doesn't have**, so the upsert fails with 42703. But the call was `await ...upsert(...)` **without checking the returned `error`** — so the POST still returned `200` with an optimistic result body. The UI showed "saved", nothing was written, and the next `GET` read back the unchanged `display_name` → silent revert on every revisit.

Why only the display name: `github_url` / `enabled` / `anonymous` are written to a *different* table, `tokentracker_user_settings`, which **does** have an `updated_at` column — so those persist.

## Fix

- Drop `updated_at` from the `tokentracker_user_profiles` upsert (the column doesn't exist; nothing reads it either).
- Check the returned `error` and return `500` on failure, so a future schema mismatch surfaces instead of silently faking success.
- The `tokentracker_user_settings` upsert is unchanged (its `updated_at` column exists).

## Deployment note ⚠️

This is an **InsForge edge function** (`dashboard/edge-patches/`). There is no automated deploy path in the repo for these — they're applied manually to InsForge. **Rebuilding the dashboard / desktop apps will NOT make this fix take effect.** A maintainer must redeploy `tokentracker-public-visibility` to InsForge. After deployment, no client/app changes are needed (both platforms call the cloud function).

## Summary by CodeRabbit

* **Bug Fixes**
  * Display name updates are now normalized (trimmed and limited to 50 characters) before saving.
  * The saved/displayed name reflects this normalized value consistently.
  * If saving the display name fails, the system now returns an explicit error instead of failing silently.

### DIFF
--- a/dashboard/edge-patches/tokentracker-public-visibility.ts
+++ b/dashboard/edge-patches/tokentracker-public-visibility.ts
@@ -182,14 +182,16 @@ export default async function (req: Request): Promise<Response> {
     // here so the POST still returned 200 with an optimistic result, the UI
     // showed "saved", but the next GET read back the unchanged name — the
     // display name silently reverted on every revisit. Do not re-add it.
+    // Trim + cap once so the saved value and the returned value can't drift.
+    let normalizedDisplayName: string | null | undefined = undefined;
     if (typeof body.display_name === "string") {
-      const trimmed = body.display_name.trim().slice(0, 50);
+      normalizedDisplayName = body.display_name.trim().slice(0, 50) || null;
       const { error: profileErr } = await client.database
         .from("tokentracker_user_profiles")
         .upsert(
           {
             user_id: userId,
-            display_name: trimmed || null,
+            display_name: normalizedDisplayName,
           },
           { onConflict: "user_id" },
         );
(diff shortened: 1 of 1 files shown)


mm7894215:
Heads up — after merging this I tested the save end-to-end against the live edge function and it **still didn't persist**. This PR's diagnosis was incomplete, and my review was too: I confirmed the `updated_at` column was absent but never ran an actual POST→GET round-trip before merging, so I missed the real bug.

**Root cause is one layer deeper:** `tokentracker_user_profiles` is a **VIEW**, not a table:

```sql
SELECT users.id AS user_id,
       COALESCE(s.display_name, users.profile->>'name', split_part(users.email,'@',1)) AS display_name,
       COALESCE(s.avatar_url, users.profile->>'avatar_url') AS avatar_url
FROM auth.users
LEFT JOIN tokentracker_user_settings s ON s.user_id = users.id;
```

You can't `INSERT`/upsert into it. Dropping `updated_at` only peeled back the first error (Postgres `42703`, column-not-exist) and exposed the real blocker — the live POST returns:

```
500 {"error":"cannot insert into view \"tokentracker_user_profiles\""}
```

So the name was never written. The display_name has to be written to the **base table** `tokentracker_user_settings` (which has `PRIMARY KEY (user_id)`); the view then surfaces it via `COALESCE(s.display_name, …)`, so the GET reads it back.

**Fix shipped** in `a4679ce` (v0.36.1): fold `display_name` into the existing `tokentracker_user_settings` upsert, drop the broken view upsert, and stop swallowing the upsert error. Verified end-to-end (POST 200 → row in `user_settings` → view returns the saved name) and the edge function is redeployed.

No action needed from you — just closing the loop so the history is accurate. Thanks for the initial dig; it got us most of the way there.

PR #128, opened 2026-06-02
feat(windows): add floating desktop pet companion

Hasasasa:
## Summary

Adds a **floating desktop pet** to the Windows tray app (follow-up to the merged #115) — the Windows counterpart of the macOS menu-bar Clawd mascot. Since the Windows taskbar can't host an animated icon, the sprite floats directly on the desktop.

## Demo

https://github.com/user-attachments/assets/238ceb27-e0bd-440b-a74f-e5ac3fa501d1

## What it does

- **Transparent floating window** — a WPF *layered* window hosting a windowless WebView2 composition surface, so only the crab's opaque pixels show (no frame, no box). Always-on-top, no taskbar entry, never steals focus.
- **State machine mirrors macOS** — calm idle rotation (every 12–25s); `sleeping` at zero usage; `disconnected` when the local API is unreachable; a typing pose while syncing. Token *volume* drives only the hover bubble, not the animation.
- **Hover interaction** — sliding the cursor over the pet leans its body and glances its eyes toward the pointer (no click needed).
- **Tap** — cycles the reaction animations, each with a localized quip.
- **Drag** — past a small threshold, hands off to the native window-move loop; position is persisted.
- **Right-click** — context menu: open/close dashboard, size (small/medium/large), close pet.
- **Types along as you type** — global typing activity plays the typing pose; ~30s of non-stop typing triggers an "overheat" gag (error pose for 5s). A normal launch now shows the pet instead of auto-opening the dashboard, and the tray-icon tooltip stays the app name.

## Data consistency & privacy

- The native host pushes currency, locale, today's usage (**the same numbers the tray shows**), syncing and connection state into the page via `window.__ttPet*` globals + `pet:*` events — the pet does **no independent polling** and never drifts from the tray.
- Typing detection is **count-only** (`GetAsyncKeyState`, "was a typing key pressed since last poll"): it detects *that* you're typing, never *which* key — in line with the project's token-counts-only privacy rule. Mouse buttons and lone modifiers are excluded.

## macOS / web impact: none (verified)

The pet ships as an **opt-in** Vite entry (`pet.html`) gated by `TOKENTRACKER_BUILD_PET=1`, enabled only by the Windows build. Adding a rollup entry reshuffles shared chunks, so keeping it off by default keeps the macOS + web builds byte-identical (this project has a history of chunk-split changes breaking the native OAuth callback).

- Default `npm run build` → main bundle hash unchanged (`main-DT0Mq89f.js`); `dist/` contains **no** pet artifacts.
- No macOS/web source references `pet.jsx` / `pet-quips.js` / `pet.html`.
- All other changed files are Windows-only (`TokenTrackerWin/`, `release-windows.yml`, `bundle-node.ps1`).

## Testing

- `dotnet build -c Debug` — 0 warnings, 0 errors; launched and exercised the pet manually (hover lean, eye glance, tap cycle, drag, right-click menu, typing pose, overheat).
- `npm run build` (default) — main bundle byte-identical; `TOKENTRACKER_BUILD_PET=1` build emits `pet.html` + `pet-*.js`.
- `npm run validate:ui-hardcode`, `validate:guardrails`, `validate:copy` — pass.

## Notes

- No version bumps in this PR — release/version lockstep is left to the maintainer's release flow.
- `ui-hardcode-baseline.json` gains a `pet.jsx` entry: 3 inline-style colors (the standalone pet has no Tailwind) and 2 false-positive `rawText` tokens that are actually code-comment fragments.

## Summary by CodeRabbit

* **New Features**
  * Floating desktop “pet” for Windows: animated, transparent, draggable, persists position/size, three sizes, reacts to hover/tap/typing, and shows token usage/cost, sync/connection state, and localized quips.
  * Tray menu adds pet toggle and size controls; right-clicking the pet opens the dashboard.
  * Pet UI served as a new standalone page and included in Windows builds only when enabled at build time.


Hasasasa:
Thanks @coderabbitai — addressed in `1b0aea2`:

1. **`RefreshSummary` skipped pet context before the first poll** ✅ Fixed. Currency + locale (and a zero-usage update) are now pushed even when `_lastStats` is null, so a cold-launched pet no longer sits in default USD/English. I left connection state out of the null branch on purpose — it's already owned by `OnServerStatusChanged`, and pushing `connected=false` there would flash "disconnected" while the server is still starting.

2. **Drag ref not cleared on native-move handoff** ✅ Fixed. The native move loop swallows the matching `mouseup`, so the page's drag flag stayed set and suppressed hover-lean + taps until the next click. Cleared `dragRef.current` at handoff.

3. **Convert `pet-quips.js` → TypeScript** — Skipping. The dashboard component tree is overwhelmingly JSX (103 `.jsx` vs 4 `.tsx`), `lib/` already has 31 `.js` files (e.g. `native-bridge.js`), and the pet entry itself is `pet.jsx`. A lone `.ts` lib file consumed by JSX would be inconsistent churn rather than alignment, so the pet files follow the prevailing dashboard style.

mm7894215:
Thanks for this — the floating pet is genuinely well done, and the "mirror the tray, no independent polling" data flow plus the privacy-conscious count-only typing detection are exactly the right calls. 👍

One thing stands out watching it in action: **the tap quips feel disconnected from the actual numbers.** The root cause is in `pickQuip()` (`dashboard/src/lib/pet-quips.js`):

```js
const candidates = [...(pool[tierFor(tokens)] || []), ...pool.personality];
return candidates[Math.floor(Math.random() * candidates.length)] || "";
```

Two issues:
1. The `personality` pool (5 lines) is **always** mixed in, while each usage tier only has 2 lines — so on any non-empty day there's just a ~2/7 chance of landing on a usage-related line. Roughly 70% of taps say generic things ("Hey there~", "Every token tells a story") that have nothing to do with today's usage.
2. Even the tier lines are purely qualitative ("Busy day!", "Heavy usage today!") — they never include the actual figure. The real number only ever shows in the hover bubble.

For reference, the macOS companion (`TokenTrackerBar/TokenTrackerBar/Views/ClawdCompanionView.swift`, `quipPool`) bakes the real data straight into the quips — today's token count and cost, 7d/30d rolling totals, active-day streak, top-model share, conversation count — and only appends the personality lines at the end. This Windows port simplified that to qualitative tiers + personality, which is where the disconnect comes from.

Would you mind tightening the quip↔data link before we merge? Concretely:
- **Interpolate the real numbers into tap quips** the way macOS does — e.g. "🚀 1.2M tokens today!", "💸 $3.45 on AI so far". `pet.jsx` already has `formatTokens(today.tokens)` and the cost on hand, so the data is right there.
- **Weight the usage-aware lines well above `personality`** (or port a slice of the macOS `quipPool` data lines) so most taps actually reflect the day's activity instead of falling back to generic chatter.

The hover bubble already nails showing the precise number — bringing that same data-awareness to the tap quips would make the pet feel like it's genuinely reacting to *your* usage rather than saying random nice things. Happy to help if useful!

Hasasasa:
Thanks for the thoughtful review @mm7894215 — you're right, the quips were saying nice things instead of *your* numbers. Fixed in `7b8dd44`.

**What changed**

- **Real figures baked into tap quips.** Ported the macOS companion's "today data" quips 1:1 (all five locales) — `tokensToday`, `tokensSpentToday`, `aiInvestedToday`, `billToday`, `aiTabToday` — so a tap now says e.g. `📈 1.2M tokens — $3.45 spent today` / `💳 AI tab today: $3.45`. They interpolate the **same formatted token/cost strings the hover bubble shows**, so the two never disagree. Cost lines are gated on today's cost rounding above zero, matching macOS.
- **Personality down-weighted to a ~25% minority.** On a non-empty day the candidate set is the usage-aware lines (today-data + the qualitative tier line); a generic personality line only surfaces ~25% of the time. Quick sampling on a massive cost-bearing day: **~75% usage-aware, ~54% contain an actual number** (was ~30% usage-aware before).
- `pickQuip` now takes `{ tokens, tokensText, costText, costValue, isSyncing }` instead of positional args.

**One scope note**

macOS stays usage-dominant partly because it also mixes in 7d/30d rolling totals, active-day streak, top-model share and conversation count. The Windows pet currently only receives **today's** tokens + cost from the host (`ApplyUsage(tokens, costUsd)`), so I ported the today-data slice and down-weighted personality directly rather than half-porting stats the pet has no data for. Wiring those richer stats through `UsagePoller` → the host→pet bridge is a clean follow-up if you'd like them — happy to do it in a separate PR, or fold it in here if you'd prefer it before merge.

macOS/web builds remain byte-identical (`main-DT0Mq89f.js`); `validate:ui-hardcode` still passes.

Hasasasa:
We'll bring the richer quips fully in line with the macOS implementation: plumbing 7d/30d rolling totals, active-day streak, top-model share and conversation count through `UsagePoller` → the host→pet bridge so the tap quips draw on the same data set as `ClawdCompanionView`'s `quipPool`, not just today's tokens + cost.

Since the bubble styling also needs some adjustment to carry the longer, data-richer lines well, we'll likely do that fuller pass as a **separate follow-up PR** rather than expanding the scope of this one. This PR already moves the quips from generic to usage-aware (today's real tokens + cost, personality down-weighted); the macOS-parity stats + bubble polish will land in the follow-up.

Let me know if you'd prefer to hold this one for the full treatment instead — otherwise this is good to review as the first step.

PR #134, opened 2026-06-03
feat(windows): align desktop pet quips with macOS + persist currency

Hasasasa:
## Summary

A follow-up to the Windows desktop pet (#128) that brings its speech bubble to full parity with the macOS companion and fixes a currency edge case.

### 1. Full macOS quip-pool parity

The pet's tap bubble previously only knew today's tokens + cost. It now reproduces the macOS companion's complete `quipPool` (`ClawdCompanionView.swift`), 1:1 and in the same order:

- today tokens / cost / usage tier
- **7-day & 30-day rolling totals + per-active-day average**
- **active days this week, perfect-week streak, current streak, all-time active days**
- **top model & runner-up share, model count, multi-tool setup**
- **today's conversation count + "busy talker"**

Taps now rotate through the pool by index (like macOS `quipIndex`) instead of the old random 25% personality weighting — because the data-rich lines outnumber the few personality lines, most taps surface real numbers and personality stays a natural minority. All 13 new string categories are localized across the 5 supported languages (en / zh-CN / zh-TW / ja / ko), ported from `Strings.swift`.

**Data sources (kept efficient):** the 7d/30d/average/active-days/conversation figures ride along in the `usage-summary` response the tray already fetches (it returns a `rolling` block that was previously discarded) — zero extra calls. Only the heatmap (all-time active days) and model breakdown (top models) need their own requests, and those are gated behind pet visibility, so nothing extra is fetched while the pet is hidden.

### 2. Two-line bubble

This is a fixed-size, content-clipping layered window (unlike macOS, where the bubble can overflow the frame), so the longer data-rich quips were truncating with an ellipsis. The bubble now wraps to two lines and the window grows to fit, while the sprite keeps the exact same visual size as before.

### 3. Currency persists across a cold launch

The pet's currency was read only from the dashboard WebView's `localStorage`. On a cold launch with just the pet on screen (no dashboard window yet), it fell back to USD even though the app last ran in, say, CNY. The chosen symbol + rate are now cached natively in `native-settings.json` — exactly how locale and theme are already persisted — so the cold-launched pet matches the app's last-used unit. (Locale/text was already correct via the existing native cache; only currency was missing one.)

## Demo

https://github.com/user-attachments/assets/2183500f-d278-4b40-b1f6-a96f67eb5afe

## Safety / isolation

The pet entry stays gated behind `TOKENTRACKER_BUILD_PET=1`, so the **default macOS/web bundle is byte-identical** (`main-DT0Mq89f.js`, no pet entry). `pet-quips.js` is imported only by `pet.jsx`. The CLI (`src/`) is untouched. C# builds clean; `ui-hardcode` + architecture guardrails pass.

## Notes

- The local server returns `streak_days: 0` (it's server-computed), so the current-streak quip stays dormant against the local backend — same as the macOS pet reads it. The string is ported faithfully and will light up if the backend ever returns a real streak.
- Model breakdown uses a fixed 30-day window (the Windows tray has no period switcher).
- No version bumps — release is the maintainer's call.

## Summary by CodeRabbit

* **New Features**
  * Rich usage statistics: 7-day and 30-day rolling stats, streak tracking, and top models breakdown.
  * Persisted currency preferences; pet quips now contextually generated from live usage data.
  * Enhanced bubble layout to accommodate longer quip text without truncation.

* **Bug Fixes**
  * Eliminated USD default flash on initial launch.

### DIFF
--- a/TokenTrackerWin/Currency.cs
+++ b/TokenTrackerWin/Currency.cs
@@ -1,9 +1,18 @@
+using System.IO;
+using System.Text.Json.Nodes;
+
 namespace TokenTrackerWin;
 
 /// <summary>
 /// Mirror of the dashboard's <c>lib/currency.ts</c> symbol + default-rate table.
 /// The tray reads the user's chosen currency (and live rates) from the WebView's
 /// localStorage; these are the fallbacks when a rate is missing.
+///
+/// The chosen symbol + rate are also cached natively (native-settings.json, alongside
+/// the locale/theme prefs) so the floating pet can show the right currency on a cold
+/// launch — when only the pet is on screen and the dashboard WebView (the live source)
+/// hasn't been created yet, there'd otherwise be nothing to read and the pet would fall
+/// back to USD even though the app last ran in, say, CNY.
 /// </summary>
 internal static class Currency
 {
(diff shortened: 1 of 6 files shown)


mm7894215 (commented):
**Verdict: ✅ Approve (nits only) — merge-ready.**

Reviewed against the project's load-bearing conventions and the isolation claims in the description. All checks pass:

- **Bundle isolation holds.** `pet-quips.js` is imported only by `pet.jsx`, and the `pet.html` entry is gated behind `TOKENTRACKER_BUILD_PET=1` (`dashboard/vite.config.js`). The default macOS/web bundle is unaffected — confirmed.
- **Token semantics are correct.** Rolling/heatmap/model figures use `billable_total_tokens` (`UsagePoller.cs`), matching the macOS `DashboardViewModel`, and never `total_tokens`. The pet does no cost math itself (server pre-computes), so the "never derive cost from `total_tokens`" rule isn't even engaged here.
- **`rolling` block reuse is genuinely free** — it already rides in the `usage-summary` response (`local-api.js`), no extra request. Heatmap + model breakdown are gated behind pet visibility as claimed.
- **`streak_days: 0` dormancy** confirmed (server hardcodes it; quip is gated on `streakDays > 1`).
- **Thread-safety OK** — `IncludeRichStats` is `volatile`, `StatsUpdated` marshals to the UI via `PostToUi`, and the default-struct `TopModels == null` is guarded by `?? Array.Empty<>()` JS-side and `Array.isArray(...)`.

Nits (non-blocking, optional cleanup):
- `PetWindow.cs` pushes `window.__ttPetCostUsd` redundantly with `__ttPetStats.todayCostUsd` (duplicate state; harmless fallback).
- Two empty-list sources for `TopModels` (`NoModels` static vs `?? Array.Empty<>()`) — could standardize.
- `pet-quips.js` `multiToolSetup` uses default `.sort()` vs macOS `.sorted()`; no practical divergence on lowercase-ASCII provider names.

No CRITICAL/HIGH/MEDIUM issues. The macOS `quipPool` port (ordering, thresholds, percent base) matches `ClawdCompanionView.swift` 1:1.

PR #136, opened 2026-06-03
Fix macOS dashboard full-screen Space behavior

fflake33:
## Summary
- Set the macOS dashboard window's activation policy before creating or reusing the window so AppKit treats it as a regular app window when it is shown.
- Give the dashboard `NSWindow` an explicit `.managed` + `.fullScreenPrimary` collection behavior so it can enter its own full-screen Space without joining another app's full-screen Space.
- Add a focused regression test that locks the dashboard Space behavior while preserving the desktop pet's intentional auxiliary full-screen behavior.

## Why
When TokenTracker is opened from the menu bar while another app is already full-screen, the dashboard should behave like a primary work window. It should be able to enter full screen itself, but it should not float into or attach to another application's full-screen Space.

## How
This keeps the existing window lifecycle intact and makes the smallest behavioral change in `DashboardWindowController`: switch the app to `.regular` before the window path branches, and set the dashboard window's `collectionBehavior` to `.managed` and `.fullScreenPrimary`. The desktop pet keeps its separate `.canJoinAllSpaces` / `.fullScreenAuxiliary` behavior.

## Test Plan
- [x] `npm test`

## Release Notes

* **Bug Fixes**
  * Improved macOS window behavior and appearance synchronization with system settings.
  * Enhanced light/dark theme synchronization to keep the app appearance current with system preferences.

* **Chores**
  * Added test coverage for macOS window-space behavior validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mm7894215/TokenTracker/pull/136?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Services/DashboardWindowController.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Services/DashboardWindowController.swift
@@ -5,9 +5,11 @@ import WebKit
 @MainActor
 final class DashboardWindowController: NSObject, NSWindowDelegate, WKNavigationDelegate, WKUIDelegate, WKScriptMessageHandler {
 
-    /// 与 `dashboard` 里 `AppLayout` 主内容区 `lg:pr-3 lg:pb-3`（12pt）一致，用于主卡圆角与窗口圆角同心近似。
+    /// Matches the `AppLayout` main content gutter `lg:pr-3 lg:pb-3` (12pt) in `dashboard`,
+    /// used to approximate concentric rounding between the main card and the window.
     private enum DashboardChromeMetrics {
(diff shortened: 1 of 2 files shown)


mm7894215 (commented):
**Verdict: ✅ Approve**

The two behavioral changes are correct and minimal:
1. Hoisting `NSApp.setActivationPolicy(.regular)` before the window branch ensures consistent state for both create-new and reuse-existing paths.
2. Setting `.collectionBehavior = [.managed, .fullScreenPrimary]` is the textbook fix for preventing the dashboard from attaching to another app's full-screen Space.

One MEDIUM nit (non-blocking): the test at lines 40-44 scans the entire Swift source for forbidden flags, which could false-positive on comments like `// Unlike .fullScreenAuxiliary`. The per-expression check at lines 36-39 is already sufficient; consider dropping lines 40-44 or narrowing to `collectionBehavior` assignment sites only.

No conflict with any other in-flight work. Ship it.

PR #160, opened 2026-06-07
feat(zed): track bring-your-own-provider usage, not just hosted models

Flowm:
## Summary

Surface Zed Agent usage for **all model providers** — not just Zed-hosted (`zed.dev`) — so bring-your-own-provider users (GitHub Copilot, OpenAI/ChatGPT, Anthropic, LM Studio, …) actually see their Zed usage instead of an empty/stale total.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Why

`extractZedTotals` dropped every thread whose `model.provider != "zed.dev"`, so anyone running the Zed Agent with their own provider saw no recent Zed usage. Verified against a real `~/Library/Application Support/Zed/threads/threads.db`:

| Provider | Type | Threads | Tokens |
|---|---|---|---|
| `zed.dev` | hosted | 9 | 656K |
| `copilot_chat` | BYO | 554 | 477M |
| `openai-subscribed` | BYO | 32 | 11M |
| `lmstudio` | BYO | 1 | 0 |

The reader surfaced **656K of ~489M tokens — 99.87% dropped**. BYO threads already carry a populated `request_token_usage` map (the same shape the hosted path sums), so the data was *present but filtered out* — not a parsing/format problem.

## What changed

- **`src/lib/rollout.js`** — remove the `provider == "zed.dev"` filter. Count all providers; skip only providers already reported by a dedicated parser via a new `ZED_DOUBLE_COUNTED_PROVIDERS` set (empty today — Zed's native providers don't overlap with any CLI parser). Each extracted thread is tagged with `provider` + `hosted`.
- **`src/lib/pricing/matcher.js`** — `normalizeZedModel()` maps Zed's inconsistent model strings (a mix of ids like `gpt-5.5`/`claude-opus-4.8` and display names like `Claude Sonnet 4`/`GPT-5 (Preview)`/`Gemini 3 Pro (Preview)`) to the pricing engine's keys. Runs **only at lookup time**, gated on `source == "zed"` (mirrors the existing `normalizeAntigravityModel`), so the raw Zed model name is preserved for display.
- **`src/commands/status.js`** — replaces "hosted models only" with "all providers" plus a thread count from the cursor, distinguishing "N threads counted" from "no usage counted yet".
- Stale comments updated in `sync.js`, the `rollout.js` section header, and the README provider table.

## Reviewer notes

- **`source` stays `"zed"`** for hosted and BYO alike (no queue-entry schema change, no dashboard change); model name differentiates them and pricing resolves per model.
- **First sync after upgrade backfills historical BYO threads** into their real `updated_at` buckets (the cumulative-delta cursor starts empty). This is an intentional, large retroactive increase in reported Zed usage; pre-existing hosted-era rows are unchanged.
- **ACP agents launched inside Zed** (Claude Code / Codex / Gemini) are **not** stored in `threads.db` (verified) — they're untouched here and remain tracked by their own CLI parsers, so there's no double-count via the Zed reader.
- **Privacy:** only `source`/`model`/token-counts/`timestamp` are read; the query selects no `folder_paths`, thread `summary`, or message content.

## Tests run

- End-to-end against a real DB: 489.7M tokens surfaced (was 656K), correct per-model cost ($1,314 total, zero tokens-but-$0 rows), hosted-era rows unchanged.
- Unit tests updated/added: `test/zed-parser.test.js` (all providers counted; real request-id-keyed map fixture), `test/pricing.test.js` (Zed normalization + source-gating), `test/zed-goose-reset-cursor.test.js` (cursor-watermark repointed at a genuinely-skipped row). Fixtures are synthetic in-test SQLite DBs (no real user data).
- `npm test` green for the changed scope; `validate:copy`, `validate:ui-hardcode`, `validate:guardrails`, and `architecture-guardrails.test.js` all pass.

## Checklist

- [x] `npm test` passes
- [x] New user-facing strings go through `dashboard/src/content/copy.csv` (no hardcoded UI text) — N/A: only a CLI status string changed (hardcoded English like the other status lines); no dashboard strings touched

> Note: no version bumps included — leaving the release/version policy to the maintainer.

## Summary by CodeRabbit

* **New Features**
  * Extended Zed Agent support to track token usage across all model providers, not limited to hosted models.
  * Added thread count reporting in status output for Zed Agent activity.

* **Documentation**
  * Updated documentation to reflect expanded Zed Agent coverage across all providers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Upgrade with `brew upgrade --cask mm7894215/tokentracker/tokentracker`. The tap 
 | **pi** (`@mariozechner/pi-coding-agent`) | ✅ Auto | Passive reader (`~/.pi/agent/sessions/**/*.jsonl`) |
 | **Craft Agents** | ✅ Auto | Passive session reader (`~/.craft-agent` + workspace session logs) |
 | **Roo Code** (VS Code extension) | ✅ Auto | Passive `ui_messages.json` reader (`rooveterinaryinc.roo-cline`) |
-| **Zed Agent** | ✅ Auto | Passive SQLite reader (`threads.db`, hosted `zed.dev` models only) |
+| **Zed Agent** | ✅ Auto | Passive SQLite reader (`threads.db`, all providers — hosted `zed.dev` + bring-your-own) |
 | **Goose** (Block) | ✅ Auto | Passive SQLite reader (`sessions.db`, cumulative deltas) |
 
 > **Do I need to install any plugin or hook manually?** No. `tokentracker` (or `tokentracker init`) handles everything on first run:
--- a/src/commands/status.js
+++ b/src/commands/status.js
@@ -230,9 +230,16 @@ async function cmdStatus(argv = []) {
   const roocodeTaskFiles = resolveRoocodeTaskFiles(process.env);
   const roocodeInstalled = roocodeTaskFiles.length > 0;
 
-  // Zed Agent — passive read of hosted threads.db (skips external ACP agents).
+  // Zed Agent — passive read of threads.db across all model providers
+  // (hosted "zed.dev" and bring-your-own alike). threadTotals tracks one entry
+  // per thread we've surfaced usage for, so its size distinguishes "DB present
+  // with usage" from "DB present but nothing counted yet" (fresh/undecodable).
   const zedDbPath = resolveZedDbPath(process.env);
   const zedInstalled = fssync.existsSync(zedDbPath);
+  const zedThreadsCounted =
+    cursors?.zed?.threadTotals && typeof cursors.zed.threadTotals === "object"
+      ? Object.keys(cursors.zed.threadTotals).length
+      : 0;
 
   // Goose (Block) — passive cumulative-delta read of sessions.db.
   const gooseDbPath = resolveGooseDbPath(process.env);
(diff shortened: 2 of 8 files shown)


Flowm:
> `5232-5266`: _⚠️ Potential issue_ | _🟠 Major_ | _🏗️ Heavy lift_
> 
> **Counting all Zed providers makes the 10k thread cap unsafe.**
> 
> This scope expansion feeds many more rows into `cursors.zed.threadTotals`, but `parseZedIncremental()` still truncates that map to the top 10k threads by lifetime volume. Once a low-volume BYO thread is evicted, its next update comes back with no baseline and the parser re-emits the thread’s full cumulative total, permanently overcounting persisted usage.

The mechanism is real but pre-existing - the 10k cap and zero-baseline re-emit are unchanged; counting BYO only changes which threads fill the map. It needs >10k Zed threads for one user, and eviction keeps the top 10k by volume, so only the smallest threads can be evicted — the overcount is tiny when it fires at all.

Excluding BYO would revert the PR. The proper fix is to the cap/eviction logic (don't zero prev for evicted threads), which applies to hosted threads too - might be better as a separate issue than scope creep here.

mm7894215:
Merged and shipping in **v0.44.0** — thank you @Flowm! 🎉

This is a genuinely high-quality first contribution: the root-cause analysis against a real `threads.db` (656K of ~489M tokens surfaced, 99.87% dropped), the lookup-time-only `normalizeZedModel` gated on `source == "zed"`, the `ZED_DOUBLE_COUNTED_PROVIDERS` extension point, and the privacy-preserving query were all spot on. The cumulative-delta cursor tests were a pleasure to review.

On the 10k `threadTotals` cap edge case you flagged with CodeRabbit — agreed it's pre-existing and out of scope here. We'll track the eviction-baseline fix (don't zero `prev` for evicted threads, which affects hosted threads too) as a separate issue.

npm publish is underway; the macOS DMG + Windows build follow via the release workflow. Thanks again for making Zed BYO usage actually visible! 🙏

---

PR #163, opened 2026-06-08
fix(limits): make Claude usage limits resilient to rate limits

Flowm:
## Problem

The Claude card in the usage-limits panel was frequently stuck on `Error: Claude API rate limited (429) — wait ~1 minute and refresh.`

Root cause (verified against the live endpoint): `GET https://api.anthropic.com/api/oauth/usage` is shared by the CLI, the menu bar app's embedded server, **and Claude Code itself** — all using the same OAuth token. When the limit trips it returns `429` with a long `retry-after` (observed **~22 minutes**, not "~1 minute"). Every dashboard navigation forced a fresh upstream call, and every restart immediately re-hit the endpoint — which *renews* the penalty instead of letting it drain. There was no fallback, so a single failure blanked the bars.

## Changes

Three independent commits:

1. **`serve cached Claude usage when the API rate-limits`** — persist the last successful read to a disk cache (separate file from Antigravity's, whose writer rewrites the whole file). On a failed fetch, serve the cached read (marked `stale`, with already-expired windows dropped) so the bars stay visible; the error only surfaces when there's no usable fallback.

2. **`read server cache on mount instead of forcing a refresh`** — the Limits page no longer forces a live upstream call on every navigation; the mount fetch reads the server's cache (in-memory + disk-backed). Only the manual refresh path still forces upstream.

3. **`back off and respect retry-after when Claude usage rate-limits`** — on a 429, persist a cooldown (real `retry-after`, capped at 60m) to a shared file under the tracker dir and **skip the upstream call until it expires**, so every surface stops hammering and the limit can drain. Serve cached data while cooling down, else an accurate `retry in ~Nm` message. Clear the cooldown on the next success.

## Verification

- `node --test test/usage-limits.test.js` — 46 pass (new tests: stale fallback served, error-when-no-cache, cooldown recorded + skips endpoint, cooldown cleared on success, expired windows dropped).
- Dashboard vitest (hook + LimitsPage) — 8 pass; `npm run dashboard:build` ✓; `npm run validate:guardrails` ✓.
- Confirmed live: 429 with `retry-after` → accurate `retry in ~17m` + cooldown written; a second call did **not** touch the endpoint; on recovery the success cache is written and the cooldown auto-cleared.

## Notes

- `src/` + `dashboard/` scope → release-eligible (npm + DMG + Windows). The shared cooldown file under `~/.tokentracker/tracker/` lets the CLI and the app's embedded server coordinate once both run this code.
- Expired-token (`401`) handling is unchanged by design — that requires the user to re-auth via Claude Code; auto-refresh was deliberately avoided to not risk desyncing Claude Code's own credentials.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced rate limit handling with persistent caching and cooldown tracking.
  * Reduces unnecessary upstream API calls during navigation.
  * Serves previously cached data during rate-limited periods instead of failing.

* **Tests**
  * Added comprehensive test coverage for rate limit scenarios and cache fallback behavior.

### DIFF
--- a/dashboard/src/hooks/use-usage-limits.test.tsx
+++ b/dashboard/src/hooks/use-usage-limits.test.tsx
@@ -59,7 +59,8 @@ describe("useUsageLimits", () => {
 
     await waitFor(() => expect(result.current.data).toEqual(freshLimits));
 
-    expect(getUsageLimits).toHaveBeenCalledWith({ refresh: true });
+    // Mount fetch reads the server cache (no forced upstream refresh).
+    expect(getUsageLimits).toHaveBeenCalledWith();
     expect(publishUsageLimitsPreloadState).toHaveBeenCalledWith(freshLimits, {
       source: "page-load",
     });
@@ -75,7 +76,7 @@ describe("useUsageLimits", () => {
 
     await waitFor(() => expect(result.current.isLoading).toBe(false));
 
-    expect(getUsageLimits).toHaveBeenCalledWith({ refresh: true });
+    expect(getUsageLimits).toHaveBeenCalledWith();
     expect(result.current.data).toEqual(freshLimits);
     expect(result.current.error).toBeNull();
   });
--- a/dashboard/src/hooks/use-usage-limits.ts
+++ b/dashboard/src/hooks/use-usage-limits.ts
@@ -63,7 +63,10 @@ export function useUsageLimits(options?: UseUsageLimitsOptions) {
     let cancelled = false;
     (async () => {
       try {
-        const res = await getUsageLimits(initialRefresh ? { refresh: true } : {});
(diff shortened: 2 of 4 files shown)


mm7894215:
Merged and shipping in **v0.44.1** — thanks again @Flowm! 🎉

Really clean fix: cache the last good read + serve it stale, stop forcing upstream on every mount, and persist a `retry-after`-aware cooldown so the CLI, the menu-bar app's embedded server, and Claude Code itself all stop hammering the shared OAuth usage endpoint and let the limit drain. The 46 tests covering stale fallback / cooldown record+skip / clear-on-success made this easy to verify. npm publish is underway; the macOS DMG + Windows build follow in the release.

---

PR #168, opened 2026-06-09
Sync macOS limit display preferences

### LINKED ISSUE #162: [feature]: mac下菜单栏限额显示是否应跟随「设置 → 限额显示」或者做一个开关 ？
### What problem does this solve?

目前「设置 → 限额显示」和「小组件 → 菜单栏」是两套配置：前者控制限额页显示哪些 provider，后者单独决定菜单栏显示什么。关掉某个 provider 后，菜单栏仍可能显示对应限额，容易让用户困惑。

### Proposed solution

在菜单栏配置处增加一个开关，例如「跟随限额显示设置」：

• 开启时：菜单栏可显示的限额项与设置中心保持一致（已关闭的 provider 不出现在菜单栏）。
• 关闭时：菜单栏可单独配置，天然比设置中心显示更少内容。

### Alternatives considered

_No response_

### Category

New AI tool integration

### Anything else?

_No response_

fflake33:
## Summary

Synchronize Usage Limits display preferences between the Dashboard settings and the macOS menu bar so disabling or reordering a provider in one place is reflected in the other.

Closes #162.

## Scope

- [ ] CLI (`src/`)
- [x] Docs / CI / config

## What changed

- Replaced display-mode-only native bridge sync with a full `limitsPreferences` snapshot containing `displayMode`, `providerOrder`, `providerVisibility`, and `updatedAt`.
- Made the macOS `LimitsSettingsStore` normalize and persist the same snapshot shape that the Dashboard stores in `localStorage`.
- Mirrored user-authored menu bar changes back to the embedded Dashboard without echoing bridge-applied snapshots back again.
- Kept backward compatibility for old `limitsDisplayMode` bridge payloads during migration.
- Added focused Dashboard and macOS unit tests, plus CI steps for the new limits preference coverage.
- **(maintainer follow-up, 6f66751)** Hidden providers are now also excluded from the menu bar metrics: `availableItemIDs` filters them (even when currently selected), the status-bar renderer filters at draw time, and applied bridge snapshots fall through to `pushSettings` so the stored selection self-heals. `MenuBarDisplayMetric`/`MenuBarDisplayPreferences` moved to `Models/MenuBarDisplayPreferences.swift` for unit-test coverage, plus a Swift↔JS provider-id parity guardrail test. This completes the literal #162 repro path (hide a provider in Settings → its metric leaves the menu bar).

## Logic

The Dashboard settings page remains the visible source of truth for Usage Limits preferences. The macOS app mirrors that state in `UserDefaults` through `NativeBridge`, so the menu bar no longer has a separate provider visibility/order state.

Each user-authored change gets a monotonic millisecond `updatedAt`. When Dashboard and native snapshots race, the newer timestamp wins. If neither side has a timestamp during first migration, the Dashboard snapshot wins the tie so the behavior is deterministic and matches the settings UI users already see.

## Tradeoffs

This implements automatic convergence instead of adding a separate "follow Usage Limits settings" toggle. That keeps the preference model smaller: there is one normalized snapshot rather than two independent configurations plus a third switch describing how they relate. The cost is less per-surface customization, but it directly addresses the confusing case from #162 where a provider hidden in Settings still appeared in the menu bar.

## Checklist

- [x] `npm test` passes
- [x] New user-facing strings go through `dashboard/src/content/copy.csv` (no hardcoded UI text; this PR adds none)

## Tests

- `npm test`
- `npm --prefix dashboard test -- src/hooks/use-limits-display-prefs.test.js`
- `xcodegen generate`
- `ruby scripts/patch-pbxproj-icon.rb`
- `xcodebuild test -scheme TokenTrackerBarTests -destination 'platform=macOS'`
- `npm run validate:copy` (passes with existing unused-copy warnings)
- `node --test test/architecture-guardrails.test.js`
- `npm run dashboard:build`
- `git diff --check`

<details>

### Risk layer triggers

### Rules / invariants

- Usage Limits provider IDs are normalized against the existing known-provider list.
- Native bridge-applied snapshots do not publish back to the Dashboard, preventing echo loops.
- A timestamped preference snapshot wins over an untimestamped legacy payload.

### Boundary matrix

- Dashboard localStorage change -> writes a full bridge snapshot.
- macOS menu bar change -> persists to UserDefaults and pushes a full Dashboard settings payload.
- Older native payload with only `limitsDisplayMode` -> accepted only before timestamped snapshot migration.

### Public exposure checklist

Not applicable.

</details>

## Summary by CodeRabbit

* **New Features**
  * Unified, persistent “limits” preferences snapshot syncing between Dashboard and macOS app with better conflict resolution and stable timestamps; menu-bar metric selection and visibility honor hidden providers.

* **Tests**
  * Extensive new and expanded unit/integration tests covering persistence, cross-tab sync, native-bridge mirroring, parity checks, and updatedAt edge cases.

* **Chores**
  * CI updated to run macOS unit tests.


mm7894215 (changes requested):
First off: this is an exemplary contribution — the echo-loop protection is genuinely sound (I traced concurrent-edit, same-millisecond, stale-snapshot and repeated-push scenarios; everything converges in 1–2 round trips), the timestamp monotonicity guards and clock-rollback handling are careful, the dashboard hook's public interface is unchanged so all consumers keep working, and the test coverage is above this repo's existing bar. The drag-reorder fix in `reorderedProviderOrder` (correct `move(fromOffsets:toOffset:)` semantics) is also a real bug fix worth calling out.

That said, two issues need to be resolved before merge:

### 1. Scope mismatch — this doesn't close #162 as written

#162 is about the **menu bar itself** (Widgets → menu bar metric selection): `MenuBarDisplayPreferences.availableItemIDs` in `StatusBarController.swift` decides which limit metrics can appear in the menu bar, and it does not read `providerVisibility` — and this PR doesn't touch it. So after merging, the literal repro from the issue still exists: hide a provider in Settings → Limits Display, and the menu bar can still show e.g. "Claude 5h 42%".

What this PR *does* unify — the popover Usage Limits list + display mode — is a real third inconsistency and worth fixing on its own. Two options:

- **(a)** Also filter the menu bar metric candidates by the synced `providerVisibility` (then `Closes #162` is accurate), or
- **(b)** Keep the PR scope as-is and drop the `Closes #162` link, retitling to make clear it syncs the popover list.

(a) would be ideal since the same snapshot you built is exactly the state the menu bar needs.

### 2. First-migration tie-break can wipe a user's popover customization

In `shouldApplyBridgeSnapshot`, `case (nil, nil): return true`. Consider a user who customized provider order/visibility **only in the popover** and never touched the Dashboard settings page: their localStorage has no limits keys, so on first launch after upgrade the Dashboard pushes a *factory-default* snapshot with no `updatedAt`, native has no `updatedAt` either → the default snapshot wins and the user's customization is silently and irreversibly overwritten. "Dashboard wins ties" is the right rule when the Dashboard actually has user-authored state, but not when its snapshot is just defaults.

Suggested fix: when none of the limits localStorage keys exist, treat the Dashboard side as "no opinion" — don't push (or mark the pushed snapshot so native keeps its state and pushes back as authoritative). Please add a test for exactly this scenario (the current migration tests all assume the Dashboard side has stored content).

### Non-blocking suggestions (fine as follow-ups)

- Swift `allProviders` ↔ JS `LIMIT_PROVIDER_IDS` is now a cross-language hard sync point. A guardrail test comparing the two lists would prevent the drift class we've been bitten by before (cf. the provider-fingerprint sync in #158's history).
- The Swift half of the echo protection (`NativeBridge.handle`'s applied→return vs rejected→push-back fork) has no test coverage — worth a case or two.
- Consider a `paths` filter (`TokenTrackerBar/**` + the hook files) on the new `macos-tests` CI job so it doesn't occupy a macOS runner on every unrelated PR; and the dashboard test step currently runs only `use-limits-display-prefs.test.js` — running the full dashboard vitest suite would be more useful if we're adding the step anyway.

Happy to re-review quickly once 1 and 2 are addressed.

mm7894215:
Re-reviewed after c5d20cd — **issue 2 (first-migration tie-break) is properly fixed**. I verified the new `hasLocalLimitsPreferenceKeys()` guard closes the only push path (the hook never pushes unsolicited on mount, it only calls `requestNativeSettings()`), ran the 27 hook tests locally, and the added regression test covers exactly the scenario from the review. Nice work.

For **issue 1 (scope / #162)** I went ahead and implemented option (a) directly on this branch (6f66751) so the PR closes #162 as written:

- `MenuBarDisplayPreferences.availableItemIDs` gains a `hiddenProviders` filter. Hidden providers are excluded **even when currently selected** — hiding is user-authored intent, unlike a transient provider outage where `keepingSelected` rightly keeps the slot.
- `buildMenuBarDisplayValues()` filters at render time too. This matters because a dashboard-applied snapshot re-renders via `.nativeSettingsChanged` before the stored selection is re-normalized; without it the literal #162 repro survived on the dashboard→menu-bar path.
- The `limitsPreferences` bridge case now falls through to `pushSettings()` on applied snapshots instead of returning early — the push self-heals the stored menu-bar selection and refreshes the dropdown candidates. The echo is convergent with your hook: the dashboard receives a snapshot identical to its local state and does not write back (covered by your "does not write back when native limitsPreferences already matches Dashboard" test).
- `MenuBarDisplayMetric`/`MenuBarDisplayPreferences` moved out of `StatusBarController.swift` into `Models/MenuBarDisplayPreferences.swift` so the unit-test target can compile them; added tests for the hidden filter and a Swift↔JS provider-id parity guardrail (`test/limits-providers-parity.test.js`) per the earlier non-blocking note.

Full CLI suite, dashboard hook tests, and macOS unit tests pass locally. Will merge once CI is green.

mm7894215 (approved):
Both blocking items resolved (issue 2 by c5d20cd, issue 1 by 6f66751). CI green across all six checks. Approving.

PR #169, opened 2026-06-09
Skip forked Codex replay usage

lhandal:
## Summary
- skip historical replay token_count rows in forked Codex rollouts
- keep skipped replay rows advancing the cumulative cursor baseline so later live deltas stay correct
- add a regression test for the replay inflation case

## Tests
- node --test test/rollout-parser.test.js
- npm test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rollout parsing to detect per-file rollout dates and fork markers so events from forked replays occurring before a rollout are skipped, ensuring live token usage is tracked accurately.

* **Tests**
  * Added tests for forked rollout scenarios to confirm historical replay tokens are excluded and cumulative totals are maintained.

### DIFF
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -824,6 +824,7 @@ async function parseRolloutFile({
       totals = totalUsage;
     }
 
+    // date matching is conservative; same-day fork replays are still counted.
     if (isForkedReplayToken({ isForkedRollout, rolloutDate, currentDate })) continue;
 
     const bucketStart = toUtcHalfHourStart(tokenTimestamp);
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -746,6 +746,9 @@ async function parseRolloutFile({
   let model = typeof lastModel === "string" ? lastModel : null;
   let totals = lastTotal && typeof lastTotal === "object" ? lastTotal : null;
   let currentCwd = null;
+  let currentDate = null;
+  let isForkedRollout = false;
+  const rolloutDate = rolloutDateFromPath(filePath);
   let currentProjectRef = projectRef || null;
   let currentProjectKey = projectKey || null;
   let eventsAggregated = 0;
@@ -756,7 +759,10 @@ async function parseRolloutFile({
     const maybeTurnContext =
       !maybeTokenCount &&
       (line.includes('"turn_context"') || line.includes('"session_meta"')) &&
-      (line.includes('"model"') || line.includes('"cwd"'));
+      (line.includes('"model"') ||
+        line.includes('"cwd"') ||
+        line.includes('"current_date"') ||
+        line.includes('"forked_from_id"'));
     if (!maybeTokenCount && !maybeTurnContext) continue;
 
     let obj;
@@ -771,6 +777,12 @@ async function parseRolloutFile({
       obj?.payload &&
       typeof obj.payload === "object"
     ) {
+      if (obj.type === "session_meta" && typeof obj.payload.forked_from_id === "string") {
+        isForkedRollout = obj.payload.forked_from_id.trim().length > 0;
+      }
+      if (obj.type === "turn_context" && typeof obj.payload.current_date === "string") {
+        currentDate = normalizeIsoDate(obj.payload.current_date);
+      }
       if (typeof obj.payload.model === "string") {
         model = obj.payload.model;
       }
@@ -811,6 +823,8 @@ async function parseRolloutFile({
     if (totalUsage && typeof totalUsage === "object") {
       totals = totalUsage;
     }
+
+    if (isForkedReplayToken({ isForkedRollout, rolloutDate, currentDate })) continue;
 
     const bucketStart = toUtcHalfHourStart(tokenTimestamp);
     if (!bucketStart) continue;
@@ -1770,6 +1784,21 @@ function toUtcHalfHourStart(ts) {
     ),
   );
   return bucketStart.toISOString();
+}
+
+function rolloutDateFromPath(filePath) {
+  const match = path.basename(String(filePath || "")).match(/^rollout-(\d{4}-\d{2}-\d{2})T/);
+  return match ? match[1] : null;
+}
+
+function normalizeIsoDate(value) {
+  const raw = typeof value === "string" ? value.trim() : "";
+  const match = raw.match(/^(\d{4}-\d{2}-\d{2})/);
+  return match ? match[1] : null;
+}
+
+function isForkedReplayToken({ isForkedRollout, rolloutDate, currentDate }) {
+  return Boolean(isForkedRollout && rolloutDate && currentDate && currentDate < rolloutDate);
 }
 
 function normalizeNonNegativeNumber(value) {
--- a/test/rollout-parser.test.js
+++ b/test/rollout-parser.test.js
@@ -616,6 +616,79 @@ test("parseRolloutIncremental splits usage into half-hour buckets", async () => 
   }
 });
 
+test("parseRolloutIncremental skips historical replay tokens in forked Codex rollouts", async () => {
+  const tmp = await fs.mkdtemp(path.join(os.tmpdir(), "tokentracker-rollout-"));
+  try {
+    const rolloutPath = path.join(
+      tmp,
+      "2026",
+      "06",
+      "09",
+      "rollout-2026-06-09T20-46-23-fork.jsonl",
+    );
+    await fs.mkdir(path.dirname(rolloutPath), { recursive: true });
+    const queuePath = path.join(tmp, "queue.jsonl");
+    const cursors = { version: 1, files: {}, updatedAt: null };
+
+    const replayUsage = {
+      input_tokens: 100,
+      cached_input_tokens: 0,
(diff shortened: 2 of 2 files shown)


mm7894215:
Thanks for the fix and the regression test — this addresses a real inflation source, and the implementation is solid.

I verified the premise independently before reviewing the code:

- Confirmed against upstream `codex-rs` source: `EventMsg::TokenCount` is in the persisted rollout item list (`rollout/src/policy.rs`), forking copies the full history into the new rollout file, and lines are stamped with the **write-time** timestamp — which explains why you key off `turn_context.current_date` instead of the line timestamp. Good call.
- Checked real rollouts on my machine: `forked_from_id` and `current_date` both exist in the wild, and `current_date` is the **local** calendar date — same convention as the filename date, so there's no timezone false-positive where a live early-morning turn in a forked session could be misclassified as replay and dropped. That was my main worry and it's cleared.
- The skip is placed after `totals = totalUsage`, so the cumulative baseline still advances across skipped replay rows and subsequent live deltas stay correct — the test asserts `lastTotal`, 👍
- All failure modes are fail-safe in the right direction (unparseable filename, missing `current_date`, non-forked file → counted, never silently dropped).

CI hadn't run on this PR (first-time contributor approval), so I ran it locally: `test/rollout-parser.test.js` 149/149 pass, full suite 830/832 (the 2 failures are local-environment issues unrelated to this change).

Two asks before merge:

1. **Document the same-day fork gap in the PR description / a code comment.** When a session is forked on the same local day it was created, the replayed rows have `current_date == rolloutDate` and are still double-counted. That's an inherent limit of the date heuristic and the conservative direction is correct (never drop live usage), but it should be stated explicitly so future "still over-counting after fork" reports aren't mistaken for a regression here.
2. **Could you share the Codex CLI version (and ideally a redacted sample of a forked rollout's first few lines) where you observed the replayed `token_count` rows?** My local rollouts are from slightly older Codex builds that don't replay TokenCount events yet, so I confirmed the behavior from upstream source rather than end-to-end. Pinning the version/shape in the PR would make the fixture authoritative.

With those two addressed this is good to go.

lhandal:
Addressed both asks.

- Added a code comment in `src/lib/rollout.js` noting the conservative same-day fork gap: same-day fork replays are still counted so live usage is not accidentally dropped.
- Pushed that in `e1bd378`.
- Local Codex version: `codex-cli 0.137.0`; the observed forked rollout metadata reports `cli_version: 0.137.0-alpha.4`.

The motivating rollout file was named like `rollout-2026-06-09T22-46-23-<id>.jsonl`. It had 3,787 `token_count` rows, with 3,194 replay rows while `turn_context.current_date` was still `2026-05-08`.

Redacted excerpt:

```jsonl
{"line":1,"type":"session_meta","payload":{"id":"<redacted>","forked_from_id":"<redacted>","timestamp":"2026-06-09T20:46:23.397Z","cli_version":"0.137.0-alpha.4","originator":"Codex Desktop"}}
{"line":2,"type":"session_meta","payload":{"id":"<redacted>","timestamp":"2026-05-08T20:51:52.262Z","cli_version":"0.129.0-alpha.15","originator":"Codex Desktop"}}
{"line":6,"type":"turn_context","payload":{"cwd":"<redacted>","model":"gpt-5.5","effort":"high","current_date":"2026-05-08"}}
{"line":16,"type":"event_msg","payload":{"type":"token_count","info":{"total_token_usage":{"input_tokens":24309,"cached_input_tokens":3456,"output_tokens":493,"reasoning_output_tokens":249,"total_tokens":24802},"last_token_usage":{"input_tokens":24309,"cached_input_tokens":3456,"output_tokens":493,"reasoning_output_tokens":249,"total_tokens":24802},"model_context_window":258400}}}
```

Focused test still passes: `node --test test/rollout-parser.test.js` -> 149/149.

mm7894215 (approved):
Both asks addressed: the same-day fork gap is documented at the skip site (e1bd378), and the codex-cli 0.137.0 / 0.137.0-alpha.4 sample pins the replayed-TokenCount behavior authoritatively. Re-ran test/rollout-parser.test.js on the updated head: 149/149. Merging.

YUDONGLING:
Thanks @mm7894215 and @lhandal for maintaining TokenTracker.

I checked several forked rollouts and found that files have `forked_from_id` first replay the parent session's history in a short time period after the child session is created. Since `current_date` cannot dela with same-day forks, we may use the UUIDv7 timestamp from the child session ID as the replay boundary.

Then, for Codex rollouts with `forked_from_id`, skip `token_count` events at or before `sessionStart + several seconds` ( this should be okey as a new turn is not very likely to finish and have a new `token_count` event within this window.).

The cumulative `lastTotal` should still be updated before skipping these events, so subsequent live deltas remain correct. If the session ID is not UUIDv7 or a timestamp cannot be parsed, the filter should fail open and count everything normally.

```
const CODEX_FORK_REPLAY_WINDOW_MS = 2500;

function isForkedReplayToken({
  isForkedRollout,
  sessionId,
  tokenTimestamp,
  source,
}) {
  if (!isForkedRollout || source !== DEFAULT_SOURCE) return false;

  const sessionStartMs = uuidV7TimestampMs(sessionId);
  const eventMs = Date.parse(tokenTimestamp);

  if (!Number.isFinite(sessionStartMs) || !Number.isFinite(eventMs)) {
    return false;
  }

  return eventMs <= sessionStartMs + CODEX_FORK_REPLAY_WINDOW_MS;
}
```

& the filter is applied after updating the cumulative baseline:

```
const delta = pickDelta({ totalUsage, lastTotal: totals, lastDelta });
totals = totalUsage;

if (isForkedReplayToken({
  isForkedRollout,
  sessionId,
  tokenTimestamp: timestamp,
  source,
})) {
  continue;
}
```

mm7894215:
Thanks for digging into this. The UUIDv7 direction is reasonable, and the same-day fork gap is real. However, I do not think a fixed 2.5-second window is safe as written.

The earlier redacted Codex 0.137 sample in this thread has the child `session_meta` written at `20:46:23.397Z` and the first confirmed replayed `token_count` at `20:46:26.530Z` — 3.133 seconds later. Because the child UUID is created no later than that metadata write, the proposed 2.5-second cutoff would miss that replay.

Even if this is added only as a supplement to the existing date guard, two boundaries still need validation:

- a large or slow replay can continue beyond the cutoff and still be counted;
- a genuine fast first turn can emit `token_count` inside the cutoff and be dropped.

Would you be willing to open a follow-up PR with redacted same-day parent/child rollout samples and regression tests for both cases? The invalid/non-v7 fail-open behavior and cumulative-baseline advancement should remain covered as well.

I would prefer a data-derived replay boundary over a fixed time window. If a time window is unavoidable, the PR should include measurements across Codex versions and history sizes that justify the chosen margin.

YUDONGLING:
> Thanks for digging into this. The UUIDv7 direction is reasonable, and the same-day fork gap is real. However, I do not think a fixed 2.5-second window is safe as written.
> 
> The earlier redacted Codex 0.137 sample in this thread has the child `session_meta` written at `20:46:23.397Z` and the first confirmed replayed `token_count` at `20:46:26.530Z` — 3.133 seconds later. Because the child UUID is created no later than that metadata write, the proposed 2.5-second cutoff would miss that replay.
> 
> Even if this is added only as a supplement to the existing date guard, two boundaries still need validation:
> 
> * a large or slow replay can continue beyond the cutoff and still be counted;
> * a genuine fast first turn can emit `token_count` inside the cutoff and be dropped.
> 
> Would you be willing to open a follow-up PR with redacted same-day parent/child rollout samples and regression tests for both cases? The invalid/non-v7 fail-open behavior and cumulative-baseline advancement should remain covered as well.
> 
> I would prefer a data-derived replay boundary over a fixed time window. If a time window is unavoidable, the PR should include measurements across Codex versions and history sizes that justify the chosen margin.

Sorry for the delayed response, and thank you for pointing out these boundary cases.

I agree that a fixed 2.5-second cutoff is not safe enough. The 2.5-second value in my example was only intended to illustrate the general idea, not as a final proposed threshold. I’ll continue investigating and collect more rollout samples before proposing a concrete implementation.

Since every rollout record includes a precise millisecond timestamp, another possibility is to examine the time gaps between consecutive records rather than relying on one fixed window from the session start. Replay events appear to be written as a dense burst, so we may be able to identify a sequence of much smaller intervals—perhaps tens of milliseconds each—and stop classifying events as replay once that timing pattern is broken.

Conceptually, this would replace a single broad 2.5-second window with multiple short, data-derived windows, which may reduce both risks: missing a slow or large replay and incorrectly dropping a genuine fast first turn?

mm7894215:
@YUDONGLING Your inter-record-gap idea checked out against 25 real forked rollouts on my machine: replayed rows are flushed ≤2ms apart, while the first genuine live turn lands ≥11s later (live cadence ≥~4.6s), so a leading-burst detector that latches off at the first multi-second gap separates them cleanly — no fixed window from session start, so neither of the two failure modes above applies. Implemented in #316 along with a one-time guarded rebuild that repairs the already-inflated history. Thanks for pushing this forward.

PR #182, opened 2026-06-12
fix(grok): incremental byte-offset resume for updates.jsonl

Yanzzp999:
## Summary

Stop re-reading every Grok session's `updates.jsonl` from byte 0 on every sync: persist per-file byte offsets and resume from the last consumed byte. On a real machine (153 sessions / 474 MB) the Grok parse step drops from ~6 s warm / ~26 s cold per sync to ~295 ms — previously this cost was paid on every auto-sync (as often as every 20 s via the notify hook) and grows unboundedly with usage.

## How it works

- `readGrokUpdateTokenEvents` accepts a `prevOffsetEntry` `{size, mtimeMs, ino}` and streams from the previously consumed byte. Unchanged file (size/inode match) → stat-only fast path, no read at all. Truncation or inode change → reset to 0 so no event is missed. Mid-stream errors do not advance the offset, so the next sync retries the same range.
- `parseGrokBuildIncremental` persists offsets in `cursors.grok.updateOffsets` (keyed by `updatesPath`), rebuilt from the sessions seen each scan so entries for deleted session dirs are pruned automatically.
- Safety comes from existing semantics: `updates.jsonl` is append-only with **cumulative** `totalTokens` deduped by the session high-watermark, so re-read events can never double-count and bytes skipped by a write race are covered by the next event's cumulative total. The cursor field is additive — older versions preserve it via spread and ignore it; no migration needed.

## Tests

Two regression tests added:
- Tail-resume proven behaviorally: the consumed head is overwritten with a same-length decoy event (`totalTokens: 999`) — a full re-read would aggregate it, offset resume must only see the appended event.
- Truncation fallback: shrunken file re-reads from 0, high-watermark prevents double counting.

`node --test test/rollout-parser.test.js`: all 13 Grok tests pass.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Checklist

- [x] `npm test` passes (884/886; the 2 failures pre-exist on a clean tree — missing optional `@mongodb-js/zstd` native dep locally + a timing-sensitive event-loop test, both unrelated)
- [x] New user-facing strings go through `dashboard/src/content/copy.csv` (no UI strings touched)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved incremental processing of rollout updates to resume safely after interruptions, including support for partial writes and resumable reads by byte offset
  * Ensured correct behavior after truncation/rotation of update files so cumulative totals don’t double-count or miss events
  * Strengthened per-session progress tracking for accurate cumulative reporting across sync runs and restarts
* **Tests**
  * Added coverage for incremental resumption, partial-write handling, and truncation/rotation correctness

### DIFF
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -7891,6 +7891,27 @@ function grokEventId(value, fallback) {
   return fallback;
 }
 
+function grokFileEndsWithNewline(filePath, size) {
+  if (!(size > 0)) return false;
+  let fd;
+  try {
+    fd = fssync.openSync(filePath, "r");
+    const buf = Buffer.alloc(1);
+    const read = fssync.readSync(fd, buf, 0, 1, size - 1);
+    return read === 1 && buf[0] === 0x0a; // trailing "\n"
+  } catch {
+    return false;
+  } finally {
+    if (fd !== undefined) {
+      try {
+        fssync.closeSync(fd);
+      } catch {
+        /* ignore close failure */
+      }
+    }
+  }
+}
+
 async function readGrokUpdateTokenEvents(updatesPath, fallbackTimestamp, prevOffsetEntry) {
   if (!updatesPath) return { events: [], offsetEntry: null };
   let stat;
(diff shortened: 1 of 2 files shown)


mm7894215:
The goal of this PR is reasonable: avoid rescanning large Grok `updates.jsonl` files by resuming from a stored byte offset. However, there is one blocker that needs to be fixed before merge.

Issue: `readGrokUpdateTokenEvents()` records `offsetEntry.size` from the current `stat.size` before reading, but malformed JSON lines are silently skipped. If sync reads the file while Grok is still writing the final JSONL line, the parser skips that partial line and still advances the stored offset to the partial line end. When Grok later completes the same line, the next sync resumes from the middle of the JSON object, so that token event is never parsed or queued.

Relevant code paths:
- `src/lib/rollout.js`: `offsetEntry = { size: stat.size, ... }`
- `src/lib/rollout.js`: `JSON.parse(line)` catch path currently `continue`s
- `src/lib/rollout.js`: caller persists `updates.offsetEntry`

I confirmed this on the PR branch with a minimal repro:
1. Write the first half of a `session/update` JSON line.
2. Run `parseGrokBuildIncremental()`: result is `eventsAggregated: 0`, but the cursor offset advances to the partial line end.
3. Append the remaining bytes so the JSON line is complete.
4. Run `parseGrokBuildIncremental()` again: result is still `eventsAggregated: 0`, `sessionSnapshots` remains empty, and the queue remains empty.

Suggested fix: only advance the offset to the last complete and safely processed newline boundary. If the tail line fails JSON parsing, keep the previous offset or advance only to before the failed line. Please also add a regression test for “partial Grok updates JSON line completes on the next scan”.

Validation I ran:
- `git diff --check origin/main...origin/pr/182` passed
- `node --test test/rollout-parser.test.js` passed, 156 tests
- After setting up clean-worktree dashboard deps/dist, the previously failing dashboard/typecheck/runtime reinstall tests passed

Verdict: please do not merge yet. The partial-tail offset advancement can cause permanent undercounting and should be fixed first.

Yanzzp999:
Thanks for the careful repro — you're exactly right, and it's fixed in e05ee6d.

**Root cause:** `readGrokUpdateTokenEvents()` recorded the full `stat.size` as the resume offset even when the final line was a partial tail. The partial line failed `JSON.parse` and was skipped, but the offset still advanced past it, so once Grok completed the line the next scan resumed from the middle of the object and never parsed that event — permanent undercount.

**Fix:** only advance the offset to the last newline-terminated line. The final byte is probed, and when the file does not end on `\n` the trailing partial line's bytes are subtracted so the offset stays on a complete-line boundary and the line is re-read once finished. The read is also bounded to the stat'd size (`end: stat.size - 1`) for exact byte accounting. This stays watermark-safe, so an occasional re-read of a completed line still can't double count.

**Regression test** (`parseGrokBuildIncremental re-reads a partial Grok updates line once it completes`) follows your repro: write half a `session/update` line → scan (`eventsAggregated: 0`, offset stays at 0) → complete the line → scan again (event aggregated + queued, `totalTokens: 100`). It fails on the pre-fix code and passes after.

**Validation:**
- `node --test test/rollout-parser.test.js` → 157 passed (was 156, +1 regression)
- `npm test` → 885 passed, 0 failed
- `git diff --check` → clean
- PR CI on e05ee6d: all checks green (test + validate + build, macOS unit tests, CodeQL, label, CodeRabbit)

Ready for another look when you have a moment.

PR #200, opened 2026-06-17
feat(windows): add in-app click-to-update via GitHub releases

Hasasasa:
## What

Adds an in-app updater to the Windows tray app so users no longer have to manually download and re-run the installer. A tray-menu **Check for Updates** item flips to **Update to vX.Y.Z** when a newer GitHub release exists; choosing it downloads `TokenTracker-Setup.exe`, runs it fully silently, and relaunches the app in place.

A quiet check on launch only *surfaces* availability (a balloon + the menu label) and never auto-installs. It self-skips dev builds (no embedded server next to the exe), so a developer run is never nudged to "update" to an official release.

## Why

The macOS app already has `UpdateChecker.swift` (check → download DMG → mount/copy → relaunch). Windows had no update path at all, so every release required a manual re-download + reinstall.

## Design notes

- **Proxy direction is intentionally opposite the loopback clients.** This client talks to github.com (external), so it keeps the default proxy behaviour (`UseProxy` stays true) — proxy/VPN users (common in CN) can only reach releases through their proxy. The loopback clients in `ServerManager`/`UsagePoller` keep `UseProxy=false` (PR #152); this one must not.
- **Silent install + self-relaunch.** A silent Inno install does not relaunch the app (its `[Run]` postinstall is `skipifsilent`), so the app drives close → install → relaunch itself: it spawns a detached `cmd` that runs the installer and then restarts the current exe (`Environment.ProcessPath`) once the upgrade finishes. The spawned process is outside the server's job object, so it survives the tray quitting; the installer's existing `CloseApplications=yes` is the backstop. The installer is per-user (no UAC).
- **Parity with macOS:** the update entry lives in the tray menu only, matching macOS where it is in the menu-bar menu and not in the pet's context menu.
- **No UI in the checker.** `UpdateChecker` only manages state and raises events (mirrors `UsagePoller`); all dialogs/balloons live in `TrayApplicationContext`. Strings for 5 locales follow the existing `TrayStrings.For(locale)` pattern in a parallel `UpdateStrings`.

## Files

- `TokenTrackerWin/UpdateChecker.cs` (new) — GitHub `releases/latest` check, version compare, asset download with progress, silent install + relaunch.
- `TokenTrackerWin/UpdateStrings.cs` (new) — localized strings (en / zh-CN / zh-TW / ja / ko).
- `TokenTrackerWin/TrayApplicationContext.cs` — menu item, event wiring, launch-time silent check.

No version bumps (left to the maintainer's release process).

## Verification

- Builds clean (0 warnings / 0 errors) on the current `main` (v0.53.4) base.
- The full loop — launch-time check → download → silent install → quit → in-place relaunch — was exercised end-to-end locally (with a temporarily lowered build version to trigger the "update available" path); confirmed via `windows-host.log`.
- GitHub API reachability through a system proxy confirmed; the latest release asset is still `TokenTracker-Setup.exe`, which the matcher selects.
- The launch-time check only notifies and never auto-installs, so even if a release ever shipped a version/tag mismatch it would not loop into repeated reinstalls.

## Summary by CodeRabbit

* **New Features**
  * Added a **“Check for Updates”** option to the system tray menu.
  * Perform an automatic update check on startup and reflect status in the menu.
  * Provide a manual update flow with download progress and localized prompts/notifications.
* **Improvements**
  * Updates now install silently when available, then relaunch the app after installation.
  * Improved upgrade behavior when a tray instance is running (avoids lock-related issues).

### DIFF
--- a/TokenTrackerWin/TrayApplicationContext.cs
+++ b/TokenTrackerWin/TrayApplicationContext.cs
@@ -33,6 +33,7 @@ internal sealed class TrayApplicationContext : ApplicationContext
     private readonly ToolStripMenuItem _petSizeMedium;
     private readonly ToolStripMenuItem _petSizeLarge;
     private readonly ToolStripMenuItem _startupItem;
+    private readonly ToolStripMenuItem _checkUpdatesItem;
 
     // Right-click-the-pet context menu (separate ToolStrip items — an item can't
     // live in two menus at once).
@@ -55,6 +56,10 @@ internal sealed class TrayApplicationContext : ApplicationContext
         TrayMenuRenderer.PaletteFor(NativeTheme.ResolveIsLight(NativeTheme.CurrentPreference));
     private Font? _menuFont;
     private Font? _summaryFont;
+
+    private readonly UpdateChecker _updateChecker = new();
+    private UpdateStrings _updateStrings = UpdateStrings.For(NativeLocalization.CurrentResolvedLocale);
+    private bool _updateBalloonShown;
 
     // Lightweight UI-thread tick so the tooltip follows a currency change within a
     // couple of seconds without needing a right-click. Cheap: it only touches the
@@ -116,6 +121,7 @@ internal sealed class TrayApplicationContext : ApplicationContext
         _startupItem = CreateMenuItem("", OnToggleStartup);
         _startupItem.Checked = LaunchAtStartup.IsEnabled;
         _startupItem.CheckOnClick = false;
+        _checkUpdatesItem = CreateMenuItem("", (_, _) => OnCheckUpdatesClicked());
         _starItem = CreateMenuItem("", (_, _) => OpenInBrowser(Constants.GitHubUrl));
         _quitItem = CreateMenuItem("", (_, _) => Quit());
 
@@ -138,6 +144,7 @@ internal sealed class TrayApplicationContext : ApplicationContext
         _menu.Items.Add(_petSizeItem);
         _menu.Items.Add(CreateSeparator());
         _menu.Items.Add(_startupItem);
+        _menu.Items.Add(_checkUpdatesItem);
         _menu.Items.Add(_starItem);
         _menu.Items.Add(CreateSeparator());
         _menu.Items.Add(_quitItem);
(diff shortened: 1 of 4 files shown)


mm7894215:
Thanks — design and threading look solid (UI-free checker, all callbacks marshalled through `PostToUi`, proxy direction intentionally opposite the loopback clients). One blocker before merge though:

**The `CloseApplications=yes` backstop you rely on doesn't actually exist.** Both the `UpdateChecker.cs` doc comment and the PR description state the installer's `CloseApplications=yes` is the backstop for any instance still holding a file lock — but `TokenTrackerWin/installer.iss`'s `[Setup]` section has no `CloseApplications` (nor `RestartApplications`).

This isn't just a stale comment. In `DownloadAndInstallAsync` you call `LaunchSilentInstaller()` *before* `QuitRequested?.Invoke()`, and `Quit()` only kills `EmbeddedServer\node.exe` after the event is posted to the UI thread. So the silent installer can start overwriting `{app}` while `node.exe` still holds a lock on `{app}\EmbeddedServer\node.exe`. With `/VERYSILENT /SUPPRESSMSGBOXES` and no `CloseApplications` fallback, that file-in-use is handled silently (deferred to reboot, or aborted) — the user sees nothing, `start` relaunches the *old* exe, and since `_updateBalloonShown` is already set they won't be re-prompted. The happy path works only because node exits fast enough; it's timing-dependent.

Fix: add `CloseApplications=yes` (and `RestartApplications=no`, since we drive the relaunch ourselves) to `installer.iss`'s `[Setup]` section. Small change, and it makes the backstop the comment promises real.

Rest LGTM — will re-review once this is in.

Hasasasa:
Thanks for the review! Pushed `68383f5`.

On the `CloseApplications` backstop — it's actually already there, just in a different file than the one you checked. The Inno script is `TokenTrackerWin/installer/TokenTracker.iss` (not `installer.iss`), and its `[Setup]` section has had `CloseApplications=yes` since the original Windows-app PR (#115, line 50). So the doc comment / PR description aren't claiming something that doesn't exist — Restart Manager does detect `node.exe` running from `{app}\EmbeddedServer\node.exe` and closes it before the overwrite, which is why the happy path holds.

That said, your underlying point is right that I shouldn't lean on timing, and `RestartApplications` was genuinely missing. I've added `RestartApplications=no` to that same `[Setup]` block — since we drive the relaunch ourselves (detached `cmd` → `start`), we don't want Restart Manager also relaunching a bare `node.exe`. That makes the close→overwrite→relaunch sequence deterministic rather than relying on node exiting fast enough.

Also folded in CodeRabbit's nit: the asset is now matched by exact name `TokenTracker-Setup.exe` (the version-less, stable name `release-windows.yml` uploads) instead of the looser `EndsWith(.exe) + Contains(Setup)`, so a future co-released `.exe` can't be picked by mistake.

Build is clean (0/0). Ready for re-review.

PR #221, opened 2026-06-25
feat(linux): add Linux desktop client

snxq:
## Summary

Add a local-use Linux desktop client for TokenTracker, focused on Arch Linux + KDE Plasma.

This introduces a Tauri v2/Rust shell that reuses the existing TokenTracker CLI and React dashboard instead of adding Linux-specific dashboard UI. The client launches a bundled local Node runtime, starts `tokentracker serve` on a dynamic loopback port, opens the dashboard in a Tauri window, and provides a KDE tray menu with Open Dashboard / Quit.

## What changed

- Add `TokenTrackerLinux/` Tauri v2 desktop client
- Reuse existing `dashboard/dist` and CommonJS CLI runtime
- Bundle pinned Node.js `22.22.2` for Linux x64
- Start the local dashboard server with:
  - dynamic `127.0.0.1` port
  - `--no-open`
  - `--no-sync` to avoid blocking desktop startup on large local logs
- Add KDE tray lifecycle:
  - close window hides to tray
  - Open Dashboard restores/focuses the window
  - Quit stops the bundled Node server and exits
- Add local Arch package files under:
  - `TokenTrackerLinux/packaging/arch/tokentracker-linux`
- Add local install docs for `makepkg -si`

## Scope

This is intentionally scoped to local/self-use Linux packaging:

- Primary target: Arch Linux + KDE Plasma + x86_64
- Install method: local `makepkg -si`
- Not a public AUR package
- Not a GitHub Release asset
- Not AppImage / Flatpak / deb / rpm
- No Linux-specific dashboard pages

## Validation

Verified:

- `LANG=C LC_ALL=C npm test`
- `npm run dashboard:build`
- `cargo test -- --nocapture`
- `cargo check`
- `npm --prefix TokenTrackerLinux run build`
- `npm --prefix TokenTrackerLinux run bundle:node`
- bundled CLI help works:
  ```bash
  TokenTrackerLinux/EmbeddedServer/node \
    TokenTrackerLinux/EmbeddedServer/tokentracker/bin/tracker.js \
    --help
- makepkg successfully builds:
tokentracker-linux-0.59.0-1-x86_64.pkg.tar.zst
- built Tauri binary starts the bundled tracker.js serve
- local dashboard responds over loopback HTTP

Known limitations

The following need manual verification on an actual Arch + KDE desktop with sudo access:

- full makepkg -si pacman install lifecycle
- /usr/bin/tokentracker-linux installed-path runtime resolution
- KDE tray UI interactions:
  - close-to-tray
  - Open Dashboard restore
  - Quit cleanup

In this development session, pacman install/uninstall was blocked by sudo password prompting, and GUI tray interaction could not be fully automated.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added a TokenTracker Linux desktop client for Arch Linux + KDE Plasma with a tray menu (Open Dashboard, Quit), singleton behavior, and automatic dashboard launch.
  * Introduced local pacman packaging (PKGBUILD, .desktop launcher, icon, install hooks) to build/install/uninstall the client.
* **Bug Fixes**
  * Added a local HTTPS proxy route that relays `https://` requests to an allowlisted set of destinations with safety header filtering.
* **Documentation**
  * Updated root and Linux-client READMEs with build/run/install/uninstall instructions for the local package.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


xiufengsun (changes requested):
Please fix the following issues before requesting merge:

**Rust code issues**:
1. `server.rs`: stdout/stderr are piped without a consumer, which may block the child process. Consider switching to `Stdio::null()` or adding a draining mechanism.
2. `tray.rs`: `expect("missing default window icon")` will panic if the icon is missing. This should be a recoverable error instead of a crash.
3. `paths.rs`: The path resolution logic may accidentally use installed paths instead of development paths. Please adjust the priority.

**Script issues**:
4. `bundle-node-linux.sh`: Uses `npm install` instead of `npm ci`, which may cause dependency drift.
5. The cleanup step deletes dependency license files. Please preserve `*.txt`, `LICENSE*`, `LICENCE*` files.

**Documentation issues**:
6. Multiple path references are inconsistent and need to be unified.

Please fix these and re-request review.

xiufengsun (changes requested):
Please fix the following issues before requesting merge:

**Rust code issues**:
1. stdout/stderr are piped without a consumer, which may block the child process. Consider switching to Stdio::null() or adding a draining mechanism.
2. expect() will panic if the icon is missing. This should be a recoverable error instead of a crash.
3. The path resolution logic may accidentally use installed paths instead of development paths. Please adjust the priority.

**Script issues**:
4. Uses npm install instead of npm ci, which may cause dependency drift.
5. The cleanup step deletes dependency license files. Please preserve *.txt, LICENSE*, LICENCE* files.

**Documentation issues**:
6. Multiple path references are inconsistent and need to be unified.

Please fix these and re-request review.

mm7894215:
`/api/native-https-proxy` has no origin check, unlike this file's other mutation endpoints (`isAuthorizedLocalMutation` — loopback origin + shared-secret token) and unlike the sibling `ip-check` proxy right below it, whose own comment explains why it's GET/HEAD-only: "an open reverse-proxy any local process can abuse... anonymously POST through user IP."

This endpoint allows POST and skips that check. Credentials are stripped before forwarding and the host allowlist is tight, so it's not a credential leak — but any page open in the user's browser while the app is running can trigger a blind, unauthenticated POST to srctyff5.us-east.insforge.app through the local server, with no origin gate at all.

Since local-api.js is shared across macOS/Windows/Linux, this lands on every platform's embedded server, not just Linux. Can you add the loopback-origin check, or drop POST to match the ip-check proxy's pattern?

snxq:
@mm7894215  Thanks for catching this. After further investigation, the root cause of the HTTPS proxy turned out to be the local proxy setup (FlClash/Clash running on 127.0.0.1:7890 with libproxy routing all HTTPS traffic through it), not a glib-networking TLS bug. When the FlClash bypass list is configured to skip the InsForge API domain, WebKitGTK handles HTTPS natively without issues.

The /api/native-https-proxy endpoint and the Tauri initialization_script shim have been removed entirely. local-api.js now matches main — no diff. This also resolves the security concern you raised: there is no longer an unauthenticated POST-capable proxy endpoint on any platform.

mm7894215:
Thanks — removing the proxy endpoint entirely (and tracking the root cause to the local Clash setup) resolves the security concern, and the PR being fully self-contained under TokenTrackerLinux/ is the right shape now. Verified locally: src/ matches main, the health probe target is an unauthenticated local stub (no restart-loop risk), and reusing the Windows preferred port 17680 means the OAuth redirect URI is already allowlisted.

Not at merge standard yet though. Remaining items:

1. **Desktop entry icon**: `icons/tokentracker-linux.svg` is a generated smiley-face mark unrelated to the TokenTracker brand, and it's what KDE users see in the launcher/taskbar via the .desktop entry. Please don't introduce non-brand icons. Your window/tray `src-tauri/icons/icon.png` is already the correct official mark — install the official `dashboard/public/icon-512.png` into `hicolor/512x512/apps/` instead of the scalable SVG, or use a faithful vector of the same black-rounded-square + white bolt.
2. **Rebase**: the branch conflicts with main in `README.md` and `.gitignore` (both trivial).
3. **Generated files**: `src-tauri/gen/schemas/*.json` (~4.6k lines) are Tauri build artifacts — drop them from the PR and gitignore them.
4. Minor: `Cargo.toml` says 0.59.0 while the repo is at 0.65.1. Since this client is outside `sync-versions.cjs`, either bump it once or note that it doesn't track the repo version.

Happy to merge once 1–3 are done.

mm7894215:
Rechecked after 44f179f.

**From the June 25 list:**
1. Icon — done. Official `dashboard/public/icon-512.png` is now installed into `hicolor/512x512/apps/`.
2. Rebase — not done. Branch is still ~68 commits behind main. `Cargo.toml`/`package.json` were bumped to `0.68.1`, but that's this branch's own stale base version, not current main (`0.74.3`) — it doesn't actually resolve the version-tracking note.
3. `src-tauri/gen/schemas/*.json` — only half done. Added to `.gitignore` going forward, but the ~4.6k already-committed lines are still tracked in the diff. Needs `git rm --cached`.

**New, found while re-checking `main.rs`/`server.rs`:**

4. `start_health_monitor()` holds the `SERVER` mutex for the full duration of `server.restart()`, which internally calls `wait_for_server(..., Duration::from_secs(20))`. If the app is quit while a restart is in flight, `stop_server()` (called synchronously from the `ExitRequested` handler) blocks on the same lock — worst case the UI hangs ~20s on quit. Drop the guard before the blocking wait (or move `wait_for_server` outside the lock) so quit can't stall on a restart in progress.

**Lower-priority, can defer given the PR's stated local-only scope, but worth a doc note:**

5. `paths.rs::resolve_runtime_paths` bakes `env!("CARGO_MANIFEST_DIR")` into the binary at compile time and compares it against `current_exe()` to decide dev-vs-installed. Breaks reproducible builds, and the dev-path shortcut silently stops working if the binary is built on one machine/path and run from another (falls through to the installed-path check rather than crashing, so it's not a bug in outcome — just a footgun).
6. `PKGBUILD`'s `build()` runs `npm ci`/`npm run build` directly against `${_repo_root}` (4 dirs up from the PKGBUILD's own location) instead of staging into `$srcdir`. Fine for the stated "local `makepkg -si` from a git checkout" use case, but won't work in a clean chroot (`extra-x86_64-build`) and isn't AUR-submittable as-is. Worth a one-line note in the README under "known limitations" so nobody assumes chroot-safety later.

Please fix #4 (the only functional blocker) and close out #2/#3, then ping for re-review. #5/#6 are fine to leave as documented limitations for now.

xiufengsun (changes requested):
The overall direction is sound: the Linux client uses an isolated, thin Tauri shell while reusing the existing Node CLI and dashboard, consistent with the native-host + EmbeddedServer model used by macOS and Windows. Because the Tauri/Rust dependencies live only under `TokenTrackerLinux/`, they will not be included in the macOS DMG or Windows installer and will not increase either package's size.

However, the current implementation does not yet meet the merge standard. Please address the following items before requesting another review:

1. **Integrate Linux into unified version synchronization**
   - `main` is currently at 0.77.1, while the newly added Linux files are still hardcoded to 0.68.1:
     - `TokenTrackerLinux/package.json`
     - `TokenTrackerLinux/src-tauri/Cargo.toml`
     - `TokenTrackerLinux/src-tauri/tauri.conf.json`
     - `TokenTrackerLinux/packaging/arch/tokentracker-linux/PKGBUILD`
   - Please add these locations to `scripts/sync-versions.cjs` and the version-consistency validation so they do not continue to drift on every release.

2. **Add real Linux CI coverage**
   - The current checks are green, but none of them run Cargo, build the Tauri Linux application, or build the Linux package. They therefore do not prove that the client compiles or installs on Linux.
   - At minimum, a Linux runner should execute the Rust tests, a release build, and the Node bundle validation. If practical, it should also validate the Arch package build.
   - The absence of this platform build gate is a merge blocker.

3. **Complete the tray icon fallback**
   - In `tray.rs`, when `default_window_icon()` returns `None`, the code simply skips setting an icon, while `tauri.conf.json` also sets `bundle.active` to `false`.
   - This removes the previous panic but can still result in a blank or invisible tray item under KDE. Since closing the window leaves the tray as the primary way to restore or quit the client, this is core functionality.
   - Please explicitly load an embedded or packaged icon and verify close, restore, and quit behavior on a real KDE environment.

4. **Define and complete the OAuth/deep-link flow**
   - The desktop entry declares `x-scheme-handler/tokentracker`, but the single-instance callback ignores the launch arguments and does not handle `tokentracker://auth/callback?...`.
   - Please choose and complete one of the following approaches:
     - implement the full Linux native OAuth/deep-link flow and route the code back into the dashboard; or
     - remove the unimplemented protocol declaration, explicitly use OAuth inside the WebView, and verify the complete login flow.
   - The current declaration and runtime behavior are inconsistent.

5. **Keep the build reproducible**
   - The Node bundling script now uses `npm ci`, but the PKGBUILD still runs `npm install` for `TokenTrackerLinux`.
   - Since a lockfile is committed, this should consistently use `npm ci`.

6. **Remove generated schemas from the PR**
   - `TokenTrackerLinux/src-tauri/gen/schemas/` contains Tauri-generated files used for IDE permission autocomplete; they are not required at runtime or for compilation.
   - The project currently has no capability files that depend on them, and `.gitignore` already ignores the entire `gen/` directory.
   - Please remove them from the PR. This eliminates roughly 4,500 lines of generated review noise, and Tauri can regenerate them locally when needed.

After these changes, please provide real Arch/KDE validation results covering at least: installation, startup, close-to-tray, tray restore, second-instance launch, automatic recovery after the Node server exits unexpectedly, login, and uninstall.

Conclusion: **the architectural direction is worth continuing, but the PR is not ready to merge yet.**

xiufengsun:
Re-checked the branch at 38c1887. Good progress — the following items from the July 13 review are now resolved:

- ✅ **Version sync** — Linux version files are integrated into `scripts/version-files.cjs`, and `npm run validate:versions` is now a CI gate.
- ✅ **Tray icon fallback** — `tray.rs` embeds a fallback icon via `include_bytes!`, no more panic / blank-icon path.
- ✅ **Reproducible build** — PKGBUILD consistently uses `npm ci`.
- ✅ **Generated schemas** — `gen/schemas/` removed from the PR.
- Bonus: the server health monitor + auto-restart (mirroring the Windows `ServerManager` pattern) and the added Node tests are appreciated.

However, **3 blockers remain open**. Please address them before requesting another review:

1. **Linux CI coverage (merge blocker).** `ci.yml` still has no job that runs `cargo test`, builds the Tauri Linux app, or validates the Arch package. Only `validate:versions` was added. A green CI currently proves nothing about whether this client compiles on Linux. At minimum, add an ubuntu runner job that runs the Rust tests and a Tauri release build; ideally also a `makepkg` validation.

2. **Deep-link declaration is still inconsistent.** The `.desktop` file declares `MimeType=x-scheme-handler/tokentracker;` with `Exec=... %u`, but the single-instance callback in `main.rs` is `|app, _argv, _cwd|` — launch arguments are ignored and `tokentracker://auth/callback` is never handled. Pick one: implement the full deep-link OAuth flow, or remove the scheme-handler declaration and verify WebView-based login end to end.

3. **Real Arch/KDE validation results.** The requested lifecycle report is still missing: install, startup, close-to-tray, tray restore, second-instance launch, auto-recovery after server crash, login, and uninstall. Please run these on an actual Arch + KDE machine and post the results in this PR.

Minor: the branch is at 0.84.4 while `main` is at 0.84.6 — sync once more before merge (the new `validate:versions` gate will catch this anyway).

csmashe:
Supplemental Fedora/GNOME validation for this PR:

I tested the Linux client and prepared additional fixes in my fork:

https://github.com/csmashe/TokenTracker/tree/linux-client-review-fixes

Test environment:

- Fedora Linux 44 Workstation, x86_64
- GNOME Shell 50.3
- User-local installation under `~/.local`

Review blocker updates on my fork:

1. **Linux CI coverage:** The branch adds an Ubuntu job that builds the dashboard and bundled Node runtime, runs `cargo test --locked`, and performs a Tauri release build. It does not currently run Arch `makepkg` validation.
2. **Deep-link declaration:** The unimplemented `tokentracker://` scheme registration and `%u` argument were removed. Login completed successfully through the WebView/loopback callback.
3. **Fedora/GNOME lifecycle results:**

| Lifecycle check | Result |
|---|---|
| Install | ✅ Passed using the user-local installer |
| Startup | ✅ Passed |
| Close to tray | ✅ Passed |
| Restore from tray | ✅ Passed |
| Second-instance launch | ✅ Passed; focused/restored the existing instance |
| Server crash recovery | ⚠️ Automated recovery tests pass, but I did not perform a live manual crash/recovery test |
| Login | ✅ Passed end to end |
| Uninstall | ✅ Passed; binary, runtime, launcher, and icon were removed |
| Reinstall | ✅ Passed and installation artifacts were verified |

Additional validation:

- `npm test`: 1,996 passed, 2 skipped, 0 failed
- Rust tests: 9 passed
- Tauri release build: passed
- Version synchronization at `0.87.4`: passed
- WebKitGTK blank-window issue resolved using the DMA-BUF renderer compatibility fallback

These Fedora/GNOME results are supplemental and do not replace the requested validation on an actual Arch/KDE machine.

@snxq, if these changes are useful, I can provide the commit hashes or open a PR targeting your `worktree-linux-client` branch.

xiufengsun:
Rechecked the exact current head 38c188767a4669170f48bfc121174c051ee38869 against current main. This remains **NO-MERGE / CHANGES REQUESTED**.\n\nThe merge blockers remain: no Linux cargo/Tauri/Arch validation job, the tokentracker:// desktop declaration still does not match the ignored launch arguments, and the requested real Arch/KDE lifecycle evidence is missing. Validation posted from a fork does not change this PR head. Please apply the fixes to this branch, push a new head, and request re-review.

snxq:
@xiufengsun,  apologies for not giving this PR enough attention lately. I’ll continue to follow the progress of this PR and work on the outstanding fixes. 

@csmashe,  you’re very welcome to open a PR against my branch with your changes.

AmineDjeghri:
Thank you for this PR.
Do you have a date on when this will be merged ?

Issue #224, opened 2026-06-28
[bug]: Antigravity shows "Not Connected" due to legacy Gemini CLI OAuth path error


xMinhx:
### Preflight checklist

- [x] I searched [existing issues](https://github.com/mm7894215/TokenTracker/issues) and didn't find a duplicate.
- [x] I'm on the [latest version](https://github.com/mm7894215/TokenTracker/releases).

### What happened?

TokenTracker is failing to fully connect to the Antigravity backend, displaying a "Not Connected" status on the dashboard. Interestingly, the Antigravity symbol is completely visible in the token counts UI component (meaning the UI assets and baseline references are active), but the underlying authorization bridge crashes with a legacy path error:

Additional Error: Gemini API error: Could not find Gemini CLI OAuth configuration.

I verified that this issue is still present on the absolute latest release after running `npm i -g tokentracker-cli@latest`. I previously had `gemini-cli` installed on this machine but have since uninstalled it in favor of the new official native Antigravity CLI tool (`agy`). 

**Environment & Paths:**
* Antigravity Executable: `/home/user/.local/bin/agy`
* Antigravity Config Directory: `/home/user/.gemini/antigravity-cli`
* OS Environment: WSL 2 (Ubuntu)

### Steps to reproduce

1. Ensure legacy `gemini-cli` is completely uninstalled.
2. Install Google Antigravity to `~/.local/bin/agy` (initializing its config in `~/.gemini/antigravity-cli`).
3. Run `tokentracker` and launch the dashboard.
4. Note that the Antigravity icon appears correctly in the token counts area, but the main module status remains "Not Connected" due to the hardcoded OAuth configuration lookup path failure.

### Expected behavior

Since TokenTracker natively displays the new Antigravity icon in the layout, the underlying authorization hook should look for active OAuth tokens inside the updated `~/.gemini/antigravity-cli` directory instead of falling back to the legacy `gemini-cli` config directory.

### Is this a regression?

Not sure

### TokenTracker version

v0.60.0

### Node.js version

v24.16.0

### How did you install it?

npm / npx (CLI)

### OS / version

Windows 11 / WSL2 (Ubuntu-26.04

### Output of `tokentracker doctor` (if applicable)

```shell
Doctor report

- [OK] runtime.base_url - base_url set
- [WARN] runtime.device_token - device token missing
- [OK] runtime.dashboard_url - dashboard_url set
- [OK] runtime.http_timeout_ms - http timeout resolved
[...]

PR #227, opened 2026-06-28
fix: seed antigravity model from settings.json and use RetrieveUserQuotaSummary for 4-window display

xMinhx:
## Problem

Two issues with Antigravity quota tracking:

**Before** — Antigravity CLI not detected or detected with missing usage data:

1. **`"antigravity-unknown"` model buckets** — The parser determines the model from `USER_SETTINGS_CHANGE` events in the transcript. Conversations that use the default model (no manual switch) have no such event, so `currentModel` stays `null` → falls through to `"antigravity-unknown"` in the dashboard. Cost attribution is broken for these rows.

2. **Missing 4-window display** — The old `GetUserStatus` endpoint returns a flat model list that we arbitrarily split into 2 windows. The `RetrieveUserQuotaSummary` endpoint returns proper structured quota groups with `bucketId`s (`3p-weekly`, `3p-5h`, `gemini-weekly`, `gemini-5h`), giving us 4 distinct bars matching what agy CLI shows.

before-not-detected

before-missing-data

**After** — 4 proper bars with correct model attribution and cache and live tracking for antigravity cli:

after

3. **Label clarity** — `"Weekly"` is ambiguous; `"7d"` is the standard shorthand used elsewhere in the dashboard.

### `src/lib/rollout.js`
- Added `readAntigravityDefaultModel(filePath)` — walks up 5 `path.dirname()` levels from the transcript path to reach the variant root (`~/.gemini/antigravity-cli/`), reads `settings.json`, seeds the model via `normalizeAntigravityTranscriptModel()`.
- Called when `currentModel` is null at parse start (fresh file or no cached model). A `USER_SETTINGS_CHANGE` event found later in the transcript still overrides the seed.

### `src/lib/usage-limits.js`
- Added `normalizeAntigravityQuotaSummary()` — parses `RetrieveUserQuotaSummary` response into `primary_window` (`3p-weekly`), `secondary_window` (`3p-5h`), `tertiary_window` (`gemini-weekly`), `quaternary_window` (`gemini-5h`).
- Updated `fetchAntigravityLimits()` — tries `RetrieveUserQuotaSummary` first, falls back to `GetUserStatus`.
- **Response is cached to disk** so the dashboard still shows usage data even when agy is not actively running — the `live` / `cached` badge on the Antigravity card indicates whether data was fetched fresh or served from cache.
- Cache functions, port probing (HTTP fallback for agy CLI), and locale labels updated for 4-window display.
- Added `isBinaryAvailable("gemini")` guard to `fetchGeminiLimits()` — returns `configured: false` when the gemini binary is not installed even if the config folder exists.
- Cleanup: stripped agy OAuth extraction functions (agy-specific token fallback was scope creep; this is handled differently on `main`).

### Dashboard locale files
- Old 3-key set (`antigravity_claude`, `antigravity_gpro`, `antigravity_flash`) replaced with 4 keys:
  - `limits.label.antigravity_claude_weekly`: `"Claude - Weekly"` → `"Claude - 7d"`
  - `limits.label.antigravity_claude_5h`: `"Claude - 5h"`
  - `limits.label.antigravity_gemini_weekly`: `"Gemini - Weekly"` → `"Gemini - 7d"`
  - `limits.label.antigravity_gemini_5h`: `"Gemini - 5h"`
- Added `limits.label.antigravity_live` and `limits.label.antigravity_cached` for the status badge.

Updated in `copy.csv`, `ja/core.json`, `zh/core.json`, `zh-TW/core.json`, `ko/core.json`.

### `dashboard/src/ui/dashboard/components/UsageLimitsPanel.jsx`
- Added Antigravity status badge showing **live** (green, fresh fetch) or **cached** (gray, served from cache) next to the provider name.

### `dashboard/src/ui/dashboard/components/usage-limits-provider-specs.js`
- Updated antigravity spec from 3 windows (`claude`, `gpro`, `flash`) to 4 (`claude-weekly`, `claude-5h`, `gemini-weekly`, `gemini-5h`) and added `quaternary_window`.

## Verification
- All 66 usage-limits tests pass
- A/B comparison against `origin/main`: 9 `antigravity-unknown` buckets on main, zero on this branch
- Copy registry validates clean
- `validate:ui-hardcode` guardrails pass
- Tested against live agy CLI endpoint at `RetrieveUserQuotaSummary`

## Summary by CodeRabbit

* **New Features**
  * Added relative-time translations for “minutes/hours/days ago” in Japanese, Korean, and Chinese.
  * Updated Antigravity usage-limit labels to reflect provider/time-window variants, including live vs cached.
  * Enhanced the Usage Limits panel with provider badge display and clearer per-window labeling.

* **Bug Fixes**
  * Antigravity model selection now falls back to saved settings when transcript info is missing.
  * Improved detection and status reporting for Antigravity (installed-but-not-running) and Gemini (not configured when credentials are absent).
  * Extended cached-usage support to the additional Antigravity window.


mm7894215:
Thanks for this — the 4-window display and the `antigravity-unknown` fix are solid, and the before/after verification is appreciated. A few things to address before we can merge:

**1. (blocking) Process detection is too broad and misfires on sibling Codeium products.** In `isAntigravityCommandLine`, this early return:
```js
if (/language_server(?:_macos(?:_(?:arm|x64))?|\.exe)?\b/i.test(command)) return true;
```
matches any `language_server` process because the suffix group is optional. Since Antigravity shares the Codeium `language_server_macos_arm` binary name with Windsurf, a Windsurf user would be detected as running Antigravity and we'd probe the wrong ports. This also makes the `--app_data_dir antigravity` / `/antigravity/` guard on the next line dead code. Please keep the Antigravity-specific markers as the gate (and match `agy` by basename so absolute paths like `/usr/local/bin/agy` still work).

**2. Non-users now get a red error instead of a neutral "not connected".** The not-running branch returns `{ configured: true, error: "Antigravity IDE is not running…" }`. On the web Limits page every provider card renders, so a user who never installed Antigravity sees a red error inconsistent with every other provider's unconfigured state. Since `detectAntigravityProcess` can't tell "installed but closed" from "never installed", please only surface that error when there's install evidence (e.g. `~/.gemini/antigravity` exists), otherwise fall back to `{ configured: false }`.

**3. No test for `readAntigravityDefaultModel()`.** The 5-level `dirname` walk + `settings.json` read is fragile and untested — sibling providers (codebuddy/workbuddy) have settings.json-fallback tests; please add one.

Minor (non-blocking): `m_ago`/`h_ago`/`d_ago` in zh/ja/ko are still `{{n}}m`/`{{n}}h`/`{{n}}d` with no localized "ago" wording.

The rest looks good — CI is green and the usage-limits coverage is thorough.

mm7894215:
Thanks for this — the 4-window model is exactly right, and I confirmed it against the live Antigravity quota UI (two groups: *Gemini Models* and *Claude and GPT models*, each with a Weekly + 5-hour limit). I pushed three commits on top to get it merge-ready:

1. **Gemini regression fix (important).** The new `isBinaryAvailable("gemini")` guard at the top of `fetchGeminiLimits` ran *before* credential detection, so an authenticated Gemini user whose `gemini` binary isn't on the launchd PATH (the standard macOS menu-bar environment — see #224) had their Gemini card silently disappear. Reworked it to a credential-aware guard: skip only when there are **no** OAuth credentials **and** no binary. This still suppresses the spurious Gemini card for an Antigravity-only `~/.gemini` folder, which was the original intent. Added a regression test.
2. **Quota-summary fallback.** `normalizeAntigravityQuotaSummary` now treats "groups parsed but none of the known bucketIds matched" as a parse failure, so it falls back to `GetUserStatus` instead of rendering an empty, error-free card if Antigravity renames the buckets upstream.
3. **macOS native sync.** The window semantics changed from 3 (Claude/G-Pro/Flash) to 4, but the native menu-bar/popover side still used the old labels and a 3-window availability fingerprint. Synced it: added `quaternaryWindow`, four menu-bar metrics (Ag Cl 7d / Ag Cl 5h / Ag Gm 7d / Ag Gm 5h), popover labels, the fingerprint, and the JS metric mirror.

Merging once CI is green. Ships in the next release.

PR #228, opened 2026-06-29
fix: reduce macOS idle energy usage

lamther:
## Summary
Reduce idle work in the macOS menu bar app by separating lightweight background reloads from full local syncs, and tighten incremental parser behavior around unchanged/truncated Codex and Claude inputs.

## Scope
- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Checklist
- [x] `npm test` passes
- [x] PR description explains why, not just what

## Changes
- Gate macOS background full syncs with `BackgroundRefreshPolicy` while keeping lightweight local data reloads on the existing 5-minute cadence.
- Keep launch/manual sync paths immediate, so explicit user sync still runs right away.
- Throttle the desktop pet animation timeline when idle/sleeping and pause it when the floating window is hidden.
- Reuse already-read file stats in Codex/Claude parsers and skip unchanged EOF files only when project attribution is disabled.
- Reset parser offsets to zero on same-inode truncation instead of treating truncated files as unchanged.
- Add regression coverage for background refresh policy, unchanged parser skips, same-inode truncation, and project-enabled unchanged-file semantics.

## Tests run
- `node --test test/rollout-parser.test.js` — 171/171 passed
- `npm test` — 1015/1015 passed
- `npm run dashboard:build` — passed
- `npm run validate:copy` — passed with existing unused-copy warnings
- `npm run validate:ui-hardcode` — passed
- `npm run validate:guardrails` — passed
- `node --test test/architecture-guardrails.test.js` — 4/4 passed
- `npm --prefix dashboard run typecheck` — passed
- `xcodebuild test -project TokenTrackerBar.xcodeproj -scheme TokenTrackerBarTests -destination 'platform=macOS'` — 46/46 passed

## Known gaps / out of scope
- This is not a full source-scoped sync redesign; default sync still scans all enabled source families.
- Provider visibility/order/display preferences remain presentation-only and do not disable collection or sync.
- Project attribution backfill for files parsed before project attribution is enabled should be handled in a follow-up with a separate attribution cursor or explicit forward-only behavior.
- Runtime energy impact was validated by code/test coverage, not Instruments measurements.

---

<details>

### Risk layer triggers

### Rules / invariants
- Manual sync remains immediate.
- Background reloads can refresh already-local data without forcing full sync every 5 minutes.
- Project-enabled parser paths must not use the unchanged-file skip that would bypass project attribution work.

### Boundary matrix
- macOS launch/manual sync: full sync still runs immediately.
- macOS idle refresh: lightweight reload every 5 minutes, full sync only after policy interval.
- Parser without project queue: unchanged EOF file can be skipped.
- Parser with project queue: unchanged file still enters parser path to preserve project semantics.

</details>

## Summary by CodeRabbit

* **New Features**
  * Added window-visibility state for the floating companion and wired it into reactive UI behavior.
  * Background syncing now follows configurable timing rules and records the last successful sync to choose between sync vs. reload.

* **Bug Fixes**
  * Improved incremental rollout parsing resilience to truncation and prevents re-parsing files already fully processed.
  * Updated animation redraw timing to throttle adaptively and pause when the companion window isn’t visible/idle.

* **Tests**
  * Added unit tests for background refresh timing and expanded incremental rollout parser coverage for project-state gating.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Models/BackgroundRefreshPolicy.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Models/BackgroundRefreshPolicy.swift
@@ -1,0 +1,16 @@
+import Foundation
+
+enum BackgroundRefreshPolicy {
+    static let defaultRefreshInterval: TimeInterval = 300
+    static let defaultSyncInterval: TimeInterval = 1_800
+
+    static func shouldRunSync(
+        now: Date,
+        lastSyncAt: Date?,
+        syncInterval: TimeInterval = defaultSyncInterval
+    ) -> Bool {
+        guard syncInterval > 0 else { return false }
+        guard let lastSyncAt else { return true }
+        return now.timeIntervalSince(lastSyncAt) >= syncInterval
+    }
+}
--- a/TokenTrackerBar/TokenTrackerBar/Services/DesktopPetWindowController.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Services/DesktopPetWindowController.swift
@@ -75,6 +75,7 @@ final class DesktopPetWindowController: NSObject, NSWindowDelegate {
     func show() {
         let panel = panel ?? makePanel()
         self.panel = panel
+        uiState.isWindowVisible = true
         panel.orderFrontRegardless()
         UserDefaults.standard.set(true, forKey: Self.showDefaultsKey)
         keepActive()
@@ -82,6 +83,7 @@ final class DesktopPetWindowController: NSObject, NSWindowDelegate {
 
     func hide() {
         panel?.orderOut(nil)
+        uiState.isWindowVisible = false
         UserDefaults.standard.set(false, forKey: Self.showDefaultsKey)
         sleepTimer?.invalidate()
         sleepTimer = nil
@@ -478,6 +480,7 @@ private struct DesktopPetHost: View {
 final class PetWindowState: ObservableObject {
     static let alwaysAllowed = PetWindowState()
     @Published var bubbleAllowed = true
+    @Published var isWindowVisible = true
     @Published var isTucked = false
     @Published var isHovered = false
     @Published var isRightEdge = true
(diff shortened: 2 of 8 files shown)


mm7894215:
Hello! Thank you very much for your excellent work on idle energy optimization and making the incremental parser more robust.

To ensure the best possible user experience and clean CI/CD checks, we have pushed a few additional refinements to your branch before merging:
1. **Restore default sync interval to 5 minutes**: A 30-minute background sync interval could cause visible latency for users who expect real-time feedback on their token consumption, making it seem like the tracker has failed. Since you introduced an outstanding `Early-skip` logic (based on same-inode and size checks) in `rollout.js`, the scanner can quickly exit in less than 1ms when idle. Reverting the interval back to 5 minutes preserves the original responsive UI experience while fully retaining your power-saving benefits.
2. **Resolve CodeQL race condition alert**: The use of file descriptor handles via `fs.open(rolloutPath, "a")` in `test/rollout-parser.test.js` triggered a CodeQL TOCTOU (file system race condition) warning. We reverted this specific append test logic back to a simpler `fs.appendFile` call. This resolves the static analysis failure while keeping the tests passing.
3. **Complete Swift docstrings**: We added standard documentation comments to the new `BackgroundRefreshPolicy` model and the new property/method in `DashboardViewModel` to resolve the Jazzy docstring coverage warning.

Local builds, all 1015 unit tests, and all remote CI checks have successfully passed. This pull request has now been merged into `main`. Thank you again for your contribution!

lamther:
Thank you for the thoughtful review and for merging this PR.

The 5-minute responsiveness trade-off makes sense. The early-skip path was intended to make the existing cadence cheaper rather than forcing the UI to feel stale, so I agree with keeping the user-visible refresh behavior responsive.

If I take this further, I’ll keep it as a separate PR focused on source-scoped sync / attribution improvements, with benchmarks and CI coverage kept separate from this merged idle-energy pass. Thanks again for the quick review and cleanup.

mm7894215:
Hi @Arkycher! We wanted to let you know that your pull request has been successfully merged, and these improvements have been officially published in the new release **v0.63.1** (both NPM CLI package and desktop bundles).

Thank you again for your outstanding contribution to reducing macOS idle energy usage and strengthening rollout parser robustness!

PR #232, opened 2026-06-30
Reduce Gemini idle sync parsing

lamther:
## Summary
- add a Gemini unchanged-file fast path using inode + size + mtime metadata, so idle syncs can skip JSON parsing when project attribution is not active
- keep a separate per-file Gemini project cursor so enabling project attribution can backfill project buckets without replaying main hourly usage
- clear matching per-file project cursors during project purge so public -> blocked/purged -> public can rebuild project buckets instead of being skipped
- preserve legacy cursor compatibility by requiring the new size/mtime metadata before fast skipping

## Tests
- `node --test test/project-usage-purge.test.js test/rollout-parser.test.js` (183/183 passed)
- `npm test` (1058/1058 passed)

## Notes
- Parser/sync-only scope; no macOS/Swift changes.
- `codex-workflow` read-only review identified the stale project cursor after purge case; this PR includes the fix and regression coverage.

## Summary by CodeRabbit

* **New Features**
  * Improved incremental processing for Gemini session data, including better handling of unchanged files and per-project progress.
  * Added support for rebuilding project-level buckets after project-specific data is cleared.

* **Bug Fixes**
  * Preserved accurate totals even when some messages do not include per-message usage details.
  * Reduced unnecessary reprocessing by using stronger file-change detection.

### DIFF
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -1185,6 +1185,7 @@ async function cmdSync(argv) {
           projectQueuePath,
           projectQueueStatePath,
           projectState: cursors.projectHourly,
+          cursors,
         });
         meta.purge_pending = false;
         meta.purged_at = new Date().toISOString();
--- a/src/lib/project-usage-purge.js
+++ b/src/lib/project-usage-purge.js
@@ -8,9 +8,10 @@ async function purgeProjectUsage({
   projectQueuePath,
   projectQueueStatePath,
   projectState,
+  cursors,
 }) {
   if (!projectKey || !projectQueuePath || !projectQueueStatePath || !projectState) {
-    return { removed: 0, kept: 0, removedBuckets: 0 };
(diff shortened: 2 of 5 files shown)


mm7894215:
Hi @lamther,

Thank you so much for submitting this high-quality PR (#232)! 

We ran a performance benchmark using our actual local Gemini session logs (81 files) and the results show a massive **~90x speedup** for incremental syncs when files are unchanged (sync time dropped from 176ms to just 2ms). It completely eliminates redundant JSON parsing and file reads, saving significant CPU and disk I/O.

Your cursor dual-track mechanism and the project cursor cleanup upon purge are also very elegant, ensuring solid statistics reconstruction and ultimate consistency. 

This is a very neat and helpful optimization! Thank you again for your outstanding contribution to TokenTracker. 👏

PR #234, opened 2026-06-30
Optimize scoped sync and Codex session scanning

lamther:
## Problem

Auto sync could still do more work than needed in several background paths: known notify/retry/OpenClaw triggers could fall back to broader provider scans, and Codex session discovery could repeatedly walk unchanged historical day directories.

## Summary

- Scope known auto sync triggers so notify, retry, and OpenClaw paths only scan the relevant provider.
- Cache Codex day session inventory so unchanged historical session directories are not repeatedly walked.
- Keep Codex rollout parsing guarded by existing stat, inode, offset, truncation, and project freshness checks.
- Scope OpenClaw agent and gateway lifecycle syncs with `--from-openclaw` to avoid accidental full provider scans.
- Preserve full scans for manual sync, unknown sources, plain full auto sync, and legacy retries without `syncSource`.
- Keep Codex rescan repair guarded against partial or unsafe rebuilds.

## Compatibility / Safety

- Manual sync and unknown-source auto sync still run full provider scans.
- Legacy retries without `syncSource` still use the previous full-scan behavior.
- The Codex inventory cache only avoids repeated directory enumeration; individual rollout files still go through the incremental parser and freshness checks.
- Corrupt or path-like cached filenames are rejected and rebuilt from disk.

## Verification

- `git diff --check`
- `node --test test/openclaw-session-plugin.test.js test/codex-source-scoped-cache.test.js` — 23/23 passing
- `node --test test/rollout-parser.test.js test/sync-codex-rescan-repair.test.js test/codex-source-scoped-cache.test.js` — 209/209 passing
- `npm test` — 1091/1091 passing
- `codex-workflow` final convergence: `20260701-pr234-post-cache-coverage-convergence`, status `done`, final vote `7/7`, hard gaps `0`

## Summary by CodeRabbit

* **New Features**
  * Added a `--source` option to run source-scoped auto syncs instead of scanning everything.
  * Rescan repairs now safely rebuild Codex project usage and update both queues/offset state consistently.
* **Bug Fixes**
  * Auto-retry backlog detection no longer misuses project queue data after cloud draining.
  * Improved handling of stale/missing project context so unchanged work is skipped without side effects.
  * Notify and OpenClaw lifecycle syncs now correctly preserve the originating sync scope.
* **Performance**
  * Enhanced rollout caching and Codex cold-scan skipping to reduce unnecessary file statting/parsing.
* **Tests**
  * Expanded coverage for source scoping, repairs, caching, and cold-scan behavior.

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -992,10 +992,11 @@ try {
     try { fs.writeFileSync(throttlePath, String(now), 'utf8'); } catch (_) {}
     const hasLocalRuntime = fs.existsSync(trackerBinPath);
     const hasLocalDeps = fs.existsSync(depsMarkerPath);
+    const syncArgs = ['sync', '--auto', '--from-notify', '--source', source];
     if (hasLocalRuntime && hasLocalDeps) {
-      spawnDetached([process.execPath, trackerBinPath, 'sync', '--auto', '--from-notify']);
+      spawnDetached([process.execPath, trackerBinPath, ...syncArgs]);
     } else {
-      spawnDetached(['npx', '--yes', fallbackPkg, 'sync', '--auto', '--from-notify']);
+      spawnDetached(['npx', '--yes', fallbackPkg, ...syncArgs]);
     }
   }
 } catch (_) {}
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -157,6 +157,43 @@ const DROID_DUP_SESSION_REPAIR_KEY = "droidDupSessionInflationRepair_2026_06";
 // one-time repair purges all source=mimo data from the local queues + cursor
 // state so the next sync (providerID-filtered reader) rebuilds it correctly.
 const MIMO_PROVIDER_REPAIR_KEY = "mimoClaudeMislabelRepair_2026_06";
+const AUTO_SYNC_SOURCE_ALIASES = new Map([
+  ["code", "every-code"],
+  ["everycode", "every-code"],
+  ["kilo", "kilo-cli"],
+  ["kilo-code", "kilocode"],
+  ["kimi_code", "kimi-code"],
+  ["roo-code", "roocode"],
+]);
+const AUTO_SYNC_SOURCES = new Set([
+  "antigravity",
+  "claude",
+  "codebuddy",
+  "codex",
+  "copilot",
+  "craft",
+  "cursor",
+  "droid",
+  "every-code",
+  "gemini",
+  "goose",
+  "grok",
+  "hermes",
+  "kilo-cli",
+  "kilocode",
+  "kiro",
+  "kimi",
+  "kimi-code",
+  "mimo",
+  "omp",
+  "opencode",
+  "openclaw",
+  "pi",
+  "roocode",
+  "workbuddy",
+  "zcode",
+  "zed",
+]);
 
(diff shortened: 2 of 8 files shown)


mm7894215:
Ran an A/B test against real data instead of a synthetic fixture — same `~/.tokentracker` + `~/.codex` snapshot, one copy on `main`, one on this branch, both re-synced after the initial pass.

Idle re-sync (steady state):
- `main`: ~0.2s, 969 files touched
- this branch: ~0.35s, 115 files touched

Fewer files touched but higher wall time — the per-file freshness check (stat on the cached git-config path) appears to cost more than main's unconditional-but-cheap early return. Ran this 3x on each side, consistent both times.

Can you share the benchmark behind the "optimize" claim? Want to compare methodology in case I'm missing something.

lamther:
Thanks for checking this against real data. I reproduced the same issue locally, and your concern was correct: my previous "optimize" claim was incomplete.

Benchmark method:
- Same real local TokenTracker + Codex snapshot for both candidates.
- Isolated temp HOME per candidate, with the device token removed so no upload path is involved.
- Real Codex data was read-only via CODEX_HOME.
- Each candidate got one warm sync, then 5 idle `sync --auto` runs.
- I instrumented filesystem calls to separate directory inventory, rollout-file reads, and git config freshness checks.

Results before the follow-up fix:

| candidate | idle median | Codex dirs read | rollout stream reads | git config stat calls |
|---|---:|---:|---:|---:|
| main @ 0188306 | ~0.70s | 101 | 0 | 0 |
| PR @ 94f179a | ~3.46s | 9 | ~2,373 | ~17,898 |

So the day inventory cache did reduce directory walks, but the project freshness path more than offset it. The regression was caused by two things:
1. repeated freshness stats against the same `.git/config` paths during one idle sync;
2. sessions with no project context being treated as permanently stale, causing old unchanged files to be streamed again on every idle sync.

I pushed a follow-up commit, `a246a6a fix: cache idle Codex project freshness checks`, that:
- memoizes `.git/config` freshness checks within a sync run;
- records a `checkedAtMs` timestamp for absent project context and throttles those re-probes with a 24h TTL, so we still recover later if a repo appears, but we do not re-read historical sessions every 5 minutes.

Results after the fix:

| candidate | idle median | Codex dirs read | rollout stream reads | git config stat calls |
|---|---:|---:|---:|---:|
| main @ 0188306 | ~0.57s | 101 | 0 | 0 |
| PR working tree @ a246a6a | ~0.59s | 9 | 2 | 56 |

Validation run locally:
- `node --test test/rollout-parser.test.js`
- `node --test test/sync-codex-rescan-repair.test.js test/codex-source-scoped-cache.test.js`
- `npm test` (1092 tests passed)

Net: the original PR revision did regress idle wall time on my real dataset; the latest commit removes that regression while keeping the reduced day-directory inventory behavior.

mm7894215:
Merged into `main` (006b974) — thanks, this is a solid win: steady-state Codex syncs now finish in the hundreds of ms. Taken over on top of your commits with two small follow-ups — dropped a duplicated `buildProjectFileContext`/`isProjectFileContextFresh` pair, and routed the Claude no-cwd path through `buildProjectFileContext(null)` so both absent-context branches persist `checkedAtMs` consistently. Full test suite green (1130/0).

PR #123, opened 2026-06-01
feat(hermes): WSL auto-discovery — distro probe + UNC fallback + status (#87)

xiufengsun:
Implements the three enhancements tracked in #87. Opened as a **draft** because the owner has no Windows + WSL hardware — needs real WSL1/WSL2 validation before merge (@simonlpaige kindly offered to test).

## What's here

**1. `wsl.exe -l -q` → `-l -v` distro auto-probe**
On Windows, when no native `%LOCALAPPDATA%\hermes` install is found, `resolveHermesPath()` now probes WSL distros and looks for `~/.hermes` over UNC. Zero-config for WSL users.

**2. `\\wsl$\` ↔ `\\wsl.localhost\` fallback**
Both aliases are attempted. Order keys off the version reported by `wsl.exe -l -v`: WSL2 tries `\\wsl$\` first, WSL1 tries `\\wsl.localhost\` first — per @simonlpaige's note that WSL1 sometimes only resolves via `\\wsl.localhost\`.

**3. Hermes in `tokentracker status`**
Hermes was **completely absent** from `status` output. Added it to JSON, the `--light` table and the human-readable list, surfacing the resolved path (UNC included) and, on Windows, the discovered distros + their versions — so "why didn't it sync my hermes" is debuggable at a glance.

## Incorporating review feedback (@simonlpaige)
- Username is read with `wsl.exe -d <distro> -e whoami` instead of concatenating `%USERNAME%` (Linux user ≠ Windows user).
- Version-aware UNC ordering (WSL1 vs WSL2).
- `status` shows the distro version.

## Safety
Discovery is fail-safe: missing `wsl.exe`, no distros, or a failing `whoami` all fall back to `~/.hermes`. macOS/Linux paths are untouched (the probe is gated on `process.platform === "win32"`).

## Tests
`test/rollout-parser.test.js` covers parsing (incl. UTF-16 NUL/BOM noise + header skip), default-distro ordering, per-distro `whoami`, and version-aware UNC ordering — all with injected `runWsl`/`existsSync` deps so they run on any platform. Full suite: 759 passing; guardrails + ui-hardcode validators clean.

## Not done here
No version bump / release — that follows after WSL validation confirms the UNC probing actually resolves on real hardware.

Refs #87

## Summary by CodeRabbit

* **New Features**
  * Added Hermes status information to command output, including installation state and resolved location.
  * On Windows, status now also shows detected WSL distros when applicable.

* **Bug Fixes**
[...]

PR #237, opened 2026-06-30
fix(hermes): prefer WSL over native Hermes when both exist

xMinhx:
## Priority fix for WSL Hermes auto-discovery

**Problem:** Original PR checked `%LOCALAPPDATA%\hermes` first and returned early, skipping the WSL probe entirely. Since Hermes Desktop (Windows native) and Hermes Agent (WSL) have independent `state.db` files, both should be scanned — and **WSL should win** when both exist (development happens inside WSL).

**Fix:** Changed the priority chain so WSL is probed before the native Windows install. The native path still acts as a fallback when no WSL distro has Hermes.

**Validation:** Tested on Windows 11 + WSL2 Ubuntu-26.04. `tokentracker status` shows:
```
Hermes Agent: state.db found (\\wsl$\Ubuntu-26.04\home\minh\.hermes); WSL distros: Ubuntu-26.04 (v2), docker-desktop (v2)
```

Also fixed two tests that failed on Windows by injecting a `noWsl` stub via the `deps` parameter so `resolveHermesPath()` does not attempt to run `wsl.exe` during tests.

All 6 WSL-specific tests pass on both Linux and Windows. Full validation results on the parent PR.

### DIFF
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -969,64 +969,65 @@ async function cmdSync(argv) {
     }
 
     if (!opts.auto) {
-      const totalParsed =
-        parseResult.filesProcessed +
-        openclawResult.filesProcessed +
-        claudeResult.filesProcessed +
-        geminiResult.filesProcessed +
-        antigravityResult.filesProcessed +
-        opencodeResult.filesProcessed +
-        cursorResult.recordsProcessed +
-        kiroResult.recordsProcessed +
-        kiroCliResult.recordsProcessed +
-        hermesResult.recordsProcessed +
-        kimiResult.recordsProcessed +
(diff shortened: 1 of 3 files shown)


mm7894215:
The WSL-over-native priority fix landed via #123 (cherry-picked your commit, credit preserved) — merged to main just now.

Dropped the second commit (per-provider sync output breakdown) since it's unrelated to WSL and has its own bug: opencodeDbResult is shown in the per-provider rows but missing from the totalFiles/totalBuckets sums, so the printed total wouldn't match the rows above it. Feel free to open that as its own PR with the total fixed.

Thanks for tracking down the priority bug and testing it on real WSL2.

PR #249, opened 2026-07-02
feat: WSL auto-discovery + TOKENTRACKER_WSL_MODE config for all Windows providers

### LINKED ISSUE #238: feat: extend WSL auto-discovery to all providers
## Extend WSL auto-discovery to all providers

PR #123 + #237 validated WSL auto-discovery for Hermes Agent on Windows 11 + WSL2. The same pattern should extend to every provider that runs inside WSL (Claude Code, Cursor, Codex, Gemini, Kiro, etc.).

### Why

Windows developers running AI agents inside WSL currently must set `TOKENTRACKER_*_HOME` manually for each provider. Since WSL state is accessible via `\\wsl$\<distro>`, the auto-probe pattern from PR #123 can be generalized.

### What needs doing

1. **Extract shared WSL probe utilities** — `probeWslDistros`, `discoverWslHermesHome`, `parseWslListVerbose` are currently hermetic to Hermes. Move to a shared helper so every provider can use them.
2. **Add WSL fallback per provider** — each `resolve*Path()` function should check WSL distros when the native Windows path is absent, similar to how `resolveHermesPath` now works.
3. **Status display** — each provider in `tokentracker status` should surface its WSL-resolved path + distro info.
4. **Configuration toggle** — per-provider or global setting to enable/disable WSL probing, avoiding `wsl.exe` overhead (~500ms) for users who don't need it.

### Open question: merge vs. exclusive

When a provider is installed **both** natively and inside WSL (different `state.db` files, different sessions), should TokenTracker:

- **Merge both** — aggregate usage from both sources for a complete picture?
- **Track one** — prefer WSL or native exclusively?

For Hermes specifically, native and WSL have truly independent sessions (different OS, different state files, different contexts). Merging them gives the user a complete view of their Hermes usage across environments. But **there may be a conflict with subscription/usage limits** — if the dashboard caps usage per-provider, merging could show more usage than expected against a plan tier.

For other providers (Claude Code, Cursor), the overlap risk is different — they may or may not share state between Windows and WSL. Each provider's merge strategy should be evaluated individually.

Decision needed per provider before extending:
- What constitutes "the same session" across WSL vs. native?
- Should usage be merged or kept exclusive?
- How does WSL discovery interact with subscription plan limits?

### Reference

- PR #123 (original Hermes WSL feature)
- PR #237 (priority fix: WSL > native)
- `src/lib/rollout.js` — `discoverWslHermesHome()` and helpers

xMinhx:
## Summary

Completes WSL auto-discovery and priority configuration for all Windows provider resolvers, replacing  with the more expressive  environment variable.

### Core module: `src/lib/wsl-probe.js`
- 12 exported helpers: 6 WSL core utilities + 6 mode arbitration functions
- `normalizeWslMode` — case/whitespace/underscore tolerant (`WSL_ONLY`, `wsl-first` both work)
- `pickWin32Path` — sole arbitration point; returns native or WSL path based on mode
- `shouldProbeWsl` / `shouldProbeNative` — prevent probing excluded sources in `wsl-only` / `native-only`
- All helpers are O(1) function-level compositions with injectable `platform`/`env` for testing

### Provider resolvers (`src/lib/rollout.js`)
Every Windows provider now routes through mode-aware arbitration:
- **New resolvers**: KimiCode, CodeBuddy, WorkBuddy, OMP, Pi, Craft, Grok Build, Droid, Copilot OTel, KiloCode/RooCode
- **Existing**: Hermes, Zed, Goose — updated to use the shared `pickWin32Path` pattern
- `pickWin32ProviderPath` internal helper centralizes repeated native/WSL arbitration
- All resolvers return `null`/no-candidate in exclusive modes when the excluded side has no candidate

### Null safety
- `status` and `sync` commands guard against `existsSync(null)` with short-circuit
- `resolveHermesDbPath` returns `null` when resolver returns `null`
- `resolveAllHermesDBPaths` returns `{ default: null, profiles: {} }` when `hermesDir` is null
- All downstream file-listing/default-model consumers null-guarded

### Status reporting
- JSON output: top-level `wsl_mode` + `wsl_distros` fields
- Human-readable: shows active mode, invalid-value warning, and distro list
- `TOKENTRACKER_SKIP_WSL_PROBE` fully removed (use `TOKENTRACKER_WSL_MODE=native-only`)

### Tests
- `test/wsl-probe.test.js`: 17 unit tests (all 6 mode helpers, `normalizeWslMode`, `pickWin32Path` matrix, exclusive-mode skip) + 2 real-WSL integration (skip on non-Win32)
- `test/rollout-parser.test.js`: 5 resolver regression tests (Zed, Goose, Kimi Code ×2, Kilocode)
- `test/status.test.js`: `native-only` does not invoke `wsl.exe`
- Full suite: 1113 pass, 0 fail, 1 skip

## Breaking Changes
- `TOKENTRACKER_SKIP_WSL_PROBE` removed — use `TOKENTRACKER_WSL_MODE=native-only`

Closes #238

## Summary by CodeRabbit

* **New Features**
  * Enhanced `status` reporting with Windows/WSL-aware output, including `WSL mode` and (when available) discovered WSL distros in both human-readable and structured formats.
  * Updated Hermes status display to reflect actual installation more accurately.
* **Bug Fixes**
  * Avoided unnecessary filesystem existence checks when provider DB path values are missing.
* **Tests**
  * Added/expanded coverage for WSL probing, resolver behavior across WSL modes (including `native-only`), UNC detection, and SQLite snapshotting.
* **Chores**
  * Updated local ignore rules for new development directories/files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,7 @@ skills-lock.json
 .roo
 .trae
 .windsurf
+.opencode/
+.serena/
+openspec/
+plans/
--- a/src/commands/status.js
+++ b/src/commands/status.js
@@ -429,6 +429,7 @@ async function cmdStatus(argv = []) {
       ...(process.platform === "win32"
         ? {
             wsl_mode: getWslMode(),
+            wsl_mode_invalid: isInvalidWslMode(),
             wsl_distros: wslDistros.map((d) => ({ name: d.name, version: d.version })),
           }
         : {}),
--- a/src/commands/status.js
+++ b/src/commands/status.js
@@ -62,6 +62,7 @@ const {
   resolveHermesDbPath,
   probeWslDistros,
 } = require("../lib/rollout");
+const { getWslMode, isInvalidWslMode, shouldProbeWsl } = require("../lib/wsl-probe");
 const { probeGrokHookState, resolveGrokHome } = require("../lib/grok-hook");
 
 async function cmdStatus(argv = []) {
@@ -268,7 +269,7 @@ async function cmdStatus(argv = []) {
   // per thread we've surfaced usage for, so its size distinguishes "DB present
   // with usage" from "DB present but nothing counted yet" (fresh/undecodable).
   const zedDbPath = resolveZedDbPath(process.env);
-  const zedInstalled = fssync.existsSync(zedDbPath);
+  const zedInstalled = Boolean(zedDbPath && fssync.existsSync(zedDbPath));
   const zedThreadsCounted =
     cursors?.zed?.threadTotals && typeof cursors.zed.threadTotals === "object"
       ? Object.keys(cursors.zed.threadTotals).length
@@ -276,7 +277,7 @@ async function cmdStatus(argv = []) {
 
   // Goose (Block) — passive cumulative-delta read of sessions.db.
   const gooseDbPath = resolveGooseDbPath(process.env);
-  const gooseInstalled = fssync.existsSync(gooseDbPath);
+  const gooseInstalled = Boolean(gooseDbPath && fssync.existsSync(gooseDbPath));
 
   // Droid (Factory CLI) — passive cumulative-delta read of *.settings.json.
   const droidSessionsDir = resolveDroidSessionsDir(process.env);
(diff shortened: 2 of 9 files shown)


mm7894215:
Merged — thanks, this is a clean piece of work. Reviewed the whole diff and ran the full suite locally (1153 pass): the win32-gating keeps macOS/Linux untouched, the `wsl.exe` probes are all fail-safe inside try/catch, and the null-safety guards downstream in `status`/`sync` are complete. Nice catch fixing the `parseWslListVerbose` whitespace-stripping bug along the way. Closes #238.

PR #251, opened 2026-07-02
ci: install sqlite3 for test fixtures

xMinhx:
## Summary

Installs `sqlite3` CLI on the Ubuntu CI runner as an explicit dependency. The test suite uses `sqlite3` to create fixture databases for Zed, Goose, Kiro CLI, and Go migration parsers.

## Background

GitHub-hosted Ubuntu runners include `sqlite3` preinstalled today, so this is **not** fixing a current CI failure. The purpose is defensive dependency hygiene — guarding against a future runner image dropping `sqlite3`, and documenting that this is a test dependency.

Uses `--no-install-recommends` to minimize image bloat.

## Impact

- Adds ~2s per CI run (apt-get update + install)
- Ensures test suite doesn't silently break if the runner image changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,11 @@ jobs:
             package-lock.json
             dashboard/package-lock.json
 
+      - name: Install sqlite3 (defensive — preinstalled on GH runners but explicit dep hygiene)
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -y sqlite3
+
       - name: Install root dependencies
         run: npm ci
 


mm7894215:
Thanks for this! Digging into the premise though — the claim that 49 tests fail with `sqlite3 ENOENT` on every CI run doesn't match what I see: the last several `main` pushes **and** PR #249's CI all pass `test + validate + build` on `ubuntu-latest` **without** installing sqlite3. The GitHub-hosted Ubuntu runner ships sqlite3 preinstalled, so there's no 49-test failure to fix right now.

The change is harmless, but it adds `apt-get update && install` to every run (a network dependency + some time). The one solid reason to keep it is defensive — guarding against a future runner image dropping sqlite3 (explicit-dependency hygiene). If that's the intent, I'm happy to merge, but let's update the description to match (and maybe `--no-install-recommends`). If you actually hit `ENOENT` in a specific environment, can you share where? Otherwise I'd lean toward not adding the extra CI dependency.

xMinhx:
My mistake — the audit tool reported 49 sqlite3 ENOENT failures, but I verified the wrong thing. GitHub Ubuntu runners already have sqlite3 preinstalled; those 49 failures are specific to my local environment. Sorry for the noise, closing this.

xMinhx:
Thanks for the thorough review — you're absolutely right. I audited the test suite locally (where `sqlite3` wasn't installed) and wrongly assumed the CI runner would have the same 49 failures. Should have checked the actual CI runs on `main` first — my mistake.

Updated the description to match reality: this is purely defensive dependency hygiene, not a bug fix. Added `--no-install-recommends` per your suggestion.

Happy to close if you'd rather not add the overhead — no hard feelings either way.

mm7894215:
Thanks for the PR. Since sqlite3 is preinstalled on GitHub-hosted runners and this isn't fixing a current failure, we'd rather keep CI lean and add the explicit install only if a runner image ever drops it. Closing for now.

PR #253, opened 2026-07-03
fix: harden macOS dashboard asset loading against stale chunk requests

paveezheng:
## Summary

Prevent the macOS app dashboard from failing with `text/html is not a valid JavaScript MIME type` when WKWebView requests a stale or missing Vite chunk, by returning a real 404 for missing static assets and forcing the embedded/system CLI server to bind the fixed macOS app port.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Checklist

- [x] `npm test` passes

---

<details>

### Risk layer triggers

### Rules / invariants

- Missing static assets under the local CLI server must not fall back to `index.html`.
- SPA fallback should apply only to real page navigations, not `assets/*.js`, CSS, or API-like paths.
- The macOS app must launch and load the same fixed local port (`:7680`) because Swift-side URLs are hardcoded to that port.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server startup to consistently use the configured port in both embedded and fallback launch paths.
  * Made app routing more precise so non-page requests return a 404 instead of loading the main app shell.
* **Tests**
  * Added coverage for server port handling on macOS.
  * Added coverage for SPA fallback behavior across page, API, and static asset requests.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Services/ServerManager.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Services/ServerManager.swift
@@ -177,7 +177,7 @@ final class ServerManager: ObservableObject {
     private func launchServer(nodePath: String, entryPath: String) {
         let process = Process()
         process.executableURL = URL(fileURLWithPath: nodePath)
-        process.arguments = [entryPath, "serve", "--no-sync", "--no-open"]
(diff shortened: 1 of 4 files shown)


mm7894215:
Merged, thanks! Verified the port-drift scenario against ServerManager.swift — the explicit --port plus the 404-for-missing-assets change closes the stale-chunk MIME failure path nicely. This will ship with the next release.

PR #256, opened 2026-07-03
Improve native menu bar catch-up after wake

lamther:
## Summary
- add wake and session-active observers for the macOS menu bar app
- debounce wake/session catch-up attempts for 60 seconds
- run a real sync only when the last successful background sync is stale; otherwise only reload local data

## Why
When macOS sleeps or locks, scheduled refresh work can be delayed. This keeps the existing 5-minute UI refresh and 30-minute regular sync cadence unchanged, while making the app catch up promptly after wake/unlock without repeatedly syncing when data is already fresh.

## Verification
- xcodegen generate
- xcodebuild test -project TokenTrackerBar/TokenTrackerBar.xcodeproj -scheme TokenTrackerBarTests -destination 'platform=macOS' -derivedDataPath /tmp/TokenTrackerBarDerivedData
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added automatic catch-up behavior when the app becomes active or the Mac wakes from sleep.
  * Introduced smarter sync timing so the app can decide whether to refresh data or just reload what’s already available.

* **Bug Fixes**
  * Improved handling of repeated wake/active events to avoid triggering duplicate refreshes.
  * Ensured catch-up sync logic behaves consistently when no prior sync exists or when data is already fresh.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Models/BackgroundRefreshPolicy.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Models/BackgroundRefreshPolicy.swift
@@ -3,6 +3,7 @@ import Foundation
 enum BackgroundRefreshPolicy {
     static let defaultRefreshInterval: TimeInterval = 300
     static let defaultSyncInterval: TimeInterval = 1_800
+    static let defaultCatchUpStaleInterval: TimeInterval = 300
 
     static func shouldRunSync(
         now: Date,
@@ -12,5 +13,15 @@ enum BackgroundRefreshPolicy {
         guard syncInterval > 0 else { return false }
         guard let lastSyncAt else { return true }
         return now.timeIntervalSince(lastSyncAt) >= syncInterval
+    }
+
+    static func shouldRunCatchUpSync(
+        now: Date,
+        lastSyncAt: Date?,
+        staleInterval: TimeInterval = defaultCatchUpStaleInterval
+    ) -> Bool {
+        guard staleInterval > 0 else { return false }
+        guard let lastSyncAt else { return true }
+        return now.timeIntervalSince(lastSyncAt) >= staleInterval
     }
 }
--- a/TokenTrackerBar/TokenTrackerBar/TokenTrackerBarApp.swift
+++ b/TokenTrackerBar/TokenTrackerBar/TokenTrackerBarApp.swift
@@ -13,11 +13,13 @@ struct TokenTrackerBarApp: App {
 final class AppDelegate: NSObject, NSApplicationDelegate {
 
     private var statusBarController: StatusBarController?
+    private var lastWakeCatchUpAttemptAt: Date?
     private let viewModel = DashboardViewModel()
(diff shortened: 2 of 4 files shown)


lamther:
CI note: the failing `CI / test + validate + build` job is caused by an existing upstream `main` guardrail failure, not by this PR's changes.

The failing step is `npm run validate:ui-hardcode`, with:

```text
UI hardcode guardrails failed:
- dashboard/src/ui/dashboard/components/ActivityHeatmap.jsx: colors increased (28 -> 29)
```

I verified the same command against current `origin/main` (`8dc9e8d`) in a clean local worktree, and it fails with the same `ActivityHeatmap.jsx` guardrail error. This PR only changes the native wake catch-up path and its Swift policy tests; the failed dashboard hardcode baseline is unrelated to this diff.

lamther:
CI note: the failing `CI / test + validate + build` job is caused by an existing upstream `main` guardrail failure, not by this PR's changes.

The failing step is `npm run validate:ui-hardcode`, with:

```text
UI hardcode guardrails failed:
- dashboard/src/ui/dashboard/components/ActivityHeatmap.jsx: colors increased (28 -> 29)
```

I verified the same command against current `origin/main` (`8dc9e8d`) in a clean local worktree, and it fails with the same `ActivityHeatmap.jsx` guardrail error. This PR only changes the native wake catch-up path and its Swift policy tests; the failed dashboard hardcode baseline is unrelated to this diff.

mm7894215:
Reviewed — clean and correct: the wake/session-active double-fire is collapsed by the synchronous debounce stamp, catch-up reuses the existing lastBackgroundSyncAt state so it complements rather than fights the 5-min/30-min cadence, and status-item updates go through the existing guarded path only. The red `test + validate + build` check is inherited from a main-side guardrail break (fixed on main in 08f7e56, unrelated to this Swift-only PR) — recycling the PR below to re-run CI against current main. Will merge once green.

mm7894215:
Merged, thanks! Ships with the next release.

mm7894215:
@Arkycher Heads up: this PR introduced a regression that caused the menu bar app to silently quit after sleep/wake (reported in #286 / #290).

**What went wrong**
`scheduleWakeCatchUp` calls `ensureServerRunning()` on wake. That path ran `killExistingServerOnPort()` with:

```bash
lsof -ti tcp:<port>
```

That lists *all* processes touching the port — including this app’s own URLSession client connections — so the app could `SIGTERM` itself. No crash dialog; the menu bar icon just disappeared.

**Fix**
Shipped in **v0.74.3** ([2869037](https://github.com/mm7894215/TokenTracker/commit/2869037)):
- only kill `LISTEN` sockets (`lsof -sTCP:LISTEN`)
- never kill our own PID
- no-op when the embedded server is already healthy

The wake catch-up idea is still good — the bug was in the port-cleanup helper it invoked. Thanks for the contribution; please keep this failure mode in mind for similar ensure/restart paths.

PR #257, opened 2026-07-03
Mint relayed cloud tokens for background sync

lamther:
## Summary
- allow non-drain local sync requests to best-effort mint a temporary cloud device token from the relayed login session
- keep explicit device tokens as the highest-precedence path
- preserve strict drain behavior: drain sync still returns 502 and does not spawn sync if token minting fails
- add success, fallback, and no-fetch guard tests for the background mint path

## Why
Background auto sync can run without a provided device token even when the dashboard already has a relayed cloud login. Best-effort token minting lets eligible background sync upload normally, while avoiding extra network work when cloud sync is off, no refresh token exists, or a device token was explicitly supplied.

## Performance notes
The new mint attempt is gated by all of these conditions: no explicit device token, cloud sync enabled, and relayed refresh token present. There are no retry loops; non-drain failures fall back to local sync.

## Verification
- node --test test/local-api-security.test.js
- npm test
- git diff --check
- codex-dynamic-workflow review: .codex-workflow/runs/20260703-175807/report.md (Final vote 3/3)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved non-drain local sync when cloud sync is enabled: a device token is now automatically minted and applied if it’s missing.
  * If token minting/issuance fails, non-drain local sync continues without a device token (only drain behavior remains erroring).
  * Non-drain local sync now reuses a cached relayed device token and correctly handles refresh rotation and CSRF token persistence.
* **Tests**
  * Expanded local API security tests with isolated relayed-login fixtures, covering device-token minting, caching, reuse across repeated requests, explicit token overrides, and failure/fallback scenarios.

### DIFF
--- a/src/lib/cloud-account.js
+++ b/src/lib/cloud-account.js
@@ -68,12 +68,13 @@ function decodeJwtExpMs(token) {
   return 0;
 }
 
-// Module-level access-token cache, keyed by the refresh token that produced it.
+// Module-level access-token cache, keyed by the cloud base URL and refresh token
+// that produced it.
 // The popover polls frequently, so caching avoids hammering /api/auth/refresh.
-let tokenCache = { refreshToken: null, accessToken: null, expMs: 0 };
+let tokenCache = { cacheKey: null, accessToken: null, expMs: 0 };
 
 function __resetCloudAccountCacheForTests() {
-  tokenCache = { refreshToken: null, accessToken: null, expMs: 0 };
+  tokenCache = { cacheKey: null, accessToken: null, expMs: 0 };
 }
 
 function csrfTokenFromRefreshPayload(data) {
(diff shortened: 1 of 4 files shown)


lamther:
CI note: the failing `CI / test + validate + build` job is caused by an existing upstream `main` guardrail failure, not by this PR's changes.

The failing step is `npm run validate:ui-hardcode`, with:

```text
UI hardcode guardrails failed:
- dashboard/src/ui/dashboard/components/ActivityHeatmap.jsx: colors increased (28 -> 29)
```

I verified the same command against current `origin/main` (`8dc9e8d`) in a clean local worktree, and it fails with the same `ActivityHeatmap.jsx` guardrail error. This PR only changes the local sync cloud-token path and local-api security tests; the failed dashboard hardcode baseline is unrelated to this diff.

mm7894215:
Reviewed against the incident history in this area — the dangerous parts are handled correctly:

- Rotated refresh/csrf tokens are persisted back to the relay store (`setRelayRefreshToken`/`setRelayCsrfToken`), so frequent minting can't strand the dashboard session.
- The issue endpoint no longer rotates-on-issue (that's the exact mechanism behind our historical 65%-upload-stall incident), so re-minting can't kill the CLI's long-lived token.
- Device identity is machine_id-anchored, so the env-token override attributes uploads to the same device row. Test coverage of the gates (explicit token, cloud-sync off, no refresh token, mint failure fallback) is thorough.

One change requested before merge: **cache the minted device token in-process.** Neither caller passes an explicit deviceToken (web `triggerLocalSync` posts no body; macOS `triggerSync()` posts `{}`), so after this PR every non-drain sync — manual refreshes, menu-bar Sync Now, and the 30-min background cadence — hits `tokentracker-device-token-issue` and inserts a fresh token row. Those rows have no `expires_at` and are only cleaned by explicit revocation, so this accumulates dozens of live bearer tokens per user per day. `mintAccessToken` already has a module-level `tokenCache` — do the same for the issued device token (key it on the refresh token + machineId, reuse until either changes). That bounds issuance to ~once per server process instead of once per sync.

Cosmetic: `issueDeviceTokenForDrainSync` is no longer drain-only — worth renaming or updating its comment in the same pass.

PR #260, opened 2026-07-03
feat: aggregate token usage from both native Windows and WSL installations

### LINKED ISSUE #250: feat: aggregate token usage from both native Windows and WSL installations
## Problem

 returns exactly one path per provider based on . If a user runs the same tool (e.g., Codex, Hermes, Kimi Code) in both native Windows **and** WSL simultaneously, tokens from the non-selected installation are silently dropped.

## Example

A developer uses Hermes Agent in both Windows () and WSL (). With  (default), only the WSL state.db is synced — the native Windows sessions are never counted.

## Proposed solution

Scan **both** installations and merge the results with deduplication:

1. Resolve both native and WSL paths (regardless of mode, or opt-in via new mode like )
2. Parse each installation independently (separate cursors, separate DB connections)
3. Merge aggregated token counts per half-hour bucket (deduplicate by task UUID / session ID)
4. Report the dual-source status in Status:
- Base URL: https://srctyff5.us-east.insforge.app
- Device token: unset
- Queue: 132612 bytes pending
- Last parse: 2026-06-30T21:05:14.732Z
- Last notify: 2026-06-28T15:56:17.247Z
- Last OpenClaw-triggered sync: never
- Last notify-triggered sync: never
- Last upload: never
- Next upload after: never
- Backoff until: never
- Codex notify: ["/usr/bin/env","node","/home/minh/.tokentracker/bin/notify.cjs"]
- Every Code notify: unset
- Claude hooks: unset
- Gemini hooks: unset
- Opencode plugin: set
- OpenClaw session plugin: unset
- OpenClaw session plugin conversation access: unset
- OpenClaw hook (legacy): unset
- Passive mode: 1 provider reading logs without hooks (gemini: writable (hook may have been removed externally))
- Codex ChatGPT plan: plus
- OpenCode ChatGPT plan: plus

## Risks

- Two SQLite DBs for Hermes/Zed/Goose → double the parse time on Windows
- Deduplication logic needed — same task running in WSL vs Windows would have different session IDs, so this is mostly about *complementary* data rather than duplicates
- Cursor management — each path needs its own last-completed cursor

## Alternatives considered

- New mode value `both` (separate from existing `wsl-first`/`native-first`)
- Per-provider opt-in: only aggregate for providers known to have independent DBs per environment

---

xMinhx:
Closes #250

Adds `TOKENTRACKER_WSL_MODE=both` — scan both native Windows and WSL installs instead of picking one. Merges their token data into the same 30-min buckets. Optional `TOKENTRACKER_WSL_PREFER` if you want one side to win on overlap.

Works for all 11 WSL-aware providers (Hermes, Zed, Goose, Kimi Code, CodeBuddy, WorkBuddy, OMP, Pi, Craft, Droid, Kilo Code).

## What changed

- New `src/lib/install-resolver.js` — centralized `resolveInstallPaths()` returns `{ native, wsl }` per provider. Every provider now calls this instead of duplicating the native-vs-WSL logic.
- New `src/lib/multi-install-parser.js` — `multiInstallParse()` runs the parser once per install path with separate cursor state. Cursors are namespaced (`cursors.hermes.native` / `cursors.hermes.wsl`) so each path tracks its own position. `mergeBothFileSources()` does the same for file-based providers (Kimi Code, CodeBuddy, etc.) by temporarily toggling the env mode.
- `src/lib/wsl-probe.js` — added `both` mode and `getWslPrefer()`.
- `src/commands/sync.js` — all providers wired. SQLite-based (Hermes, Zed, Goose) use `multiInstallParse`. File-based use `mergeBothFileSources`.
- `src/commands/status.js` — shows `wsl_prefer` when mode is `both`.
- `src/lib/rollout.js` — `resolveHermesPath()`, `resolveZedDbPath()`, `resolveGooseDbPath()` refactored to use the centralized resolver. `pickWin32ProviderPath()` delegates to it.
- 14 commits, ~1,300 lines added.

## How to test

```bash
TOKENTRACKER_WSL_MODE=both tokentracker sync
TOKENTRACKER_WSL_MODE=both TOKENTRACKER_WSL_PREFER=wsl tokentracker sync
```

## Tested

- 264 unit tests, 247 pass (17 pre-existing `sqlite3 ENOENT`)
- Self-contained Windows test: creates temp SQLite DBs, runs resolver + parser pipeline, validates queue output
- Real Windows 11: resolver discovers both `C:\Users\...\Local\hermes` and `\\wsl$\Ubuntu-26.04\home\minh\.hermes`, parser aggregates 3 sessions (600 input, 300 output) correctly
- CodeRabbit: 4 comments + 5 nitpicks, all addressed

Only CI failure is a pre-existing dashboard color guardrail (`ActivityHeatmap.jsx: colors 28 -> 29`). Not our code.

## Summary by CodeRabbit

* **New Features**
  * Added support for Windows environments that have both native and WSL installs, letting the app discover and combine data from multiple install locations.
  * Improved parsing coverage for several providers so events, sessions, and settings can be collected from either install source.

* **Bug Fixes**
  * Fixed path detection issues when an install directory is missing or undefined.
  * Improved cursor handling so data from different install types is merged more reliably without losing progress or totals.

### DIFF
--- a/src/commands/status.js
+++ b/src/commands/status.js
@@ -62,7 +62,7 @@ const {
   resolveHermesDbPath,
   probeWslDistros,
 } = require("../lib/rollout");
-const { getWslMode, isInvalidWslMode, shouldProbeWsl } = require("../lib/wsl-probe");
+const { getWslMode, getWslPrefer, isInvalidWslMode, shouldProbeWsl } = require("../lib/wsl-probe");
 const { probeGrokHookState, resolveGrokHome } = require("../lib/grok-hook");
 
 async function cmdStatus(argv = []) {
@@ -223,14 +223,14 @@ async function cmdStatus(argv = []) {
 
   // oh-my-pi — passive scan only (no hooks).
   const ompAgentDir = resolveOmpAgentDir(process.env);
-  const ompInstalled = fssync.existsSync(path.join(ompAgentDir, "sessions"));
+  const ompInstalled = Boolean(ompAgentDir) && fssync.existsSync(path.join(ompAgentDir, "sessions"));
   const ompFiles = ompInstalled ? resolveOmpSessionFiles(process.env) : [];
 
   // pi (@mariozechner/pi-coding-agent) — passive scan only (no hooks).
   // Skip when its agent dir collides with omp's; sync would dedupe anyway.
   const piCollides = piAgentDirCollidesWithOmp(process.env);
   const piAgentDir = resolvePiAgentDir(process.env);
-  const piInstalled = !piCollides && fssync.existsSync(path.join(piAgentDir, "sessions"));
+  const piInstalled = !piCollides && Boolean(piAgentDir) && fssync.existsSync(path.join(piAgentDir, "sessions"));
   const piFiles = piInstalled ? resolvePiSessionFiles(process.env) : [];
 
   // Craft Agents — passive scan only (no hooks).
@@ -429,6 +429,7 @@ async function cmdStatus(argv = []) {
       ...(process.platform === "win32"
         ? {
             wsl_mode: getWslMode(),
+            wsl_prefer: getWslMode() === "both" ? getWslPrefer() : undefined,
             wsl_mode_invalid: isInvalidWslMode(),
             wsl_distros: wslDistros.map((d) => ({ name: d.name, version: d.version })),
           }
@@ -539,6 +540,9 @@ async function cmdStatus(argv = []) {
         const lines = [
           `- WSL mode: ${wslMode}${modeSuffix}`,
         ];
+        if (wslMode === "both") {
+          lines.push(`  prefer: ${getWslPrefer()}`);
+        }
         if (wslDistros.length > 0) {
           lines.push(`  distros: ${wslDistros.map((d) => `${d.name} (v${d.version ?? "?"})`).join(", ")}`);
         }
--- a/src/commands/status.js
+++ b/src/commands/status.js
@@ -62,7 +62,7 @@ const {
   resolveHermesDbPath,
   probeWslDistros,
 } = require("../lib/rollout");
-const { getWslMode, getWslPrefer, isInvalidWslMode, shouldProbeWsl } = require("../lib/wsl-probe");
+const { getWslMode, isInvalidWslMode, shouldProbeWsl } = require("../lib/wsl-probe");
 const { probeGrokHookState, resolveGrokHome } = require("../lib/grok-hook");
 
 async function cmdStatus(argv = []) {
@@ -429,7 +429,6 @@ async function cmdStatus(argv = []) {
       ...(process.platform === "win32"
         ? {
             wsl_mode: getWslMode(),
-            wsl_prefer: getWslMode() === "both" ? getWslPrefer() : undefined,
             wsl_mode_invalid: isInvalidWslMode(),
             wsl_distros: wslDistros.map((d) => ({ name: d.name, version: d.version })),
           }
@@ -540,9 +539,6 @@ async function cmdStatus(argv = []) {
         const lines = [
           `- WSL mode: ${wslMode}${modeSuffix}`,
         ];
-        if (wslMode === "both") {
-          lines.push(`  prefer: ${getWslPrefer()}`);
-        }
         if (wslDistros.length > 0) {
           lines.push(`  distros: ${wslDistros.map((d) => `${d.name} (v${d.version ?? "?"})`).join(", ")}`);
         }
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -741,7 +741,7 @@ async function cmdSync(argv) {
         const nativeDb = path.join(appData, "goose", "sessions", "sessions.db");
         const wslDir = wsl.shouldProbeWsl(process.env) ? wsl.discoverWslHome(".local/share/goose/sessions") : null;
         const wslDb = wslDir ? path.join(wslDir, "sessions.db") : null;
-        const goosePaths = resolveInstallPaths("goose", { nativeValue: nativeDb, wslValue: wslDb });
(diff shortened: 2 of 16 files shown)


mm7894215:
The failing `test + validate + build` was our fault — a comment in main's ActivityHeatmap.jsx tripped the ui-hardcode guard (`#252` parsed as a hex color), red since v0.65.2. Fixed on main now; reopening cycle below to re-run CI against fresh main. Review coming after that.

mm7894215:
Reviewed in depth (fresh CI is green after the main fix). The core design is sound: default-mode behavior is byte-identical for existing users, `mergeBothFileSources` passes env copies rather than mutating `process.env`, WSL discovery returning only UNC paths keeps native/WSL from aliasing the same physical files, error isolation preserves the first install's committed progress when the second throws, and the probe caching means no per-provider wsl.exe storm. Test coverage is genuinely good.

Two things to fix before merge:

**1. First-time `both` enable silently truncates the newly-added install's backfill (medium).** `ensureNamespacedCursors` (install-resolver.js:27-39) seeds BOTH namespaces with a deep copy of the existing flat cursor. For hermes/zed that cursor is a forward-only watermark (`lastCompletedStartedAt` → `WHERE started_at >= since`, rollout.js:3898; `lastUpdatedAt` → `WHERE updated_at > cursor`). The install that was never parsed under single mode inherits the other install's watermark and permanently skips every completed session older than it — exactly the historical data this feature exists to surface. cursor-migration.test.js:108-111 currently asserts this copy-to-both as intended, and its mock parserFn can't catch the truncation. Suggested fix: seed only the namespace the pre-`both` mode was actually tracking with the flat state, and start the other namespace empty so it backfills fully; add a cumulative-delta test that puts pre-watermark history in the newly-added install and asserts it lands in the queue. (Goose is unaffected — full-table scan + per-session dedup.)

**2. `TOKENTRACKER_WSL_PREFER` doesn't actually discard anything.** The prefer branch (multi-install-parser.js:71-91) restores the preferred install's hourly buckets after the second parse, but each `parse*Incremental` has already appended its rows to the queue inline (`enqueueTouchedBuckets`) — readers take the latest entry per (source, model, hour), which is the summed row emitted before the restore. So prefer mode produces the same totals as plain `both` while logging a misleading "discarded contributions" warning. Either dedup at emit time, or drop the knob and document `both` as always-sum (my preference — simpler, and overlap within the same provider+model+hour across two installs is rare).

Nits, non-blocking: `resolveInstallPaths`'s `providerName` param is unused (install-resolver.js:4); prefer-mode `bucketsQueued` sums stats for buckets it claims to discard.

Happy to re-review quickly once the migration seeding is fixed — that's the one real data-correctness gate.

xMinhx:
> Reviewed in depth (fresh CI is green after the main fix). The core design is sound: default-mode behavior is byte-identical for existing users, `mergeBothFileSources` passes env copies rather than mutating `process.env`, WSL discovery returning only UNC paths keeps native/WSL from aliasing the same physical files, error isolation preserves the first install's committed progress when the second throws, and the probe caching means no per-provider wsl.exe storm. Test coverage is genuinely good.
> 
> Two things to fix before merge:
> 
> **1. First-time `both` enable silently truncates the newly-added install's backfill (medium).** `ensureNamespacedCursors` (install-resolver.js:27-39) seeds BOTH namespaces with a deep copy of the existing flat cursor. For hermes/zed that cursor is a forward-only watermark (`lastCompletedStartedAt` → `WHERE started_at >= since`, rollout.js:3898; `lastUpdatedAt` → `WHERE updated_at > cursor`). The install that was never parsed under single mode inherits the other install's watermark and permanently skips every completed session older than it — exactly the historical data this feature exists to surface. cursor-migration.test.js:108-111 currently asserts this copy-to-both as intended, and its mock parserFn can't catch the truncation. Suggested fix: seed only the namespace the pre-`both` mode was actually tracking with the flat state, and start the other namespace empty so it backfills fully; add a cumulative-delta test that puts pre-watermark history in the newly-added install and asserts it lands in the queue. (Goose is unaffected — full-table scan + per-session dedup.)
> 
> **2. `TOKENTRACKER_WSL_PREFER` doesn't actually discard anything.** The prefer branch (multi-install-parser.js:71-91) restores the preferred install's hourly buckets after the second parse, but each `parse*Incremental` has already appended its rows to the queue inline (`enqueueTouchedBuckets`) — readers take the latest entry per (source, model, hour), which is the summed row emitted before the restore. So prefer mode produces the same totals as plain `both` while logging a misleading "discarded contributions" warning. Either dedup at emit time, or drop the knob and document `both` as always-sum (my preference — simpler, and overlap within the same provider+model+hour across two installs is rare).
> 
> Nits, non-blocking: `resolveInstallPaths`'s `providerName` param is unused (install-resolver.js:4); prefer-mode `bucketsQueued` sums stats for buckets it claims to discard.
> 
> Happy to re-review quickly once the migration seeding is fixed — that's the one real data-correctness gate.

Thanks for the thorough deep dive and the green light on the core design! Really appreciate the detailed breakdown on the cursor edge case, that's a great catch! I'll fix it.

mm7894215:
Merged — thanks for the thorough work across both review rounds, this is a solid feature.

One residual edge I took over and fixed on top (`1d9e2c8`) rather than asking for a third round: the flat→namespaced cursor migration always treated the WSL install as the flat cursor's owner (`legacyKey`'s native branch was unreachable — the dual-parse path only runs in `both` mode). For a user whose flat cursor was actually tracking the native install, that wiped the native namespace's dedup snapshots, and hermes/zed/goose would re-add that install's entire history to the hourly buckets on the next sync.

The migration now asks the DBs instead of guessing: `multiInstallParse` accepts a `detectInstall` probe, and each provider checks which install's DB contains the flat cursor's own session/thread ids. Exactly one match → the other namespace starts empty and backfills fully (your intended behavior). Anything else — no match, both match, probe failure — seeds every namespace, trading a bounded backfill gap for a guaranteed no-double-count.

Ships in v0.68.0.

PR #268, opened 2026-07-05
fix: harden Linux browser launcher fallbacks

Jacobinwwey:
- **Issue:** n/a
- **Problem:** Linux browser launch currently assumes `xdg-open` exists and should always be invoked. In headless sessions that produces a blind spawn attempt, and in WSL or minimal desktop environments it misses the launcher the user actually has.
- **Trade-offs:** This keeps the change inside the existing CLI/browser-launch path instead of introducing a Linux desktop host. The chosen approach adds small PATH probing and environment checks to reduce false launches; the cost is a few extra filesystem lookups before spawning, which is negligible relative to a CLI startup.

- Added a Linux/WSL launch resolver in `src/lib/browser-auth.js` that picks `wslview` for WSL, skips browser spawning entirely when no GUI session is present, and falls back through `xdg-open`, `gio open`, and `sensible-browser`.
- Kept the macOS and Windows branches behaviorally intact while making `openInBrowser()` return a boolean so callers/tests can distinguish a real launch from a no-op.
- Added a regression test file for WSL detection, headless Linux behavior, launcher precedence, and detached spawn semantics.
- **Note:** This PR deliberately does not touch Linux release packaging. Artifact production, `prepack` hygiene, and desktop integration are follow-up work so the diff stays atomic.

- [x] Added Unit Tests for `src/lib/browser-auth.js` (covering WSL launcher selection, headless Linux no-op, and `gio` fallback)
- [x] Ran Integration Test Suite against local Ubuntu runtime via direct module invocation and targeted Node test execution
- [x] Verified backward compatibility with existing macOS/Windows command selection paths by leaving their branch logic intact and unmodified in tests
- **Benchmark:** Not applicable. The new path adds bounded PATH probing before spawn; no hot-path or algorithmic regression is expected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser opening reliability across macOS, Linux, and WSL.
  * Browser links now use the best available system launcher when a graphical session is present.
  * In headless environments, the app now avoids launching a browser when no suitable command is available.
  * macOS browser launches now fall back more gracefully if the primary approach fails.

### DIFF
--- a/src/lib/browser-auth.js
+++ b/src/lib/browser-auth.js
@@ -1,6 +1,8 @@
 const http = require("node:http");
 const crypto = require("node:crypto");
 const cp = require("node:child_process");
+const fs = require("node:fs");
+const path = require("node:path");
 
 const { DEFAULT_BASE_URL } = require("./runtime-config");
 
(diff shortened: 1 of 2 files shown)


mm7894215:
Merged, thanks. One note: the description says the macOS/Windows branches were kept behaviorally intact, but the diff switches Windows from `cmd /c start` to PowerShell `Start-Process`. That's arguably an improvement (`start` truncates URLs at `&`), but please keep the description in sync with the diff.

PR #269, opened 2026-07-05
fix: build dashboard before source-tree pack

Jacobinwwey:
- **Issue:** n/a
- **Problem:** `dashboard/dist/` is declared as a published file, but a source-tree `npm pack` path can still emit a tarball without it because `prepublishOnly` refreshes pricing data only. That leaves a silent release footgun: the package contract says the dashboard bundle ships, but local pack/publish does not enforce it.
- **Trade-offs:** This fix enforces the package contract at pack time instead of relying on workflow discipline. It slightly increases pack/publish work by rebuilding the dashboard before packaging, but it converts a latent broken-artifact risk into a deterministic build requirement.

- Added a `prepack` script in `package.json` that builds `dashboard/dist` and refreshes the pricing seed before `npm pack` / `npm publish` package assembly.
- Hardened `scripts/acceptance/npm-install-smoke.cjs` so the packed artifact must contain `dashboard/dist/index.html`, not just the CLI entrypoints.
- Added regression coverage in `test/npm-publish-workflow.test.js` for the new `prepack` contract and the strengthened smoke requirement.
- **Note:** This PR does not introduce Linux release artifacts yet. It only makes the existing npm distribution path honor the package's declared contents.

- [x] Added Unit Tests for npm publish/packaging guardrails (covering `prepack` contract and required `dashboard/dist/index.html` smoke validation)
- [x] Ran targeted Node test suite: `node --test test/npm-publish-workflow.test.js`
- [x] Verified backward compatibility with the current npm publish workflow contract: Node 20 publish path and `dashboard/dist` inclusion assertions remain intact
- **Benchmark:** Not applicable. The cost is one dashboard build before pack; no runtime hot path changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Packaging now runs required build steps before the package is created, ensuring the published tarball includes the latest dashboard and pricing content.
  * Release verification now fails if the dashboard build output (dashboard/dist/index.html) is missing from the npm package.

* **Tests**
  * Added test coverage to confirm the npm prepack script wiring and to validate expected tarball contents, including the dashboard build file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "graph:auto-index": "node scripts/graph/auto-index.cjs",
     "graph:scip": "node scripts/graph/generate-scip.cjs",
     "pricing:build-seed": "node scripts/build-pricing-seed.cjs",
+    "prepack": "npm run dashboard:build && npm run pricing:build-seed",
     "prepublishOnly": "node scripts/build-pricing-seed.cjs",
     "test": "node --test test/*.test.js",
     "validate:copy": "node scripts/validate-copy-registry.cjs",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "graph:scip": "node scripts/graph/generate-scip.cjs",
     "pricing:build-seed": "node scripts/build-pricing-seed.cjs",
     "prepack": "npm run dashboard:build && npm run pricing:build-seed",
-    "prepublishOnly": "node scripts/build-pricing-seed.cjs",
     "test": "node --test test/*.test.js",
     "validate:copy": "node scripts/validate-copy-registry.cjs",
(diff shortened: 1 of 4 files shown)


mm7894215:
Thanks for the writeup, but closing this one — the prepack rebuild would actually break our npm releases. The publish workflow builds the dashboard with VITE_INSFORGE_* injected at that build step only, then runs `npm publish`; a `prepack` that rebuilds the dashboard would run inside publish without those env vars and silently overwrite the good bundle (dead leaderboard fetch and cloud OAuth for every npm install). A verify-only prepack would avoid that, but the path it guards (source-tree `npm pack` outside CI) isn't part of our release flow, so we'd rather not add the moving part.

PR #270, opened 2026-07-05
feat: add Linux x64 release tarball workflow

Jacobinwwey:
- **Issue:** n/a
- **Problem:** The project currently ships native release artifacts for macOS and Windows, but Linux users have no first-party release output. The CLI works on Linux, yet there is no reproducible upstream workflow that assembles a self-contained Linux bundle and attaches it to a GitHub release.
- **Trade-offs:** This PR starts with a self-contained `linux-x64` tarball instead of a Linux tray host or distro-native packages. That keeps the artifact small in scope and reviewable, at the cost of deferring `arm64`, `.deb/.rpm`, and desktop-shell integration to follow-up work.

- Added `scripts/release/bundle-node-linux.sh` to assemble a self-contained `build/linux-x64` bundle with a pinned Node runtime, packaged tracker sources, production dependencies, dashboard assets, and a top-level `tokentracker` launcher.
- Added `.github/workflows/release-linux.yml` as a manual/reusable Ubuntu workflow that verifies the requested version, builds the dashboard, bundles the Linux runtime, verifies required files, packages `TokenTracker-linux-x64.tar.gz`, and uploads it to the target GitHub release.
- Added `test/release-linux-workflow.test.js` to lock the workflow contract and bundle-script invariants in CI.
- **Note:** This PR intentionally does not wire Linux into the existing macOS/Windows fan-out workflow yet. That integration is follow-up work once the standalone Linux artifact path lands.

- [x] Added Unit Tests for `release-linux.yml` and `bundle-node-linux.sh` contract checks (runner, step order, asset naming, required files, launcher generation)
- [x] Ran targeted test suite: `node --test test/release-linux-workflow.test.js`
- [x] Verified backward compatibility with the current release model by keeping the existing macOS/Windows workflows untouched
- **Benchmark:** Not applicable. This adds a release-time bundle path only; no runtime hot path changes.

## Summary by CodeRabbit

* **New Features**
  * Added a Linux release workflow that can be run manually or called from other workflows.
  * Packages the app into a standalone Linux x64 tarball and publishes it as a GitHub release asset.
  * Includes a self-contained Linux bundle with the app, dashboard build, and embedded Node runtime.

* **Bug Fixes**
  * Prevents overlapping release jobs for the same version.

### DIFF
--- a/.github/workflows/release-linux.yml
+++ b/.github/workflows/release-linux.yml
@@ -20,6 +20,7 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     permissions:
       contents: write
     steps:
@@ -32,10 +33,12 @@ jobs:
           node-version: 20
 
       - name: Verify version
+        env:
+          VERSION: ${{ inputs.version }}
         run: |
           PKG_VERSION=$(node -p "require('./package.json').version")
-          if [[ "$PKG_VERSION" != "${{ inputs.version }}" ]]; then
-            echo "::error::package.json version ($PKG_VERSION) does not match input (${{ inputs.version }})"
+          if [[ "$PKG_VERSION" != "$VERSION" ]]; then
+            echo "::error::package.json version ($PKG_VERSION) does not match input ($VERSION)"
             exit 1
           fi
 
--- a/.github/workflows/release-linux.yml
+++ b/.github/workflows/release-linux.yml
@@ -1,0 +1,84 @@
+name: release Linux
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to release (e.g. 0.68.0)"
+        required: true
+        type: string
+  workflow_call:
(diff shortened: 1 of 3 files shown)


mm7894215:
Appreciate the careful work here — the pinned runtime plus SHASUMS verification in the bundle script is nicely done. Closing though: Linux users are already covered by `npm i -g tokentracker-cli`, and a third release platform is a maintenance commitment we don't want to take on without demonstrated demand (there's no issue asking for this yet).

Two notes in case this gets revisited later:
- The tests shell out to python3/PyYAML, which isn't a project dependency — `npm test` breaks on machines without PyYAML installed.
- `gh release create` without `--draft` conflicts with the draft-first invariant in release-dmg.yml (releases/latest must never see a half-published release). A Linux job would need to join that workflow's fan-out and upload to the existing draft instead of creating its own release.

If real demand shows up we'll revisit — feel free to open an issue to track it.

PR #281, opened 2026-07-07
Fix Copilot auth.db schema v0 quota detection

EncodeTS:
## Summary

Fixes GitHub Copilot quota detection for installs whose `~/.config/github-copilot/auth.db` stores schema-v0 OAuth tokens.

On those installs, `oauth_tokens.token_ciphertext` is a plaintext GitHub OAuth token BLOB, not AES-GCM ciphertext. TokenTracker previously attempted the macOS keychain decrypt path for every auth.db row, so Copilot appeared disconnected even though Copilot CLI / compatible clients were authenticated.

## What changed

- Select `token_schema_version` from `oauth_tokens`.
- Decode schema-v0 `token_ciphertext` BLOBs as UTF-8 plaintext only when they match GitHub token shape.
- Preserve the existing AES-GCM + macOS keychain path for encrypted schemas.
- Keep token contents out of logs and test output.
- Add focused coverage for:
  - schema-v0 auth.db token decoding
  - encrypted/keychain auth.db decoding
  - mocked `fetchCopilotLimits` quota lookup with schema-v0 credentials

## Validation

- Original local status loop now reports `copilot.token_set=true`.
- `node --test test/copilot-auth-db.test.js test/usage-limits.test.js`
- `npm test`
- `npm run dashboard:build`
- `npm run validate:copy && npm run validate:ui-hardcode && npm run validate:guardrails`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Enhanced Copilot usage tracking to fetch limits more reliably across different stored sign-in token formats.
* **Bug Fixes**
  * Improved auth.db/OAuth token handling, including legacy plaintext and newer encrypted token variants.
  * Better token recovery behavior on macOS, preferring the correct GitHub.com token and avoiding unnecessary keychain access when not needed. 
* **Tests**
  * Expanded Copilot token and `fetchCopilotLimits` coverage for additional token-schema scenarios.

### DIFF
--- a/src/lib/usage-limits.js
+++ b/src/lib/usage-limits.js
@@ -6,6 +6,7 @@ const path = require("node:path");
 const http = require("node:http");
 const https = require("node:https");
 const { performance } = require("node:perf_hooks");
+const { promisify } = require("node:util");
 
 const {
   detectClaudeCodeSubscriptionDetails,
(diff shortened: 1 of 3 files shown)


mm7894215:
Merged. Cross-checked the schema-v0 assumption against other auth.db readers in the wild — v0 plaintext BLOB + AES-GCM for later schemas matches, and `token_schema_version` has been there since the table was introduced, so no regression risk for existing encrypted installs. Moving the limits path onto the async sqlite/keychain readers was a nice bonus.

Ships in v0.74.0.

PR #289, opened 2026-07-08
fix(windows):hide dashboard on minimise to clear desktop click-through ghost

wen-shion:
## Summary

Fix the Windows "invisible barrier" bug where minimising the dashboard left a transparent, click-eating "ghost" over the desktop. Every minimise path now routes through `Hide()` (SW_HIDE) instead of `WindowState=Minimized`, which fails to tear down the DWM extended frame on this acrylic window.

## Root cause

`DashboardWindow` is a transparent frosted-glass window: `WindowStyle.None` + `WindowChrome.GlassFrameThickness = -1` (sheet of glass) + `DwmExtendFrameIntoClientArea(-1)` + Acrylic backdrop, hosting a `WebView2CompositionControl` (DirectComposition visual).

When minimised via `WindowState = Minimized`, the DWM extended frame / DirectComposition visual is **not torn down** — it keeps capturing mouse input at the window's pre-minimise rectangle (92% of the work area, centred). Because acrylic is see-through, the user sees the desktop but can't click anything inside that rectangle: desktop icons stop highlighting until the window is hidden or `Win+D` is pressed. Only the sliver above the window's top edge (where the first icon row sits) stays clickable.

This affects **every** minimise path, not just the minimise button: the injected `win:min` button, clicking the taskbar button, `Win+D` (Show Desktop), taskbar right-click → Minimize, and any programmatic `SW_MINIMIZE`.

`Hide()` (SW_HIDE) — the path the close button and `Esc` already use — fully removes the HWND from hit-testing and demonstrably clears the ghost.

## Reproduce

Before the fix: run the Windows app from a debug build, open the dashboard from the tray, then minimise it (minimise button / taskbar click / `Win+D`). Desktop icons inside the window's former rectangle (92% of the work area, centred) stop responding — only the top row, above the window's centred top edge, stays clickable — until the window is hidden or `Win+D` is pressed again.

## Fix

In `TokenTrackerWin/DashboardWindow.cs`, route all minimise paths through `Hide()`:

- **`win:min` button**: `Hide()` instead of `WindowState = WindowState.Minimized`.
- **`StateChanged`**: when `WindowState == Minimized`, call `Hide()`. This catches every shell-driven minimise (taskbar click, `Win+D`, right-click → Minimize, programmatic `SW_MINIMIZE`).

Reopening is unchanged: the tray double-click calls `ShowDashboard()`, which `Show()`s the window and restores `WindowState` from `Minimized` to `Normal`.

## Behavioural change / trade-off

Minimise no longer minimises-to-taskbar; it tucks the dashboard to the tray (same as close / `Esc`), reopened via the tray. This is a **pragmatic, symptom-layer fix**: the root cause is the DWM/DirectComposition teardown failure for this specific `WindowStyle.None` + sheet-of-glass + Acrylic combination, and a true root-cause fix would require reworking the acrylic window architecture or manually tearing down the composition visual on minimise — both far more invasive. The behavioural change is consistent with the macOS popover model and the existing close behaviour.

Verified locally: minimise button, taskbar-click minimise, and `Win+D` all leave the desktop fully clickable (10×25 icon grid; previously only the first row was clickable after minimise).

## Scope

- [ ] CLI (`src/`)
- [ ] Docs / CI / config

## Checklist

- [x] `npm test` passes — C#-only change; the JS test suite is unaffected
- [x] New user-facing strings go through `dashboard/src/content/copy.csv` — n/a, no new strings
- [x] Commits follow conventional style (`fix(windows): ...`)

---

<details>

Not triggered — this PR is purely client-side window-state logic (no public exposure, auth/session, cross-endpoint, or external gateway changes).

</details>

<details>

Not requesting `@codex` review.

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * When the window is minimized, the embedded web content now hides instead of remaining visible.
  * Restoring the window makes the embedded web content visible again, improving consistency across minimize/restore actions.

### DIFF
--- a/TokenTrackerWin/DashboardWindow.cs
+++ b/TokenTrackerWin/DashboardWindow.cs
@@ -105,7 +105,29 @@ internal sealed class DashboardWindow : Window
         Content = _webView;
 
         Loaded += async (_, _) => await InitializeWebViewAsync();
-        StateChanged += (_, _) => SyncMaxGlyph();
+        StateChanged += (_, _) =>
+        {
+            SyncMaxGlyph();
+            // Minimising this window (WindowStyle.None + WindowChrome
+            // GlassFrameThickness=-1 "sheet of glass" + DwmExtendFrameIntoClientArea(-1)
+            // + Acrylic, hosting a WebView2CompositionControl / DirectComposition visual)
+            // leaves a transparent, hit-testable "ghost" at the window's pre-minimise
+            // rectangle: the DComp visual isn't torn down on minimise, so it keeps
+            // presenting / capturing mouse input over the desktop (you can see the desktop
+            // but can't click anything inside the old window rect; desktop icons stop
+            // highlighting). Collapsing the WebView2CompositionControl removes its DComp
+            // visual from the window's composition tree, so the ghost stops presenting /
+            // hit-testing, while the HWND itself truly minimises — staying taskbar- and
+            // Win+D-restorable. Restored to Visible on Normal. This covers every minimise
+            // path uniformly (the minimise button, taskbar click, Win+D "show desktop",
+            // right-click -> Minimize, programmatic SW_MINIMIZE) with no behavioural
+            // change (minimise still minimises-to-taskbar) and without Hide(), which broke
+            // Win+D restore (the HWND left DWM's minimised-windows list) and tray restore
+            // (WindowState stayed Minimized, so ShowDashboard re-fired this handler).
+            _webView.Visibility = (WindowState == WindowState.Minimized)
+                ? Visibility.Collapsed
+                : Visibility.Visible;
+        };
         KeyDown += (_, e) => { if (e.Key == System.Windows.Input.Key.Escape) Hide(); };
         _server.StatusChanged += OnServerStatusChanged;
     }


mm7894215:
### Code review

Found 2 issues:

1. Infinite hide loop and window stuck permanently hidden when restoring from tray (bug due to [DashboardWindow.cs](https://github.com/mm7894215/TokenTracker/blob/3c20d8eec5f71aa7a7ce9ee33f7b98c5abcf40a9/TokenTrackerWin/DashboardWindow.cs#L107-L122))

Calling `Hide()` in `StateChanged` leaves `WindowState` as `Minimized`. During `ShowDashboard()`, the first call to `Show()` triggers `StateChanged` while the state is still `Minimized`, triggering `Hide()` again immediately before the state can be reset to `Normal`.

2. Breaks double `Win+D` (Show/Restore Desktop) system workflow (bug due to [DashboardWindow.cs](https://github.com/mm7894215/TokenTracker/blob/3c20d8eec5f71aa7a7ce9ee33f7b98c5abcf40a9/TokenTrackerWin/DashboardWindow.cs#L107-L122))

Routing shell minimisation to `Hide()` takes the HWND out of DWM's active minimised windows list. When pressing `Win+D` again, the OS cannot restore the window automatically.

wen-shion:
Updated — rewrote the fix to address both review comments.

Instead of routing minimise through `Hide()` (which broke Win+D restore — the
HWND left DWM's minimised-windows list — and broke tray restore — `WindowState`
stayed `Minimized`, so `ShowDashboard()` re-fired `StateChanged` before the
state could reset), the `StateChanged` handler now collapses the
`WebView2CompositionControl` when `WindowState == Minimized` (restored to
`Visible` on `Normal`).

Collapsing the control drops its DirectComposition visual from the window's
composition tree, so the "ghost" (transparent, hit-testable rectangle left at
the pre-minimise bounds) stops presenting / hit-testing, while the HWND itself
truly minimises — staying taskbar- and Win+D-restorable. This covers every
minimise path (button, taskbar click, Win+D, right-click → Minimize,
programmatic) uniformly, with no behavioural change (minimise still
minimises-to-taskbar). `Hide()` is no longer used for minimise.

Verified locally: minimise button, taskbar-click minimise, and Win+D all leave
the desktop fully clickable; Win+D's second press restores the window; tray
double-click restores the window.

PR #291, opened 2026-07-09
fix: improve Codex refresh freshness

lamther:
## Summary

Fix Codex refresh staleness in the macOS app without making background polling heavier.

## What changed

- Trigger a throttled lightweight sync when the menu bar popover opens, then debounce repeated dashboard reloads.
- Refresh the installed local notify handler from `serve` and safely repair Codex notify when TokenTracker is missing.
- Preserve and runtime-chain an existing SkyComputerUseClient notify command when TokenTracker takes over Codex notify.
- Make `status` report Codex notify as configured only when it points at TokenTracker, so Sky-only configs no longer look healthy.
- Add regression coverage for Sky coexistence, stale notify backups, popover-open refresh routing, and status detection.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Verification

- [x] `node --test test/macos-popover-anchor-window.test.js test/init-uninstall.test.js test/status.test.js test/codex-config-multiline-notify.test.js test/local-api-background.test.js test/sync-background.test.js`
- [x] `npm test` (1328 tests passed)
- [x] `git diff --check`
- [x] `cd TokenTrackerBar && xcodebuild build -scheme TokenTrackerBar -destination 'platform=macOS'`

## Notes

The full `TokenTrackerBarTests` scheme currently fails before this change's tests due an unrelated test-target wiring issue: `WeeklyLimitResetDetector.swift` references `Strings`, which is not available in that test target. The macOS app build itself succeeds.

## Summary by CodeRabbit

* **New Features**
  * Popover opening now uses timing-aware “lightweight sync” logic and conditionally queues follow-up loads.
  * Startup now proactively repairs Codex notification wiring when needed.
* **Bug Fixes**
  * Notify command chaining avoids duplicating payload arguments; status now validates Codex notify via exact command match.
  * Uninstall messaging clarifies when restore is skipped because the current notify isn’t managed.
  * Improved notify/TOML parsing and safer restore gating for mismatched current configuration.
* **Tests**
  * Expanded coverage for popover refresh timing/queueing, Codex notify repair/backup behavior, status output, and multiline escape/restore scenarios.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Models/BackgroundRefreshPolicy.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Models/BackgroundRefreshPolicy.swift
@@ -4,6 +4,8 @@ enum BackgroundRefreshPolicy {
     static let defaultRefreshInterval: TimeInterval = 300
     static let defaultSyncInterval: TimeInterval = 300
     static let defaultCatchUpStaleInterval: TimeInterval = 300
+    static let defaultPopoverOpenSyncInterval: TimeInterval = 60
+    static let defaultPopoverOpenLoadInterval: TimeInterval = 30
 
     static func shouldRunSync(
         now: Date,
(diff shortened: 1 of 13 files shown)


mm7894215:
### Code review

Found 4 issues:

1. Infinite background process-spawning loop and security bypass risk when chaining SkyComputerUseClient without constraints. (bug due to removing exclusion in shouldChainNotify)

https://github.com/mm7894215/TokenTracker/blob/8f5b67064e8efdd4b66c3609e9e903a7a2f0d4d5/src/commands/init.js#L1151-L1156

2. Popover open sync events are silently dropped without being queued or retried if a background loading operation is already in progress. (bug due to isLoading guard in refreshForPopoverOpen)

https://github.com/mm7894215/TokenTracker/blob/8f5b67064e8efdd4b66c3609e9e903a7a2f0d4d5/TokenTrackerBar/TokenTrackerBar/ViewModels/DashboardViewModel.swift#L281-L283

3. Unreachable dead code in restoreNotify and uninstall.js caused by shadowing/subsuming conditional checks. (bug due to logic order in restoreNotify)

https://github.com/mm7894215/TokenTracker/blob/8f5b67064e8efdd4b66c3609e9e903a7a2f0d4d5/src/lib/codex-config.js#L58-L64

4. TOML config parser failure when argument values contain escaped double quotes because the scanChunk helpers do not handle backslash escapes. (bug due to escape-sequence mismatch in scanChunk)

https://github.com/mm7894215/TokenTracker/blob/8f5b67064e8efdd4b66c3609e9e903a7a2f0d4d5/src/lib/codex-config.js#L282-L308

PR #312, opened 2026-07-15
fix: refresh Claude and other local sources in desktop app

LinYYY:
## Summary

- make local dashboard reload run an all-local background sync so Claude, Gemini, OpenCode, and other local log/database sources refresh without cloud upload, Cursor network access, or deep Codex archive work
- repair existing runtime integrations when the desktop server starts, instead of repairing only Codex notify
- install Claude `Stop` plus `SessionEnd` hooks for per-response updates and session-end fallback
- allow hook-triggered local source sync without requiring a cloud device token
- add a native-shell fallback scan every 60 seconds, covering all local providers even when hooks are absent or overwritten
- re-read visible local dashboard aggregates every 30 seconds and on focus, with single-flight guards to prevent overlapping work

## Root cause

The DMG starts the embedded server with `--no-sync`. Dashboard reload previously used a bounded background path whose source allowlist only included Codex and Every Code. Desktop startup only repaired the Codex notify integration, and the generated notify handler required a cloud device token before starting any local parse.

After those paths were repaired, a second stale-data gap remained: a dashboard left open had no periodic aggregate refresh, and the native server had no fallback scan. A missed or overwritten provider hook could therefore leave local data stale indefinitely until the user clicked Refresh.

## Validation

- targeted Node tests for sync, local API, hook repair, reload, and native periodic sync passed
- dashboard Vitest coverage for local sync payloads and periodic refresh passed
- dashboard TypeScript check, ESLint, production build, copy checks, UI hardcode checks, and architecture guardrails passed
- macOS Release app and DMG built successfully with Xcode 26.6 and were installed as version 0.79.8
- installed notify handler runs without a cloud token; direct trigger advanced `last_notify_spawn` and `last_parse`
- installed native server advanced `last_parse` and the queue file on its 60-second fallback interval without a manual refresh
- local model-breakdown endpoint returned both Codex and Claude sources
- full Node suite: 1530/1531 passed under load; the single timing-sensitive concurrency assertion passed 6/6 when rerun in isolation

## Summary by CodeRabbit

- **New Features**
  - Added dashboard local auto-refresh for usage stats, including a safer reload-trigger flow.
  - Expanded local/background sync controls, including an “all local sources” option for lightweight runs.
  - Added native desktop background syncing with clean startup/shutdown behavior.
  - Added runtime integration repair during server startup and improved Claude usage hooks support (session + usage events).

- **Bug Fixes**
  - Improved local usage refresh sequencing after sync completes.
  - Updated Claude hook configuration detection to the new usage-hooks model.

- **Tests**
  - Added coverage for local sync option forwarding, dashboard auto-refresh behavior, runtime repair, and native background sync.

### DIFF
--- a/dashboard/src/lib/api.local-sync.test.ts
+++ b/dashboard/src/lib/api.local-sync.test.ts
@@ -1,0 +1,46 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { triggerLocalSync } from "./api";
+
+vi.mock("./local-api-auth", () => ({
+  getLocalApiAuthHeaders: vi.fn(async () => ({ "x-tokentracker-local-auth": "local-token" })),
+}));
+
+describe("triggerLocalSync", () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it("requests an all-local background sync for local dashboard reloads", async () => {
+    const fetchMock = vi.fn().mockResolvedValue(new Response(JSON.stringify({ ok: true }), {
+      status: 200,
+      headers: { "Content-Type": "application/json" },
+    }));
+    vi.stubGlobal("fetch", fetchMock);
+
+    await expect(
+      triggerLocalSync({ auto: true, background: true, allLocalSources: true }),
+    ).resolves.toEqual({ ok: true });
+    expect(fetchMock).toHaveBeenCalledWith(
+      "/functions/tokentracker-local-sync",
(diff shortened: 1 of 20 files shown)


mm7894215:
Merged — thanks @LinYYY. Verified the local-source refresh against real data here; totals stay consistent and it fixes the stale-dashboard gap on the desktop app. Ships with the next release.

PR #313, opened 2026-07-15
perf(codex): optimize context and sync hot paths

lamther:
## Summary

Reduce repeated Codex history work in context breakdowns and incremental syncs, especially for installations with large session inventories and large persisted cursor state.

This PR keeps the existing refresh cadence, notification behavior, and scheduler wakeups unchanged. It changes only how much historical work a refresh performs.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Why

Two hot paths still scaled with historical state even when little or nothing changed:

1. Context breakdowns repeatedly considered and parsed broad Codex history.
2. Normal append syncs materialized the full historical event-key collection as a `Set` and then copied it back to an array.

With hundreds of thousands of persisted keys, that work increases CPU time, allocation pressure, and energy use on every refresh.

### Bounded context discovery and parsing

- Select candidate sessions by requested date plus cached file metadata.
- Cache stable session parse results and invalidate on append, inode replacement, truncation, archive moves, and relevant timezone/date-boundary changes.
- Run the 60-second metadata audit only when a context request arrives; no new timer or wakeup is introduced.
- Preserve support for archives and duplicate/split session files.

### Incremental Codex sync

- Keep normal append processing proportional to newly read events instead of rebuilding a collection from all historical event keys.
- Avoid the historical `Set` construction and array copy for idle and safe append paths.
- Conservatively fall back to historical deduplication for rewrites, truncation, archive moves, and legacy cursors without `lastTotal`.
- Clear stale rollout hashes when cumulative-delta migration rebuilds Codex buckets, preventing rebuilt events from being skipped.

The parser hot path becomes `O(new events)` for safe appends. Reading and rewriting the monolithic `cursors.json` is intentionally unchanged and still scales with cursor size.

## Correctness invariants

- Refresh cadence and notification timing are unchanged.
- Rewrites, truncations, inode changes, archive moves, and legacy cursor shapes retain conservative recovery behavior.
- Append results, queue output, project queue output, and persisted cursor bytes remain deterministic against the baseline.
- The Context fixture change from `188` to `76` fixes an incorrect cross-date cumulative baseline; it is not a performance-driven data change.

## Benchmark

Baseline: upstream `v0.79.8`. Candidate: `32c48a3`. Same machine and deterministic corpus for each before/after pair.

### Synthetic 379k-key corpus, median of 5 runs

| Scenario | v0.79.8 | This PR | Change |
| --- | ---: | ---: | ---: |
| Context cold | 30.45 ms | 21.00 ms | -31.0% |
| Context warm | 1.157 ms | 0.513 ms | -55.7% |
| Idle parser | 145.28 ms | 1.06 ms | -99.3% |
| Active append | 143.77 ms | 6.14 ms | -95.7% |
| Cursor commit | 1282.63 ms | 214.22 ms | -83.3% |

The idle and append paths reduced historical `Set` construction / array materialization from `1 / 1` to `0 / 0`. Persisted cursor SHA-256 remained identical.

### Frozen real-shape idle corpus, two fresh 7-run batches

Corpus: 27,706 Codex files, 13.72 GiB logical session data, and a 61.38 MiB cursor. Baseline/candidate order alternated within each batch to reduce ordering bias.

| Batch | Metric | v0.79.8 | This PR | Change |
| --- | --- | ---: | ---: | ---: |
| A | Wall time | 942.33 ms | 890.94 ms | -5.5% |
| A | Total CPU | 1215.25 ms | 1146.23 ms | -5.7% |
| B | Wall time | 1075.40 ms | 1001.22 ms | -6.9% |
| B | Total CPU | 1371.14 ms | 1125.81 ms | -17.9% |

All 14/14 runs produced byte-identical cursor, queue, and project queue outputs with matching SHA-256 values.

End-of-process RSS varied with V8 GC timing across reruns, so this PR does not claim a stable memory reduction. The deterministic allocation counters above are the stronger evidence: the 379k-key idle and append paths no longer construct or copy the historical collection.

## Verification

- [x] `npm run ci:local`
- [x] 38/38 focused Codex context/sync tests
- [x] 5 independent synthetic before/after benchmark runs
- [x] Two fresh 7-run frozen real-shape before/after benchmark batches
- [x] Conventional commit title

`npm run ci:local` passed on the exact pushed commit, including the full test suite, copy/UI/architecture gates, and dashboard build.

## Risk and rollback

The main risk is stale-cache or deduplication behavior around file replacement and recovery paths. Tests cover append, rewrite, truncate, inode replacement, archive moves, duplicate/split sessions, timezone boundaries, and legacy cursor fallback. The change is confined to Codex parsing/context/sync internals and can be reverted as one commit.

## Known gap / out of scope

The full `cursors.json` payload is still serialized and rewritten on each sync. Replacing it with a sidecar/index or schema migration could yield a larger win for very large cursors, but it has a wider compatibility and recovery surface and is deliberately not included here.

<details>

### Risk layer triggers

### Rules / invariants

- Existing Codex totals and queue semantics remain unchanged outside the corrected cross-date Context baseline.
- Unsafe file-history transitions always use conservative recovery.
- No change to refresh cadence, notifications, or scheduler behavior.

### Boundary matrix

| Boundary | Expected behavior | Verification |
| --- | --- | --- |
| Stable historical session | Reuse cached parse | focused tests + context benchmark |
| Safe file append | Parse only new bytes/events | focused tests + append benchmark |
| Rewrite/truncate/inode change | Invalidate and recover conservatively | focused regression tests |
| Archive move/duplicate session | Preserve totals without double count | focused regression tests |
| Legacy cursor without `lastTotal` | Use historical dedup fallback | focused regression tests |

</details>

<details>
<summary><strong>Codex review context</strong></summary>

- **Delta since last Codex review:** New PR; one commit based on upstream `v0.79.8`.
- **Intended behavior / invariants:** Reduce repeated historical work without changing cadence or observable sync semantics.
- **Edge cases covered:** append, rewrite, truncate, inode replacement, archive move, duplicate/split session, timezone boundary, legacy cursor, cumulative-delta migration.
- **Tests run:** `npm run ci:local`; 38 focused tests; 5-run synthetic benchmark; two fresh 7-run frozen real-shape batches.
- **Known gaps / out of scope:** Monolithic cursor serialization and writeback remain unchanged.

Codex cache invalidation and historical event deduplication during uncommon file lifecycle transitions.

### Verification method

Deterministic before/after fixtures, byte/SHA output comparison, focused regression coverage, and full local CI.

### Uncovered scope

Long-duration production soak across sleep/wake cycles; cadence itself is unchanged.

</details>

## Summary by CodeRabbit

* **Performance**
  * Improved Codex rollout scanning and synchronization efficiency, especially for incremental updates and large histories.
  * Added caching and duplicate detection to reduce repeated parsing and avoid double-counting token usage.

* **Accuracy**
  * Refined context breakdowns, token attribution, session counts, date-range handling, and zero-token turn processing.
  * Improved handling of archived, split, rewritten, and overlapping rollout files.

* **Diagnostics**
  * Added synchronization and parsing diagnostics to help report discovery, filtering, caching, and processing activity.

* **Bug Fixes**
  * Corrected cursor and hash rebuilding behavior after rollout migrations and baseline resets.


mm7894215:
Verified this against real local data (large Codex history): per-source token totals match main exactly and a re-sync is idempotent, and the steady-state speedup is real. Nice work.

Before merging, could you drop the benchmark from the PR? It's a dev-only tool and doesn't need to live in the repo:

- `scripts/benchmark-codex-hot-paths.js`
- `test/codex-hot-path-benchmark.test.js` (only exercises that script)

Please keep the functional changes and the hot-path / sync / context tests. Once those two files are removed I'll merge.

lamther:
Thanks for validating this against the large local Codex history. I removed both repository benchmark artifacts in `8490cb4`:

- `scripts/benchmark-codex-hot-paths.js`
- `test/codex-hot-path-benchmark.test.js`

The functional changes and all retained context/sync hot-path tests are unchanged.

Validation after removal:

- `node --test test/codex-context-breakdown.test.js test/codex-context-hot-path.test.js test/codex-sync-hot-path.test.js` — 37/37 passed
- `npm run ci:local` — 1553/1553 tests passed, plus copy/UI/architecture validations and the dashboard production build
- Clean temporary merge onto current `main` (`4b8a9ef`) — 37/37 focused tests passed

Ready for re-review.

mm7894215:
Thanks — this is in good shape. The core paths (context grouping dedup, append-only vs historical sync switch, migration hash reset) are sound and well covered; CI is green and no correctness issue blocks the merge. Three small, non-blocking items I'd like tidied up first:

1. **`attributeTurn` drops buffered exec ends on zero-token turns** — `src/lib/codex-rollout-parser.js:653-659`. The early return now clears `pendingExecEnds` without absorbing them, so `exec_command_end` rows on a zero-token turn no longer carry over to the next turn (main used to let them). Token totals are unaffected, but the exec-command breakdown (by_executable / duration / call count) can under-report. Please run `for (const p of pendingExecEnds) absorbExecEnd(p)` before the reset, matching main.

2. **Parsed-group cache hit skips `seenTokenEvents` backfill** — `src/lib/codex-context-breakdown.js:248-266`. On a cache hit the group's token-event fingerprints aren't merged into the shared `seenTokenEvents`. It's harmless today because cacheable groups are partitioned by a unique rollout UUID, so cross-group keys never collide — but that's an implicit invariant. Please either backfill the fingerprints on hit, or add a short comment pinning down why it's safe.

3. **Diagnostics reduce runs even when diagnostics is off** — `src/lib/rollout.js:194-199`. `codexParseCandidates` does an O(n) reduce unconditionally, though it's only consumed under `if (syncDiagnostics)`. Gate it behind the flag so production sync skips the scan (the `filterColdCodexRolloutFiles` `discoveredRollouts` reduce has the same pattern).

None are blockers. Once these are in I'll merge.

lamther:
Thanks for the precise follow-up. Addressed all three items in `c780930`:

- Zero-token turns now absorb buffered `exec_command_end` rows before clearing pending state; token attribution remains unchanged.
- Documented the parsed-group cache invariant: cacheable groups are partitioned by rollout UUID, so cache hits do not need to backfill `seenTokenEvents`.
- Moved both Codex candidate-count reductions behind `syncDiagnostics`, so production paths skip those scans when diagnostics are disabled.

Regression coverage now verifies that zero-token turns retain exec call metadata without adding tokens, and uses a throwing `Array#reduce` Proxy to prove diagnostics-disabled filter/parser paths do not run either reduction.

Validation on the pushed commit:

- `node --test test/codex-context-breakdown.test.js test/codex-context-hot-path.test.js test/codex-sync-hot-path.test.js` — 39/39 passed
- `npm run ci:local` — 1555/1555 tests passed, plus copy/UI/architecture validations and the dashboard production build
- The two new regressions fail on pre-fix `8490cb4` and pass on `c780930`
- Clean temporary merge onto current `origin/main` (`4b8a9ef`) — no conflicts; 39/39 focused tests passed

Ready for re-review.

xiufengsun (approved):
Re-reviewed the latest head and the clean merge result on current main. The requested exec-stat preservation, cache invariant documentation, and diagnostics gating are addressed. Focused context/sync tests and required checks pass; no correctness blocker remains. Approved for merge.

PR #325, opened 2026-07-18
perf(codex): incrementally parse appended context rollouts

lamther:
## Summary

Keep Codex Context Breakdown refreshes current without repeatedly JSON-parsing multi-hundred-megabyte rollout prefixes. Growing single-file sessions validate the complete cached prefix and parse the appended JSONL records through one stable read handle; refresh cadence and API behavior are unchanged.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Why

A machine with about 28,900 Codex rollout files and 16 GB of sessions had roughly 1 GB of candidate data for a current-day Context query. Active 100-200 MiB sessions were fully JSON-parsed after every append, so a small refresh could consume hundreds of milliseconds to seconds of CPU.

This change reduces work per refresh instead of delaying, debouncing, or coalescing refreshes.

## Implementation

- Cache the parsed aggregate and a compact resume state per single-file rollout group.
- Use an O(n) scan to select only the most recently modified eligible single-file rollout for resumable SHA-256 state. Other sessions keep the conservative full-parse path, avoiding cold-path hashing across all sessions.
- On a valid append, open the file once, hash the complete cached prefix `[0, endOffset)`, and parse only the suffix through that same handle. Accept the merge only when metadata is unchanged, or when a concurrently unlinked handle passes a second full-snapshot SHA-256 check.
- Keep full parsing for overlapping live/archive fragments and every unsafe or non-selected state.
- Retain only cumulative token baselines, pending tool/skill names, sanitized exec summaries, metadata, final timestamp signatures, and internal hash state. Raw command arguments and tool output payloads are not cached.
- Add diagnostics for incremental hits, fallbacks, parser bytes read, and prefix-validation bytes.

## Safety invariants

Incremental parsing is allowed only when all of these hold:

- same file path, device, and inode
- file size strictly increased
- the prior parsed snapshot ended at a JSONL newline
- SHA-256 matches across the complete prior prefix `[0, endOffset)`
- prefix validation and suffix parsing use the same open file handle
- the handle metadata is unchanged after suffix parsing, or an unlinked handle's complete parsed snapshot hash still matches
- appended token timestamps do not move backwards

Replacement before validation, truncation, mutation anywhere in the prior prefix, incomplete records, unstable linked handles, timestamp regression, and multi-file overlap groups fall back to a fresh full parse. If the path is atomically replaced after validation, the already-open unlinked handle is rehashed and returned only as a coherent byte-identical snapshot; the new path is picked up on the next refresh.

## Performance

Benchmarked in alternating fresh Node processes against `origin/main` on the same frozen data:

| Case | Baseline median | Candidate median | Result |
| --- | ---: | ---: | --- |
| Cold parse: 64 sessions, 1.059 GB (6 samples each) | 5,788.7 ms | 5,966.1 ms | 3.1% one-time hash-state cost; identical result SHA-256 |
| Append refresh: 194 MiB session + 978,218 B (10 samples each) | 985.1 ms | 124.9 ms | about 7.9x faster / 87.3% lower latency |

The candidate validates 202,992,923 prefix bytes sequentially to prove append safety, but JSON-parses only the 978,218-byte suffix. Baseline and candidate produced identical full-result SHA-256 hashes before and after the append.

## Checklist

- [x] `npm test` passes
- [x] No new user-facing strings
- [x] PR description explains why, not just what

## Verification

- `node --check src/lib/codex-rollout-parser.js`
- `node --check src/lib/codex-context-breakdown.js`
- `node --test test/codex-context-breakdown.test.js test/codex-context-hot-path.test.js` (36/36)
- `npm run ci:local` (1,657/1,657 tests, copy/UI/architecture guardrails, production dashboard build)
- Real 1.059 GB cold-result equivalence benchmark
- Real 194 MiB append-result equivalence benchmark
- Regression mutates a prefix larger than 4 KiB near its beginning before appending and confirms the incremental path is rejected
- Regression atomically replaces the path between validation and suffix parsing, asserts exactly one matching open and a first-handle suffix read, then confirms the coherent old snapshot and next-refresh replacement equivalence

<details>

### Risk layer triggers

- [x] Cross-module parser/cache invariant

### Rules / invariants

- Never return or cache prompt text, assistant text, command arguments, or tool output payloads.
- Never merge an append into a cache entry unless identity, growth, full-prefix, record-boundary, stable-handle, and timestamp gates pass.
- Preserve synchronous freshness for active sessions; no debounce or polling interval change.

### Boundary matrix

| Boundary | Expected behavior |
| --- | --- |
| Same inode + valid append | Validate prefix and parse suffix on one open handle |
| Replaced before validation / truncated / mutated prefix | Full parse |
| Path replaced after the validated handle opens | Rehash the stable handle; read the new path next refresh |
| Partial final JSONL record | Full parse after completion |
| Timestamp regression | Abort resume and full parse |
| Live/archive overlap | Existing full merge/dedup path |

</details>

<details>
<summary><strong>Codex review context</strong></summary>

- **Intended behavior / invariants:** identical breakdown output with append work proportional to new bytes
- **Edge cases covered:** pending tools/skills/exec attribution, duplicate boundary events, prefix rewrite, incomplete record, concurrent path replacement, replacement/truncation, overlap groups, timestamp regression
- **Tests run:** see Verification
- **Known gaps / out of scope:** cursor-store generation commit cost and the existing 60-second metadata audit for previously pruned historical sessions

Cross-refresh attribution at the cached byte boundary.

### Verification method

Deterministic targeted tests plus full-result hashes on frozen production-scale fixtures.

### Uncovered scope

No refresh scheduler cadence or live tracker state was changed.

</details>

## Summary by CodeRabbit

* **Performance**
  * Added safer incremental/resumable parsing for cached Codex contexts, reusing previously parsed results when the file prefix is validated.
  * Improved JSONL incremental reading with byte-offset tracking to reduce hot-path work.

* **Reliability**
  * Strengthened resume safeguards with additional stability and identity checks; automatically falls back to full reparse when invalidated.

* **Diagnostics**
  * Added clearer visibility into incremental parsing behavior, including hit/fallback counters, prefix validation bytes, and bytes read.

* **Tests**
  * Expanded coverage for append-only updates, resume invalidation scenarios, and file-handle stability.

### DIFF
--- a/src/lib/codex-context-breakdown.js
+++ b/src/lib/codex-context-breakdown.js
@@ -266,35 +266,21 @@ function selectAppendHashPaths(candidateGroups) {
 }
 
 async function digestFilePrefix(
-  filePath,
+  sourceHandle,
   endOffset,
   diagnostics = null,
-  sourceHandle = null,
 ) {
   const end = Math.max(0, Number(endOffset) || 0);
   const hash = crypto.createHash("sha256");
   if (end === 0) return hash.digest("hex");
   let bytesRead = 0;
-  if (sourceHandle && typeof sourceHandle.read === "function") {
-    const buffer = Buffer.allocUnsafe(PREFIX_HASH_READ_BYTES);
-    while (bytesRead < end) {
-      const length = Math.min(buffer.length, end - bytesRead);
-      const result = await sourceHandle.read(buffer, 0, length, bytesRead);
-      if (!result.bytesRead) break;
-      hash.update(buffer.subarray(0, result.bytesRead));
-      bytesRead += result.bytesRead;
-    }
-  } else {
-    const stream = fs.createReadStream(filePath, {
-      start: 0,
-      end: end - 1,
-      highWaterMark: PREFIX_HASH_READ_BYTES,
-    });
-    for await (const value of stream) {
-      const chunk = Buffer.isBuffer(value) ? value : Buffer.from(value);
-      bytesRead += chunk.length;
-      hash.update(chunk);
-    }
+  const buffer = Buffer.allocUnsafe(PREFIX_HASH_READ_BYTES);
+  while (bytesRead < end) {
+    const length = Math.min(buffer.length, end - bytesRead);
+    const result = await sourceHandle.read(buffer, 0, length, bytesRead);
+    if (!result.bytesRead) break;
+    hash.update(buffer.subarray(0, result.bytesRead));
+    bytesRead += result.bytesRead;
   }
   if (diagnostics) {
     diagnostics.prefix_validation_bytes =
@@ -338,10 +324,9 @@ async function openValidatedResumeFile(cached, entry, diagnostics = null) {
 
     const expected = cached.contentHashState.copy().digest("hex");
     const actual = await digestFilePrefix(
-      entry.filePath,
+      handle,
       cached.endOffset,
       diagnostics,
-      handle,
     );
     if (actual !== expected) return null;
     if (diagnostics) diagnostics.stat_calls += 1;
@@ -377,15 +362,6 @@ function rememberParsedGroup(key, value) {
 }
 
 function splitCapturedParse(parsed) {
-  if (!parsed || typeof parsed !== "object") {
-    return {
-      parsed,
-      resumeState: null,
-      endOffset: 0,
-      appendable: false,
-      contentHashState: null,
-    };
-  }
   const {
     resumeState = null,
     endOffset = 0,
@@ -537,10 +513,9 @@ async function computeCodexContextBreakdown({
             try {
               const expectedSnapshot = resumed.contentHashState.copy().digest("hex");
               const actualSnapshot = await digestFilePrefix(
-                singleEntry.filePath,
+                resumeHandle,
                 singleEntry.stat.size,
                 diagnostics,
-                resumeHandle,
               );
               stableAfterParse = actualSnapshot === expectedSnapshot;
             } catch {
--- a/src/lib/codex-context-breakdown.js
+++ b/src/lib/codex-context-breakdown.js
@@ -246,24 +246,36 @@ function selectAppendHashPaths(candidateGroups) {
   return newest ? new Set([newest.filePath]) : new Set();
 }
 
-async function digestFilePrefix(filePath, endOffset, diagnostics = null) {
+async function digestFilePrefix(
+  filePath,
+  endOffset,
+  diagnostics = null,
+  sourceHandle = null,
+) {
   const end = Math.max(0, Number(endOffset) || 0);
   const hash = crypto.createHash("sha256");
   if (end === 0) return hash.digest("hex");
-  const stream = fs.createReadStream(filePath, {
-    start: 0,
-    end: end - 1,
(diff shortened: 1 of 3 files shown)


lamther:
Review disposition: the final outside-diff `reasoning_output_tokens` finding is pre-existing on `origin/main` (the same subtraction and hardcoded zero are present there) and is untouched by this PR. Baseline and candidate production fixtures produce identical result SHA-256 hashes. I am keeping it out of this performance PR; it needs a separate correctness PR with dedicated output-contract tests.

xiufengsun (changes requested):
Thanks for the performance work here. The overall direction is good, but I found one correctness blocker in the new incremental reader, so this is not merge-ready yet.

## 1. Correctness blocker: JSONL records larger than 64 KiB are corrupted

In `readStableHandleChunks`, every yielded chunk is a `subarray` of the same reusable 64 KiB buffer. `readCodexObjectsIncremental` then retains tails from incomplete lines in `fragments`. When a JSONL record spans multiple reads, the next `sourceHandle.read` overwrites the bytes referenced by the previously retained fragment. The reconstructed line is therefore corrupted, `JSON.parse` silently returns `null`, and the record is skipped while the parse is still reported as an incremental hit.

I reproduced this on the current head with an appended `exec_command` function-call record containing a 90 KiB argument, followed by a token-count record:

- cold parse: the 76-token delta is attributed to `Execution`;
- incremental parse: the same 76 tokens are incorrectly attributed to `Text Response`;
- total tokens remain identical, so a totals-only equivalence check does not detect the error;
- diagnostics still report `incremental_parse_hits: 1` and no fallback.

Please ensure that fragments retain owned bytes, for example by yielding/copying an independent buffer before the reusable buffer is filled again, and add a regression test with an appended JSONL record larger than 64 KiB. The test should compare the complete cold and incremental outputs, including `tool_calls_breakdown`, `skills_breakdown`, and `exec_command_breakdown`, not only grand totals.

Relevant code: `src/lib/codex-rollout-parser.js`, around lines 983-1021.

## 2. Remove unnecessary generalized/dead code while fixing this

Please keep this already-large hot-path change as small as possible and remove code that has no current consumer:

- `parseCodexLine` returns an `empty` field, but neither call site reads it. Return only the parsed object/null instead of carrying an unused result field.
- `digestFilePrefix` has a no-handle `createReadStream` branch, but both production call sites pass a stable handle, and the PR safety invariant explicitly requires validation and suffix parsing to use that same handle. Make the handle required and remove the unused path-based branch unless there is a real call site that needs it.
- `splitCapturedParse` contains a non-object fallback that is unreachable under the `parseCodexRolloutFile` return contract and both current call sites. Remove that defensive branch or show and test a real caller that can produce it.

The race-condition tests and resume-safety checks are useful and should stay. This cleanup request is specifically about unused fields and generalized branches, not about removing necessary safety coverage.

After the fix, please rerun the two targeted Context test files and the full local CI suite.

xiufengsun (approved):
Thank you for the thorough work on this optimization and for addressing the incremental-buffer ownership issue with a strong oversized-record regression test. The latest revision preserves cold/resumed correctness while delivering a meaningful refresh-speed improvement. Approved — great work!

PR #328, opened 2026-07-18
fix(i18n): label profile start date correctly

LceAn:
## Summary

- Correct the dashboard profile footer translations so the displayed date is labeled as the user's start date, not a leaderboard rank.
- Rename the `StatsPanel` prop and local value from `rankLabel`/`rankValue` to `startDate`/`startDateValue` to make the data contract explicit.
- Add regression coverage for Simplified Chinese, Traditional Chinese, Japanese, Korean, and German.

## Why

`DashboardView` passes the earliest usage date into this footer, and the English source copy has always said `Started`. Several translations instead said `Rank`, producing combinations such as `?? 2026-03-01` in both native desktop apps and the web dashboard.

The separately reported dashboard refresh flash in the released `v0.81.2` is already addressed on current `main` by commit `849b4b3` (`fix(dashboard): keep period refreshes in place`), which keeps rendered content mounted after the initial load. This PR does not duplicate that upstream fix.

## Verification

- `node --test test/localization-regressions.test.js test/dashboard-layout-adjustments.test.js`
- `node scripts/validate-copy-registry.cjs`
- `node scripts/ops/validate-ui-hardcode.cjs`
- `vitest run StatsPanel.test.jsx UsageOverview.test.jsx`
- `vite build`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the profile footer to display the localized start-date label instead of a ranking label.
  * Corrected German, Japanese, Simplified Chinese, and Traditional Chinese translations for the start-date text.

* **Tests**
  * Added regression coverage across multiple locales to verify the correct localized label and date display.

### DIFF
--- a/dashboard/src/content/i18n/de/core.json
+++ b/dashboard/src/content/i18n/de/core.json
@@ -40,7 +40,7 @@
   "error.boundary.hint": "Öffne die Konsole und lade die Seite neu",
   "error.boundary.no_details": "Keine Fehlerdetails verfügbar",
   "error.boundary.action.reload": "Neu laden",
-  "identity_card.rank_label": "Rang",
+  "identity_card.rank_label": "Beginn",
   "identity_card.streak_label": "Aktive Tage",
   "identity_card.rank_placeholder": "—",
   "identity_card.streak_value": "{{days}} Tage",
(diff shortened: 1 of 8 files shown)


xiufengsun (approved):
Thank you for fixing the misleading profile start-date translations and clarifying the StatsPanel data contract. The cross-locale regression coverage makes this a clean and reliable fix. Approved — great work!

PR #343, opened 2026-07-20
feat: show readable current device names

LceAn:
## Summary

- use the operating system hostname as the cloud device display name while keeping the stable anonymous machine ID as the device identity
- mark the current machine as "This device" in device cards and the device filter
- adopt legacy generated/hash names safely without overwriting names users customized themselves
- retain only the non-sensitive issued device UUID across page reloads so the current-device marker survives sync throttling; auth tokens remain memory-only

## Compatibility and privacy

- no database migration or new column is required because device identity and display name are already separate
- the hostname is used only as the display name when cloud sync is enabled
- existing user-renamed devices remain unchanged
- legacy dashboard and CLI-generated names continue to be recognized to avoid duplicate devices after upgrading

## Validation

- 52 device/local API Node tests
- 15 targeted dashboard tests
- dashboard TypeScript and ESLint checks
- copy, UI hardcode, and guardrail validation
- production dashboard build
- narrow-window browser layout check (360 px) with a long Windows hostname

## Summary by CodeRabbit

* **New Features**
  * Device names now prefer your system hostname when available, with deterministic machine-based disambiguation and safe fallbacks.
  * The dashboard highlights the current device with a localized “current” label/badge and keeps this selection consistent across sessions.
* **Bug Fixes**
  * Improved device adoption and rename handling to preserve customized names, avoid name collisions, and reliably recover stable naming after race conditions.
* **Localization**
  * Added the “current device” label for German, Japanese, Korean, Simplified Chinese, and Traditional Chinese.

### DIFF
--- a/dashboard/edge-patches/tokentracker-device-flow-poll.ts
+++ b/dashboard/edge-patches/tokentracker-device-flow-poll.ts
@@ -37,11 +37,13 @@ async function sha256Hex(input: string): Promise<string> {
 // deno-lint-ignore no-explicit-any
 async function issueDeviceToken(client: any, userId: string, clientInfo: string | null, machineId: string | null) {
   const platform = "cli-device-flow";
-  // The machine-id suffix keeps display names unique per machine: two Macs on
-  // the default "MacBook-Pro.local" hostname used to collapse into ONE device
-  // (the name-keyed unique index), overwriting each other's hourly rows and
-  // revoking each other's tokens on every login.
-  const deviceName = `TokenTracker CLI${clientInfo ? ` (${clientInfo})` : ""}${machineId ? ` #${machineId.slice(0, 8)}` : ""}`.slice(0, 128);
+  // New CLIs send `<platform>-<arch> <hostname>` as client_info. Identity is
+  // anchored by machine_id, so the human-facing name can be the actual system
+  // hostname instead of an opaque generated label.
+  const hostname = clientInfo?.replace(/^\S+\s+/, "").trim() || null;
+  const generatedLegacyName = `TokenTracker CLI${clientInfo ? ` (${clientInfo})` : ""}${machineId ? ` #${machineId.slice(0, 8)}` : ""}`.slice(0, 128);
+  const generatedLegacyBareName = `TokenTracker CLI${clientInfo ? ` (${clientInfo})` : ""}`.slice(0, 128);
+  const deviceName = (hostname || generatedLegacyName).slice(0, 128);
 
   // Device identity resolution — same machine-anchored scheme as
   // tokentracker-device-token-issue.ts: (1) reuse by (user, machine_id),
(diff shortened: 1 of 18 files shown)


mm7894215:
Thanks for the PR — the identity/display-name separation, legacy-name back-compat handling, privacy boundary (UUID in localStorage, token memory-only), and test coverage are all solid. CI is fully green. One blocking issue before merge, plus two notes:

**Blocking: same-hostname multi-machine registrations reintroduce a unique-index collision**

`tokentracker_devices_active_unique` (partial unique index on `(user_id, platform, device_name)` for non-revoked rows) still applies to machine-anchored rows. The old `#<machineId8>` name suffix existed specifically to fix a real production incident here (two Macs both defaulting to `MacBook-Pro.local` collapsing into one device — see the comment this PR removes, and the history around issue #187).

With this PR, a second new machine with the same hostname under the same account will:
1. hit the unique index on upsert (silently swallowed by `ignoreDuplicates`),
2. find nothing on the machine_id fallback lookup,
3. get a 500 from `tokentracker-device-token-issue` / `tokentracker-device-flow-poll`,

i.e. that machine can never cloud-sync until the user renames one of the hosts. Existing devices are unaffected (machine_id reuse path + the RPC swallows `unique_violation`), so this only hits newly-registered same-hostname machines — but that's exactly the scenario that caused the original incident.

Suggested fix (either is fine):
- On insert conflict, retry with a disambiguated name like `<hostname> #<id8>` while keeping the plain hostname as the primary display, or
- Keep the unique suffix in the DB `device_name` and show the readable hostname in the UI layer only.

Please also add a test covering two machines registering with the same hostname.

**Non-blocking notes**

1. Deploy order matters: the two edge functions (`tokentracker-device-token-issue`, `tokentracker-device-flow-poll`) must be deployed *before* any client release carrying this change — otherwise legacy `machine_id IS NULL` adoption will fail and device history splits.
2. Minor: the "this device" badge in `DashboardPage.jsx` is non-reactive (reads localStorage per render), so it only appears on the next render after the first sync establishes the session. Cosmetic only.

Happy to re-review once the collision fallback + test land.

LceAn:
Addressed in `8d79ef6` (rebased onto current `main` at `5ef8b55`).

Changes made from the review:

- Both `tokentracker-device-token-issue` and `tokentracker-device-flow-poll` now try the readable hostname first, re-select a concurrent machine-ID winner, then retry a genuine same-hostname collision as `<hostname> #<machineId8>` and re-select once more for a concurrent fallback insert.
- Added a behavioral regression test that executes both edge issuance paths against an in-memory database enforcing the active `(user_id, platform, device_name)` and machine-ID uniqueness rules. Two machines under the same user with `MacBook-Pro.local` now register as distinct rows: `MacBook-Pro.local` and `MacBook-Pro.local #bbbbbbbb`.
- Fixed the legacy CLI parsing case: `client_info = "darwin-arm64"` no longer becomes the display hostname and keeps its generated legacy label.
- Made the current-device marker reactive to the first successful cloud sync, and applied the suggested dropdown-label readability cleanup.

I agree with the deployment-order note: the two edge functions need to be deployed before a client release containing the hostname upload change so legacy machine-ID-less rows are adopted in place.

Validation completed: device/local API tests, targeted dashboard tests, TypeScript, ESLint, copy/UI/architecture validators, and the production dashboard build.

PR #373, opened 2026-07-27
fix(codex): repair indeterminate lineage history

lamther:
## Summary

- parse Codex JSONL by physical LF records instead of `node:readline`, preserving legal raw `U+2028` and `U+2029` inside JSON strings
- reuse the byte-level reader for full Codex context parsing and the historical lineage repair
- enforce exact byte budgets, bounded record memory, fatal UTF-8 decoding, and atomic fail-closed repair behavior

## Root cause

Codex permits raw `U+2028` and `U+2029` characters inside JSON strings. Node's `readline` treats those characters as line separators, so a valid physical JSONL record was split and failed `JSON.parse`. The one-time lineage repair then remained retryable forever with `usage_lineage_scan_indeterminate`, leaving previously inflated Codex totals unchanged.

## Safety

- physical records are framed from raw bytes using literal LF only
- CRLF and unterminated EOF byte spans are exact
- malformed JSON and invalid UTF-8 keep the repair indeterminate
- scan byte limits are enforced while chunks accumulate, so oversized unterminated records cannot exceed the memory/I/O budget
- guarded rebuilds leave main/project queues, upload offsets, file cursors, hashes, and usage buckets unchanged on failure

## Verification

- focused regression suite: 303 passed, 0 failed
- full `npm run ci:local`: passed, including 1,908 Node tests, copy/locale/UI/architecture guardrails, and dashboard production build
- independent code and test-coverage reviews: no remaining blocking findings
- read-only real-data scan: 31,038 candidate files, 42,681,543,098 bytes, 10,068,097 records, 135 affected files, 0 parse failures
- formerly blocked rollout: 68 physical records, 288,400 bytes, 0 parse failures, stable snapshot

No manual TokenTracker sync was run and no `~/.tokentracker` data or settings were modified during validation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Codex session and rollout JSONL parsing with strict byte-level record splitting, correct CRLF/LF handling across chunk boundaries, and accurate physical byte accounting (including unterminated tails).
  - Strengthened invalid UTF-8 handling with configurable behavior (throw, preserve corrupted records, or skip invalid lines), improving incremental resume and aggregation correctness.
  - Enhanced lineage scanning and max-bytes enforcement to prevent over-budget/truncated input from producing incorrect results.

- **Reliability**
  - Improved rescan/repair safeguards so repairs are deferred without mutating queue files, queue state, or cursor state when historical data is malformed or contains invalid UTF-8.

### DIFF
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -9,6 +9,7 @@ const { resolveInstallPaths, ensureFlatCursor } = require("../lib/install-resolv
 const { multiInstallParse, mergeBothFileSources } = require("../lib/multi-install-parser");
 const wsl = require("../lib/wsl-probe");
 const { ensureDir, readJson, writeJson, openLock } = require("../lib/fs");
+const { physicalJsonlRecords } = require("../lib/jsonl-lines");
 const {
   listRolloutFiles,
   listRolloutFilesDeep,
@@ -4265,7 +4266,6 @@ async function scanForInterleavedCodexUsage(
 
 async function scanCodexUsageLineages(filePath, maxBytes = Infinity) {
   let stream = null;
-  let rl = null;
   try {
     const before = await readCodexFileSnapshot(filePath);
     const state = createUsageDeltaState();
@@ -4276,9 +4276,9 @@ async function scanCodexUsageLineages(filePath, maxBytes = Infinity) {
       encoding: "utf8",
       highWaterMark: 32 * 1024,
     });
-    rl = readline.createInterface({ input: stream, crlfDelay: Infinity });
-    for await (const line of rl) {
-      bytesRead += Buffer.byteLength(line, "utf8") + 1;
+    for await (const record of physicalJsonlRecords(stream)) {
+      const { line } = record;
+      bytesRead += record.physicalBytes;
       if (bytesRead > byteLimit) return { affected: false, indeterminate: true };
       if (!line.trim()) continue;
       let obj;
@@ -4308,7 +4308,6 @@ async function scanCodexUsageLineages(filePath, maxBytes = Infinity) {
   } catch (_e) {
     return { affected: false, indeterminate: true };
   } finally {
-    if (rl) rl.close();
     if (stream) stream.destroy();
   }
 }
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -3819,6 +3819,7 @@ async function repairCodexRescanInflation({
       cursors: tmpCursors,
       queuePath: tmpQueue,
       projectQueuePath: tmpProjectQueue,
+      invalidRecordPolicy: "throw",
     });
     let tmpRaw = "";
     try {
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -4272,11 +4272,10 @@ async function scanCodexUsageLineages(filePath, maxBytes = Infinity) {
     const byteLimit = Number.isFinite(maxBytes) ? Math.max(0, maxBytes) : Infinity;
     let bytesRead = 0;
     let affected = false;
-    stream = fssync.createReadStream(filePath, {
-      encoding: "utf8",
-      highWaterMark: 32 * 1024,
-    });
-    for await (const record of physicalJsonlRecords(stream)) {
+    stream = fssync.createReadStream(filePath, { highWaterMark: 32 * 1024 });
+    for await (const record of physicalJsonlRecords(stream, {
+      maxPhysicalBytes: byteLimit,
+    })) {
       const { line } = record;
       bytesRead += record.physicalBytes;
       if (bytesRead > byteLimit) return { affected: false, indeterminate: true };
--- a/src/lib/codex-rollout-parser.js
+++ b/src/lib/codex-rollout-parser.js
@@ -920,6 +920,8 @@ async function parseCodexRolloutFile(filePath, {
   return result;
 }
 
+const fatalUtf8Decoder = new TextDecoder("utf-8", { fatal: true });
+
 function parseCodexLine(lineBuffer, diagnostics) {
   let content = lineBuffer;
   if (content.length > 0 && content[content.length - 1] === 0x0d) {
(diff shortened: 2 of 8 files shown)


xiufengsun:
Tested this against local data and hit a blocking regression. Please re-evaluate carefully before this lands.

`physicalJsonlRecords` decodes with `fatal: true` and nothing catches it. `parseRolloutFile` is the steady-state sync path, and `parseRolloutIncremental` has no per-file try/catch — so one bad byte throws out of the entire run: no cursors written, `enqueueTouchedBuckets` never reached, every Codex file in that run discarded. sync.js only logs `warnProviderParseFailure("Codex", …)`.

Same file (valid line / line containing `0xFF` / valid line):

```
main        no throw, 2 events, cursor written, 150 tokens queued
this branch TypeError: The encoded data was not valid for encoding utf-8
            cursor {}, no queue file at all, 0 tokens
```

Two triggers, neither needs disk corruption:

**1. Live session mid-flush.** The last record is cut inside a multi-byte character. Truncating a `turn_context` whose `cwd` holds CJK at 10/11/13/14 bytes throws every time on this branch; main parses the file fine. For CJK users with an active session that window is routine, not exotic.

**2. Upgrade path — and this one does not self-heal.** The current release tolerates a half character and persists `offset = st.size` mid-character. After upgrading:

```
release build:   cursor offset = 418, 100 queued
release again:   no throw, 150 queued
this branch:     TypeError, 0 queued
```

The throw happens before the cursor is advanced, so every later sync fails at the same offset. Codex tracking sits at zero until the file changes inode — archived sessions never do.

`parseCodexLine` in codex-rollout-parser.js already handles this correctly: same fatal decoder, wrapped in try/catch, returns null, line skipped. The hot path should fail the same way.

The U+2028 fix itself checks out — 150 tokens vs main's 50 on a rollout carrying a raw separator inside a `token_count` record. Perf is fine too (24.6MB synthetic file, 308–336ms vs main's 316–348ms). Fatal decoding is orthogonal to that fix: a `fatal` option on `physicalJsonlRecords` defaulting to non-fatal would keep the parser skipping bad lines, with `scanCodexUsageLineages` opting in so the repair stays fail-closed.

Worth adding a regression test as well: a rollout with an invalid UTF-8 byte mid-file should still count the records around it. Nothing covers that today, which is why CI is green.

lamther:
Confirmed and fixed in commit 83dd8f5. Thank you for the concrete reproduction.

The steady-state and one-time repair paths now have separate explicit policies:

- parseRolloutFile records invalid UTF-8 physical records and skips only a terminated bad record, so one bad line cannot abort the Codex run.
- An unterminated malformed JSON or UTF-8 tail does not advance the cursor. The cursor stays at the last committed physical record and the completed line is retried on the next sync.
- Legacy cursors starting inside a CJK UTF-8 sequence skip that residual terminated record and continue to later valid records.
- The project-context-only recovery path uses the same physical LF reader and committed-offset rule, including raw U+2028/U+2029 and legacy partial tails.
- scanCodexUsageLineages keeps the default strict UTF-8 policy. Any bad byte still returns usage_lineage_scan_indeterminate, preserving atomic fail-closed history repair.
- Invalid-record physical byte spans remain exact, so byte budgets and cursor evidence are not lost.

Added regressions for:

- valid / high-value invalid token_count / valid: 2 events, 150 tokens, cursor at EOF
- legacy cursor inside a CJK code point
- ASCII and mid-CJK unterminated tails completed across two syncs
- legacy project-context-only partial tail completed across two syncs
- raw U+2028 in the project-context-only path
- lineage invalid UTF-8 atomic defer

Verification on the final commit:

- focused suite: 311 passed, 0 failed
- independent final review: no blocking or important findings
- npm run ci:local: passed, including the full Node suite, copy/locale/UI/guardrail validation, architecture guardrails, and dashboard production build

lamther:
Follow-up hardening pushed in 41122f1 and 27e576a.

An independent review found one additional fail-closed bypass: an older rescan/fork repair could rebuild through the tolerant steady-state parser and then mark the lineage migration complete without the strict lineage scan. That could silently commit a lower historical total.

Fixed by adding an explicit invalid-record policy:

- steady-state sync defaults to skip and remains resilient
- every historical atomic rebuild explicitly uses throw
- strict mode validates UTF-8 and JSON for every physical record, including records outside the token fast path
- any strict failure occurs only in throwaway cursors and queues, leaving live cursors, queue, and queue state unchanged

Added an atomic regression with a corrupted historical token_count record. It verifies repair returns false and all live state remains byte-for-byte unchanged.

The prior CodeQL rollup also reported two test-only TOCTOU annotations. The fixtures now hold one FileHandle for stat, append, and utimes instead of reusing a checked path.

Final verification after both commits:

- focused suite: 312 passed, 0 failed
- independent final review: no Critical or Important findings
- npm run ci:local: passed again on HEAD 27e576a

Waiting for the refreshed GitHub checks and CodeQL annotations.

lamther:
Final follow-up pushed in `8cf398d`:

- narrowed final-record error handling so consumer errors propagate instead of being mistaken for decode failures
- guarded the malformed-record test marker lookup
- removed the unused import and unreachable branch
- added coverage for unsupported UTF-8 policies

Focused Codex sync suite: **314 passed, 0 failed**. Full `npm run ci:local`: **passed**.

xiufengsun:
Merged. Thanks for the thorough follow-up on the fatal-decode regression — the separate steady-state (skip) vs historical-rebuild (throw) policies are the right split.

Verified against local data before merging:

- raw U+2028 in a token_count record: 150 tokens vs main's 50
- mid-file 0xFF: no throw, surrounding usage intact, cursor reaches EOF
- CJK-truncated tail and a legacy cursor inside a UTF-8 sequence: no throw, completed on the next sync
- unterminated ASCII tail: main loses that record permanently once the cursor passes it, this branch picks it up on the next sync
- full read-only replay of 1113 real rollouts (5,016,583,146 tokens): totals, per-model breakdown and all 1113 cursor offsets byte-identical to main, and ~20% faster

I dropped the physicalJsonlLines wrapper on merge — it had no production caller left after the rollout.js import was removed, and its coverage folded cleanly into the record-mode tests.

Issue #374, opened 2026-07-27
WSL dual-install: sessions are parsed and counted, but never attributed to a project (projectKey stays null)


evolv3ai:
## Summary

With `TOKENTRACKER_WSL_MODE=both` on Windows, WSL sessions are discovered and their tokens **are** aggregated correctly — but they never receive a project identity. Every WSL session ends with `projectRef: null` / `projectKey: null`, so WSL work is absent from the projects view and from `tokentracker sessions` output, while the same sessions' tokens still appear in the usage/model totals.

Net effect: the dashboard undercounts **projects**, not spend. WSL work is invisible as projects/sessions but silently included in totals.

## Environment

- `tokentracker-cli` 0.85.0 (global npm install)
- Windows 11 Pro 26200, Node 25.9.0, npm 11.12.1
- `TOKENTRACKER_WSL_MODE=both`, `TOKENTRACKER_NO_TELEMETRY=1`
- Distros: `Ubuntu-24.04` (WSL2), `docker-desktop` (WSL2)
- Agents present in WSL: Claude Code, Codex CLI, OpenCode, Gemini, pi, Copilot

## What already works (not a discovery bug)

Discovery and parsing are solid, so this is narrowly an attribution bug:

- `tokentracker status` correctly reports both roots, e.g.
  `Claude Code: projects found (native: C:\Users\<u>\.claude\projects | WSL: \\wsl$\Ubuntu-24.04\home\<wsluser>\.claude\projects)`
- `cursors.json` holds 4,212 per-file cursors: 2,791 native + **1,421 under `\\wsl$\Ubuntu-24.04\...`**
- WSL transcripts are read **completely** — cursor `offset` equals the file size exactly (e.g. `1,149,060 == 1,149,060`).
- **Token aggregation works.** I took the largest WSL Claude transcript, independently parsed its `message.usage` records into `(model, 30-min slot)` pairs, and looked each up in `hourly.buckets`: **19 of 24 slots present**. The 5 missing are `<synthetic>` model entries carrying 0 tokens, correctly excluded. Native control session: 6/6.

## The bug

Cursor entry for a WSL Claude transcript (`\\wsl$\Ubuntu-24.04\home\<wsluser>\.claude\projects\-home-<wsluser>-dev-<project>\<uuid>.jsonl`):

```json
{
  "inode": 2494624,
  "offset": 1149060,
  "fileIdSize": 1149060,
  "claudeCwd": "/home/<wsluser>/dev/<project>",
[...]

PR #380, opened 2026-07-27
fix(session-analytics): discover Claude/Codex sessions in WSL installs too

evolv3ai:
Follow-up to the last section of https://github.com/mm7894215/TokenTracker/issues/374#issuecomment-5088580132 — the sessions pipeline scanning only the native home. Separate from the attribution fix in v0.86.1; no shared code.

## Why

`discoverSessionFiles` built its Claude and Codex roots from `os.homedir()` alone. `sync` already walks both Claude homes (`src/commands/sync.js`), so on a Windows host with a WSL install the same work is **counted in the token totals but missing from the session browser, its project list, and the `tracker sessions` export**. `TOKENTRACKER_WSL_MODE` had no effect on any of them.

That split is the confusing part in practice: the numbers look right, so nothing signals that the session-level views are showing a subset.

## What

Build the roots through the existing `wsl-probe` helpers (`shouldProbeNative` / `shouldProbeWsl` / `discoverWslHome`), the same way `sync.js` does, so one env var gates both sides consistently. Discovery is additive; duplicates synced between environments collapse via the existing Codex session-id pass and `claudeMessageDedupKey` downstream, so no new dedup logic.

WSL is probed **only when `home === os.homedir()`**. `discoverWslHome` resolves `\wsl$` independently of whatever `home` it's called alongside, so probing on an injected home splices the machine's live WSL transcripts into what the caller expects to be an isolated fixture. I hit this on the first attempt — two suite tests started seeing 467 real sessions where they asserted 1.

## Verification

Windows 11 + WSL2 Ubuntu, against `main` @ 0.87.0:

| `TOKENTRACKER_WSL_MODE` | rows | projects |
|---|---|---|
| `native-only` | 635 | 7 |
| `both` | 1082 | 20 |

`native-only` is byte-identical to pre-change output — the native path is untouched.

Tests: `providerRoots` is made dependency-injectable (`platform`, `homedir`, `discoverWslHome`) and exported, so root selection is exercised without needing a WSL install on the runner — the new file passes on Linux CI. Eight cases cover mode gating, an unresolvable WSL home, non-win32, and the isolation guard. Removing the guard fails the isolation test and leaves the other seven green.

Ran `session-analytics` + `local-api` + `wsl` + `dual-install` suites in `both` / `native-only` / `wsl-first` against this branch and against unpatched `main`: failure counts identical (11 / 13 / 11), pass counts +8 (the new tests). Those pre-existing failures are unrelated and environment-coupled — several suites read the real `~/.claude`, `~/.codex` and `\wsl$` shares rather than fixtures, so they don't reproduce on the ubuntu CI runner. Happy to file that separately if useful.

## Notes

- No user-facing strings, no Swift, no parser changes — privacy rule unaffected (metadata only; this changes *which directories are listed*, nothing about what's read from them).
- Grok is left native-only: `resolveGrokHome` has its own override handling and I didn't want to change that behaviour blind.

## Summary by CodeRabbit

* **New Features**
  * Enhanced session data discovery to scan both native Windows and WSL locations (multi-root).
  * Added/extended mode controls to support native-only, WSL-only, or combined discovery.
  * Improved multi-location discovery with de-duplicated results.

* **Bug Fixes**
  * Added robust fallback when WSL locations can’t be resolved, ensuring native results still load.
  * Prevented WSL probing on non-Windows platforms and ensured safe handling of mismatched home paths.

* **Tests**
  * Added coverage for WSL root discovery behavior, mode gating, and Codex vs. Claude path differences.

### DIFF
--- a/src/lib/session-analytics.js
+++ b/src/lib/session-analytics.js
@@ -31,6 +31,7 @@ const readline = require("node:readline");
 const { listClaudeProjectFiles, listRolloutFilesDeep, claudeMessageDedupKey } = require("./rollout");
 const { parseCodexRolloutFile } = require("./codex-rollout-parser");
 const { computeRowCost } = require("./pricing");
+const wsl = require("./wsl-probe");
 
 // Bump the sidecar when derived metrics change so cached rows are rebuilt
 // instead of leaving the dashboard on the previous (over-counted) heuristic.
@@ -761,14 +762,42 @@ async function scanGrokSession(filePath) {
   });
 }
 
-async function discoverSessionFiles(home) {
+// Multi-root discovery: on Windows a native install and a WSL one both count.
+// `sync` already walks both Claude homes (src/commands/sync.js), so without this
+// the WSL work lands in the token totals but is invisible in the session browser
+// and its project list. TOKENTRACKER_WSL_MODE gates which sides are probed;
+// duplicated files synced between environments collapse via the Codex session-id
+// pass below and claudeMessageDedupKey downstream.
+function providerRoots(home, providerDir, env) {
+  const roots = [];
+  if (process.platform !== "win32" || wsl.shouldProbeNative(env)) {
+    roots.push(path.join(home, providerDir));
+  }
+  // Only the real native home has a WSL sibling worth probing. discoverWslHome
+  // resolves \\wsl$ independently of `home`, so callers that inject their own
+  // home (tests, a custom HOME) would otherwise get the machine's live WSL
+  // sessions spliced into an isolated fixture.
+  const isRealHome = path.resolve(home) === path.resolve(os.homedir());
+  if (process.platform === "win32" && isRealHome && wsl.shouldProbeWsl(env)) {
+    const wslRoot = wsl.discoverWslHome(providerDir, { env });
+    if (wslRoot) roots.push(wslRoot);
+  }
+  return [...new Set(roots)];
+}
+
+async function discoverSessionFiles(home, env = process.env) {
   const grokHome = resolveGrokHome(home);
-  const [allClaude, codex, archived, grok] = await Promise.all([
-    listClaudeProjectFiles(path.join(home, ".claude", "projects")),
-    listRolloutFilesDeep(path.join(home, ".codex", "sessions")),
-    listRolloutFilesDeep(path.join(home, ".codex", "archived_sessions")),
+  const claudeRoots = providerRoots(home, ".claude", env);
+  const codexRoots = providerRoots(home, ".codex", env);
+  const [claudeGroups, codexGroups, archivedGroups, grok] = await Promise.all([
+    Promise.all(claudeRoots.map((r) => listClaudeProjectFiles(path.join(r, "projects")))),
+    Promise.all(codexRoots.map((r) => listRolloutFilesDeep(path.join(r, "sessions")))),
+    Promise.all(codexRoots.map((r) => listRolloutFilesDeep(path.join(r, "archived_sessions")))),
     listGrokSessionFiles(path.join(grokHome, "sessions")),
   ]);
+  const allClaude = [...new Set(claudeGroups.flat())];
+  const codex = [...new Set(codexGroups.flat())];
+  const archived = [...new Set(archivedGroups.flat())];
   // Claude Memory stores thousands of background observer transcripts beside
   // real Claude Code sessions. They contain <synthetic>/haiku bookkeeping and
   // no user coding outcome, so scanning them both slows the card dramatically
--- a/src/lib/session-analytics.js
+++ b/src/lib/session-analytics.js
@@ -776,12 +776,19 @@ function providerRoots(home, providerDir, env, deps = {}) {
   if (platform !== "win32" || wsl.shouldProbeNative(env)) {
     roots.push(path.join(home, providerDir));
   }
-  // Only the real native home has a WSL sibling worth probing. discoverWslHome
-  // resolves \\wsl$ independently of `home`, so callers that inject their own
-  // home (tests, a custom HOME) would otherwise get the machine's live WSL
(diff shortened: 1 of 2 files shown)


evolv3ai:
Pushed `09f49a8`, and adding verification from a second WSL host.

### The CI caveat, stated plainly

**The green checks on this PR do not validate the WSL code paths.** GitHub runners have no WSL distro and no `\wsl$` share, so `discoverWslHome` returns null there and every WSL branch is skipped. CI confirms the change is inert where WSL is absent and that the unit tests are hermetic — nothing more. I should have said that when I opened this rather than letting a green tick imply coverage it doesn't provide.

### Independent verification on real hardware

Run on a second Windows 11 + WSL2 box, distro `Ubuntu-24.04`, separate from the machine the PR was written on:

- `providerRoots` resolved to `\wsl$\Ubuntu-24.04\home\<user>\.claude`; every probe succeeded, no `WinError 3`, no `\wsl.localhost` fallback needed. That box is the one where the UNC-prefix fragility noted in #374 would surface, and it didn't.
- Sessions delta: **570 → 840 rows (+270), 43 → 52 projects.** Strict multiset superset, zero rows lost.
- Origin confirmed at the filesystem rather than inferred from the metadata output: 8 of the 9 gained projects have a WSL project directory and no native one. The largest single contributor (198 rows, 73% of the delta) traces to 198 WSL Codex rollout files and 0 native.
- A third control run ruled out elapsed time as an explanation: `native-only` re-run after `both` returned byte-identical to the first run, 0 differences.

Full suite there: **1878 pass / 62 fail on this branch vs 1865 / 67 on `5e31c6e`** with no PR commits — same machine, same freshly-installed deps. Diffing the failure *sets* rather than the counts: **0 failures unique to this branch**, 5 unique to base (timing-sensitive, flaky), 62 shared and pre-existing. Both failures whose names touch this area — `readClaudeScienceFrames snapshots a WAL DB reached over a WSL UNC path` and `Vite dev server handles current repo session analytics` — fail identically on base and branch.

Caveat worth keeping: one paired run can't fully exclude a regression that flaked green. The change is one function with dedicated tests, so I'd call that low but not zero.

### What `09f49a8` changes

The verification flagged a latent trap I'd left in. The home-identity check that keeps injected homes hermetic is **inert on the real CLI path** — `buildSessionAnalytics` defaults `home` to `os.homedir()`, so the comparison is always true there — and fires only for a threaded home. But threading a non-default home is precisely how you'd scan a tree that *isn't* `os.homedir()`, and such a caller would have lost WSL discovery silently, with no error and no failing test.

Same default, now expressed as a named `probeWsl` dep so the decision is reachable and documented, with two tests pinning the escape hatch in both directions (10 tests total). Red-green re-checked: forcing the default to always-probe fails the isolation test alone.

Happy to squash if you'd prefer a single commit.

xiufengsun (changes requested):
Thanks for the detailed real-hardware verification. I re-reviewed the current head `09f49a8b33747765eead8d113fbcbf7cff1c4ad5` against `main`, and there are two correctness issues that need to be addressed before merge.

1. **Mirrored Claude sessions are double-counted across native and WSL roots.**

   `new Set(claudeGroups.flat())` only deduplicates identical path strings, so the native path and its WSL/UNC copy both survive. `claudeMessageDedupKey` cannot collapse them here: `scanClaudeSession()` deliberately scopes `seenMessages` per file. Downstream, `summarizeSessions()` counts both rows, while `mergeSessionFragments()` merges the browser row by *adding* their token/cost totals.

   I reproduced this with byte-identical native and WSL copies of one Claude session containing 12 tokens:

   - expected: summary `1 session / 12 tokens`, browser `1 session / 12 tokens`
   - actual: summary `2 sessions / 24 tokens`, browser `1 session / 24 tokens`

   Please add deterministic cross-root duplicate ownership for identical mirrored copies while preserving legitimate divergent resume/subagent fragments that share a session ID. The regression coverage should include both an identical-copy case (count once) and a divergent-copy case (retain the non-overlapping fragments). A simple session-ID-only drop would lose legitimate fragments.

2. **The generic root selector turns `wsl-first` and `native-first` into `both`, including for Codex.**

   `shouldProbeNative()` and `shouldProbeWsl()` both return true for the two `*-first` modes, so `providerRoots()` returns both roots whenever both exist. That contradicts the documented mode contract: `wsl-first` selects WSL with native fallback, `native-first` selects native with WSL fallback, and only `both` aggregates both.

   Claude sync intentionally unions allowed installs, but Codex sync still uses `resolveInstallPaths()` and selects one install in the `*-first` modes. The current PR therefore makes Codex session analytics scan native + WSL while Codex token totals scan only the selected install. Please split the provider policies (Claude may keep its intentional union; Codex should follow the existing resolver semantics) and add a Codex mode matrix covering `wsl-first`, `native-first`, `both`, `wsl-only`, and `native-only`.

The PR-specific and surrounding tests are otherwise healthy. I also ran the full local suite; its two failures reproduce unchanged on the base commit and are unrelated. The green CI does not catch the two issues above because the new tests exercise root selection only and currently assert the incorrect default union, without testing cross-root mirrored Claude content.

xiufengsun:
Rechecked the exact current head 09f49a8b33747765eead8d113fbcbf7cff1c4ad5 against current main after the v0.87.10 release. This remains **NO-MERGE / CHANGES REQUESTED**.\n\nThe two merge blockers from the prior review are still unresolved in this head:\n1. Mirrored Claude sessions discovered through native and WSL roots can be counted twice; de-duplication is path-based and does not establish identical-copy ownership while preserving divergent fragments.\n2. The generic root selector still maps provider modes such as wsl-first/native-first to both roots, which is inconsistent with the provider-specific Codex semantics.\n\nPlease push a new head with provider-specific root policy plus adversarial tests for identical native/WSL copies and divergent fragments, and then request re-review. The green CI does not clear these semantic blockers.

PR #382, opened 2026-07-28
fix(grok): cache context breakdown like Codex/Claude

Gyarados4157:
## Summary
- Grok **Context Breakdown** was re-scanning every `updates.jsonl` whenever the dashboard switched provider tabs (Claude ↔ Codex ↔ Grok), so users with large Grok histories saw repeated “Scanning session logs…”.
- Align Grok with the Codex/Claude approach:
  - **Per-file parse cache** (`FILE_PARSE_CACHE`, size+mtime keyed) — same idea as Codex `PARSED_GROUP_CACHE`
  - **In-flight request coalescing** so concurrent panel mounts share one scan
  - **Result cache** still keyed with inventory signature for correctness when files change
  - **UI sticky cache** in `ContextBreakdownPanel` so revisiting a source keeps the last payload instead of flashing the skeleton (`loading && !data`)

## API surface
No API change. Still `GET /functions/tokentracker-usage-category-breakdown?source=grok|codex|claude` with the same response shape.

## Test plan
- [x] `node --test test/grok-context-breakdown.test.js` (includes result cache, in-flight, file invalidation)
- [x] `dashboard` vitest `ContextBreakdownPanel.test.jsx` (sticky source revisit)
- [ ] Manual: open Usage Overview → expand Grok → wait for first scan → switch to Claude/Codex → switch back to Grok (should not re-show long scan)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the context breakdown panel from flashing back to the loading state when revisiting a previously viewed source.
  * Avoided stale breakdown content by showing a loading hint when switching to an uncached source until fresh data is ready.
* **Performance**
  * Improved repeated Grok context breakdown requests with stronger caching and automatic coalescing of duplicate in-flight computations.
  * Ensured breakdown totals update when underlying session/update data changes.
* **Tests**
  * Added coverage for cached vs. uncached behavior and concurrency/deduplication.

### DIFF
--- a/dashboard/src/ui/dashboard/components/ContextBreakdownPanel.jsx
+++ b/dashboard/src/ui/dashboard/components/ContextBreakdownPanel.jsx
@@ -703,7 +703,9 @@ export function ContextBreakdownPanel({ from, to, source = "claude", referenceTo
     let cancelled = false;
     const cacheKey = `${from || ""}|${to || ""}|${source || "claude"}`;
     const cached = cacheRef.current[cacheKey];
-    if (cached) setData(cached);
+    // On a cache miss, reset data so the previous source's payload never
+    // renders under the new source's display branch while the fetch runs.
+    setData(cached || null);
     setLoading(true);
     onLoadingChange?.(true);
     setError(null);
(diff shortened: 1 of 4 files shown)


xiufengsun:
Pushed a follow-up commit addressing the CodeRabbit finding: on a cache miss the panel now resets `data` explicitly, so the previous source's payload can no longer flash under the new source's display branch while the fetch runs. Added a regression test for the uncached-source visit (12/12 vitest pass locally). Will merge once CI is green.

xiufengsun:
CI green with the follow-up. Thanks for the caching work — the per-file parse cache plus in-flight coalescing is exactly the Codex/Claude pattern this needed. Merging.

PR #397, opened 2026-07-31
fix(codex): scan both Windows and WSL installs, fix WSL shadow dropping Windows usage

HandSonic:
## Summary

`resolveInstallPaths` in `wsl-first`/`native-first` modes returns a **single** path (the preferred install) with `wsl=null`. For codex this is wrong — a user may have it installed in **both** Windows and WSL, and each install has its own `sessions/` tree. Picking one silently drops the other's usage.

Additionally, when the WSL `~/.codex` exists but has **no `sessions/` subdir** (codex installed on Windows, not in WSL), `wsl-first` mode picked the empty WSL shell as the single install, causing `listRolloutFiles` to scan a non-existent `sessions/` dir, silently return 0 files, and **all Windows codex usage went missing**. This reproduced as codex collection suddenly stopping at the date WSL was installed.

## Root cause

`src/lib/wsl-probe.js` `resolveAllWin32Paths` collapses to a single path in all modes except `both`. `discoverWslHome` only checks the WSL `~/.codex` dir **exists**, not that it has a `sessions/` subdir. Combined, a hollow WSL shell won over the real Windows install.

## Fix

In `src/commands/sync.js`, instead of trusting the single-path resolution, collect **every** codex install whose `sessions/` actually exists and push each as its own source. The `codexHashes` event-dedup (keyed by `sessionUUID:eventTimestamp`) makes overlapping scans idempotent even if the same file is reachable from both paths.

```js
const codexSessionsExists = (p) => {
  if (!p) return false;
  try { return fssync.existsSync(path.join(p, "sessions")); } catch { return false; }
};
const codexInstalls = new Set();
if (codexSessionsExists(codexNativeValue)) codexInstalls.add(codexNativeValue);
if (codexSessionsExists(wslCodexDir)) codexInstalls.add(wslCodexDir);
// Also honor resolveInstallPaths pick (env-overridden paths, future logic).
const codexPaths = resolveInstallPaths({ nativeValue: codexNativeValue, wslValue: wslCodexDir });
if (codexPaths.native && codexSessionsExists(codexPaths.native)) codexInstalls.add(codexPaths.native);
if (codexPaths.wsl && codexSessionsExists(codexPaths.wsl)) codexInstalls.add(codexPaths.wsl);
for (const install of codexInstalls) {
  sources.push({ source: "codex", sessionsDir: path.join(install, "sessions"), ... });
}
```

## Scope

- [x] CLI (`src/`)

## Verification

On a machine where WSL Ubuntu has a `~/.codex` shell without `sessions/` and codex is installed on Windows:

- **Before**: `runCodexParse input: 0 files (0 codex)` — `sources=[{sessionsDir:"\\wsl$\Ubuntu\home\kylin\.codex\sessions", exists:false}]`, codex queue records after 2026-07-14: **0**
- **After**: `runCodexParse input: 988 files (988 codex)` — Windows install scanned, codex queue records 2026-07-14 onward: **498** (7/14 → 7/23 recovered; 7/24+ pending cursor reset which is a data-state issue, not a code bug)

## Checklist

- [x] New test `test/codex-wsl-shadow.test.js` passes (verifies Windows install is scanned when WSL shell lacks `sessions/`)
- [x] `install-resolver.test.js`: 28/28 pass
- [x] `dual-install-integration.test.js`: 1/1 pass
- [x] No regression in `codex-source-scoped-cache` / `codex-config-multiline-notify` (38/39 pass; the 1 failure is pre-existing — fails without this PR too)
- [x] Commits follow conventional style (`fix:`)

## Why both installs, not just fallback

A simpler fix would be "if WSL lacks `sessions/`, fall back to Windows". But that still loses WSL usage when the user genuinely has codex in both environments. Scanning **both** (when each has `sessions/`) is the correct behavior — no usage is dropped regardless of where codex is installed.

## Note on other tools

`every-code`, `claude`, `gemini`, `opencode` use the same `resolveInstallPaths` pattern and may have the same shadowing issue, but they are out of scope for this PR. If maintainers want, I can follow up with the same treatment for them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex syncing across native and WSL installations.
  * Prevented empty WSL installations from hiding valid native session data.
  * Included active and archived sessions when available.
  * Prevented duplicate session records during repeated syncs.

* **Tests**
  * Added coverage for missing, empty, and populated WSL installations.
  * Verified supported installation-selection modes and token totals.
  * Improved cross-platform path handling in synchronization checks.

### DIFF
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -589,17 +589,31 @@ async function cmdSync(argv, context = {}) {
       const wslCodexDir = process.platform === "win32" && wsl.shouldProbeWsl(process.env)
         ? wsl.discoverWslHome(".codex")
         : null;
+      // resolveInstallPaths in wsl-first/native-first modes returns a SINGLE
+      // path (the preferred install) with wsl=null, which silently drops the
+      // other install's usage. For codex a user may have it installed in BOTH
+      // Windows and WSL, and each install has its own sessions/ tree — we must
+      // scan both or usage goes missing. So instead of trusting the single-path
+      // resolution, collect every install whose sessions/ actually exists and
+      // push each as its own source. The codexHashes event-dedup (keyed by
+      // sessionUUID:eventTimestamp) makes overlapping scans idempotent even if
+      // the same file is somehow reachable from both paths. (issue #codex-wsl-shadow)
+      const codexSessionsExists = (p) => {
+        if (!p) return false;
+        try { return fssync.existsSync(path.join(p, "sessions")); } catch { return false; }
+      };
+      const codexInstalls = new Set();
+      if (codexSessionsExists(codexNativeValue)) codexInstalls.add(codexNativeValue);
+      if (codexSessionsExists(wslCodexDir)) codexInstalls.add(wslCodexDir);
+      // Also honor resolveInstallPaths' pick (it may include env-overridden
+      // paths or future resolution logic we don't want to bypass entirely).
       const codexPaths = resolveInstallPaths({ nativeValue: codexNativeValue, wslValue: wslCodexDir });
-      if (codexPaths.native) {
-        sources.push({ source: "codex", sessionsDir: path.join(codexPaths.native, "sessions"), codexInventoryCache: true });
+      if (codexPaths.native && codexSessionsExists(codexPaths.native)) codexInstalls.add(codexPaths.native);
+      if (codexPaths.wsl && codexSessionsExists(codexPaths.wsl)) codexInstalls.add(codexPaths.wsl);
+      for (const install of codexInstalls) {
+        sources.push({ source: "codex", sessionsDir: path.join(install, "sessions"), codexInventoryCache: true });
         if (!isBackgroundLightweightSync || backgroundCodexUsageRepair) {
-          sources.push({ source: "codex", sessionsDir: path.join(codexPaths.native, "archived_sessions"), deep: true });
-        }
-      }
-      if (codexPaths.wsl) {
-        sources.push({ source: "codex", sessionsDir: path.join(codexPaths.wsl, "sessions"), codexInventoryCache: true });
-        if (!isBackgroundLightweightSync || backgroundCodexUsageRepair) {
-          sources.push({ source: "codex", sessionsDir: path.join(codexPaths.wsl, "archived_sessions"), deep: true });
+          sources.push({ source: "codex", sessionsDir: path.join(install, "archived_sessions"), deep: true });
         }
       }
     }
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -589,31 +589,26 @@ async function cmdSync(argv, context = {}) {
       const wslCodexDir = process.platform === "win32" && wsl.shouldProbeWsl(process.env)
         ? wsl.discoverWslHome(".codex")
         : null;
-      // resolveInstallPaths in wsl-first/native-first modes returns a SINGLE
-      // path (the preferred install) with wsl=null, which silently drops the
-      // other install's usage. For codex a user may have it installed in BOTH
-      // Windows and WSL, and each install has its own sessions/ tree — we must
-      // scan both or usage goes missing. So instead of trusting the single-path
-      // resolution, collect every install whose sessions/ actually exists and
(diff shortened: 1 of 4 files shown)


xiufengsun (changes requested):
Reviewed exact head `57545e794585fccdccc3a6398f5b3689677ed2f4` merged onto current `main` `739e759950e4a470a3886feaee9868b5bf4ba5b6`.

NO-MERGE: this fixes the reported empty-WSL-shell case by bypassing the install-mode resolver, but that changes the meaning of every WSL mode and regresses existing sync behavior.

The new block in `src/commands/sync.js:592-618` unconditionally adds both native and WSL installs whenever each has a `sessions/` directory. That means `wsl-first`/`native-first` become union scans instead of fallback modes, and even `wsl-only` can scan native data. The exact-head CI failure is real; I reproduced it locally with:

```
node --test test/codex-wsl-shadow.test.js test/codex-source-scoped-cache.test.js test/sync-codex-rescan-repair.test.js
```

Result: 57 passed, 4 failed (`full auto sync still scans flat Codex archives`, the bounded lineage scan diagnostic, background refresh liveness, and native lineage repair; the last replayed 4,700,000,000 tokens instead of 380).

Also, `test/codex-wsl-shadow.test.js:64-92` does not simulate Windows or a discovered WSL home: it never mocks `process.platform` and creates only `$HOME/.codex`. It therefore passes on the pre-existing native path and does not prove the reported scenario.

Please keep `resolveInstallPaths` as the authority for `wsl-first`, `native-first`, `wsl-only`, `native-only`, and `both`, but filter/validate candidates by the existence of their `sessions/` directory before selecting a preferred fallback. Add a real Windows-mode matrix test with both native and WSL candidates, including an empty WSL `.codex`, then rerun the three suites above.

HandSonic:
Review addressed in commit `1cdc771`:

**1. Mode semantics restored** — the union-scan block in `cmdSync` is deleted; `resolveInstallPaths` is again the single authority for `wsl-first`/`native-first`/`wsl-only`/`native-only`/`both`. Candidate validation moved into the resolver as an opt-in `requireAnyChild` filter: a candidate qualifies only if it holds `sessions/` or `archived_sessions/`. Empty WSL shell falls back to native in `wsl-first`/`auto`; `wsl-only` never selects native (`shouldProbeNative` gate untouched — CodeRabbit #1). `archived_sessions` inclusion keeps archive-only installs eligible — this was the root cause of the 4 suite failures: the union block's `sessions/`-only gate found zero sources in archive-only fixtures, so lineage repair silently never ran and buckets kept the seeded 4,700,000,000 instead of being repaired to 380.

**2. Three suites green** — `node --test test/codex-wsl-shadow.test.js test/codex-source-scoped-cache.test.js test/sync-codex-rescan-repair.test.js` → **69/69 pass**, both with isolated HOME and on a real Windows host with an actual empty WSL shell. Control: stashing the fix on the old head reproduces your exact "4 failed".

**3. Real Windows-mode matrix test** — `codex-wsl-shadow.test.js` rewritten hermetic: mocked `process.platform=win32`, `discoverWslHome` mocked per-case (no host WSL dependency), WSL \.codex in 3 states (missing/empty-shell/populated) × modes wsl-first/auto/wsl-only + `both`. Empty WSL falls back to native; populated WSL preferred; wsl-only with empty WSL yields zero. Verified to fail on the pre-fix head (cases 3,5,6,7,8). CodeRabbit #3 (run sync twice, assert zero delta) and #4 (env cleanup deletes originally-absent vars) included.

Mode matrix (15/15 verified): wsl-first → WSL when populated else native; native-first/native-only → native; wsl-only → WSL or nothing; both → union of populated.

One extra 1-liner: `sync-codex-rescan-repair.test.js:430` assertion used a literal `.git/config` path that fails on Windows (backslash separators) — verified failing identically on main; switched to `path.join(".git","config")`.

xiufengsun (approved):
Re-reviewed exact head 1cdc771e2d7b38fefa3c8a12f1faaba22aaf918b against the current main 3ff11b056aeeed921302444baad464d04b3b92fd in a clean temporary merge. Codex WSL shadow matrix and resolver/rescan suites pass 91/91; the merge is conflict-free and preserves all configured WSL/native modes.

PR #399, opened 2026-07-31
fix(codebuddy): collect traces as authoritative token source, fix 90% undercounting

HandSonic:
## Summary

CodeBuddy CLI writes authoritative token usage to `~/.codebuddy/traces/<pid>/trace_*.json` (and `~/.workbuddy/traces/` for WorkBuddy). Each trace JSON carries `trace.modelInfo.{totalInputTokens, totalOutputTokens, totalCachedTokens, callCount}` with 100% field stability (verified on 286 real traces).

The existing jsonl reader only sees **~10% of messages** — 90% of assistant messages in newer CodeBuddy versions (glm-5.2) lack `providerData.rawUsage` in `~/.codebuddy/projects/**/*.jsonl`. This causes massive undercounting.

Additionally, the extension-log path overlaps with the jsonl path using a different dedup key namespace, causing **3x inflation** when both are active.

## Root cause

1. **Traces dir not scanned** — `resolveCodebuddyProjectFiles` only scanned `projects/` and extension logs, never `traces/`
2. **90% of jsonl messages lack rawUsage** — newer CodeBuddy (glm-5.2) stops writing `providerData.rawUsage` on most assistant messages; token data only exists in traces
3. **Extension-log double-counting** — jsonl and extension-log use different dedup keys (`entry.uuid` vs `codebuddy:extension-log:agentId:sec`), so the same LLM round-trip is counted twice
4. **WorkBuddy rawUsage fallback missing** — older WorkBuddy sessions only write `providerData.model` without `rawUsage`; some carry the camelCase top-level `usage` field instead

## Fix

1. **`resolveCodebuddyProjectFiles`**: scan `traces/` dir, return `{path, kind}` objects (`jsonl | trace | log`)
2. **`parseCodebuddyIncremental`**: add trace JSON branch — read `modelInfo`, dedup by `traceId`, skip jsonl messages for traced sessions (`tracedSessionIds`) to avoid double-counting
3. **`resolveWorkbuddyProjectFiles`**: same `traces/` scan (WorkBuddy is the renamed successor of CodeBuddy, same trace schema)
4. **`parseWorkbuddyIncremental`**: same trace JSON branch
5. **WorkBuddy rawUsage fallback**: when `providerData.rawUsage` is absent, fall back to top-level `usage` field (camelCase convention)
[...]

PR #402, opened 2026-08-01
fix(codebuddy): count function_call usage, fix ~60% token undercount

HandSonic:
## Summary

The CodeBuddy jsonl parser only aggregates `type=="message" && role=="assistant"` records, but **every LLM round-trip that ends in a tool call is recorded as `type=="function_call"`** with full `providerData.rawUsage`. On a real install those records are **~93% of all round-trips** and carry **~14x** the assistant-message token volume — the parser silently drops them.

Verified on real `~/.codebuddy` data (109 projects, 2,245 jsonl files, 33k deduped LLM round-trips):

| Metric | Before (main) | After (this PR) | Ground truth |
|---|---|---|---|
| net input | 66.5M | **172.8M** | 172.2M |
| output | 5.3M | **9.6M** | 9.5M |
| cache read | 806M | **2.42B** | 2.37B |

(Residual <0.1% is incremental-scan timing — new sessions written between the two scans.)

The WorkBuddy reader already aggregates every usage-carrying record for exactly this reason (`rollout.js` WorkBuddy header comment: *"Usage lives on function_call records too"*). CodeBuddy uses the identical transcript format — the fix was simply never ported back.

## Root cause

```js
// rollout.js (main): drops 93% of LLM round-trips
if (!entry || entry.type !== "message" || entry.role !== "assistant") continue;
```

A single user "turn" in CodeBuddy is many LLM round-trips: model → tool call → tool result → model → … → final text. Only the final text reply is a `message` record; every intermediate round-trip is a `function_call` record with its own `providerData.rawUsage`.

## Fix

1. **Aggregate ANY record with `providerData.rawUsage`**, regardless of record type (mirrors the WorkBuddy reader).
2. **Dedup per round-trip via `providerData.messageId`** (the response-level id shared by the function_call/message pair of the same round-trip) before falling back to `entry.uuid`/`entry.id`. Per-record ids would count a round-trip once per record type.
3. **reasoning_tokens from `completion_tokens_details`** (prompt-side is always 0 on real data), subtracted from `completion_tokens` to avoid double-counting output.
4. **Cache split also reads `prompt_cache_hit_tokens` / `prompt_cache_write_tokens`** (DeepSeek-style mirrors present in real rawUsage); net input subtracts both cache reads and cache writes.

## Note on the trace-based approach (#399)

Closed #399 in favor of this. Real-data verification showed `trace.modelInfo` covers only **~64%** of a session's jsonl usage (17%–100% per session — one trace per agent workflow, while sessions accumulate many workflows + sub-agent round-trips), so traces cannot be the authoritative source. The "90% of messages lack rawUsage" observation from #399 was correct, but the usage isn't missing — it lives on `function_call` records, which this PR reads directly. No trace files needed.

## Tests

- `test/rollout-parser.test.js`: 229 pass / 2 fail — both pre-existing on main (`#141` codebuddy resolver fixture, `#154` kiro), unrelated to this change
- Real-data verification: parser output matches an independent jsonl-only ground-truth scan within 0.1%

## Scope

- [x] CLI (`src/lib/rollout.js` — `parseCodebuddyIncremental` only; WorkBuddy untouched)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved usage tracking for function-call and other response records.
  * Prevented duplicate usage entries by consistently recognizing shared response identifiers.
  * Improved token reporting, including cached reads, cached writes, reasoning tokens, and completion output.

### DIFF
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -8179,9 +8179,15 @@ async function parseCodebuddyIncremental({
       } else {
         let entry;
         try { entry = JSON.parse(line); } catch { continue; }
+        if (!entry || typeof entry !== "object") continue;
 
-        if (!entry || entry.type !== "message" || entry.role !== "assistant") continue;
-
+        // Usage is carried on ANY record with providerData.rawUsage — assistant
+        // messages AND function_call records. Each LLM round-trip (whether it
+        // ends in a text reply or a tool call) carries its own usage; on real
+        // installs function_call records are ~93% of round-trips and ~14x the
+        // assistant-message token volume, so filtering by record type drops
+        // the majority of usage. Aggregate them all; dedup per round-trip.
+        // (Same convention as the WorkBuddy reader — same transcript format.)
         const provider = entry.providerData;
         const rawUsage = provider && typeof provider === "object" ? provider.rawUsage : null;
         if (!rawUsage || typeof rawUsage !== "object") continue;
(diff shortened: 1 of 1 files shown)


xiufengsun (approved):
Reviewed exact head 4ea3347e8b5e6e39c4e96624c4b2b9a741bcdb8f against main f54ee17464e4f0052293165278998176cf87892c. Parser suite 231/231 and an independent function_call/cache/reasoning/dedup fixture pass; no blocking review threads remain.

PR #410, opened 2026-08-02
fix(mac): interpolate value/cost in pet limit reset and today-cost strings

fflake33:
## Summary

The desktop pet bubble on macOS showed the literal text `(value)` and `(cost)` instead of the actual usage-limit reset time and daily cost. `Strings.petLimitReset(_:)` and `Strings.petCostToday(_:)` accept a parameter but passed it to `t()` as the literal string `"(value)"` / `"(cost)"` — never as Swift string interpolation — so `t()` returned that text verbatim across all five locales.

This switches both helpers to Swift string interpolation `\\(value)` / `\_(cost)` in every locale column (en / zh-CN / zh-TW / ja / ko), matching the interpolation style already used elsewhere in `Strings.swift` (e.g. `"\(subject) limit reset 🎉"`). It also adds a native test that compiles `Strings.swift` with a harness, cycles each locale preference, and asserts the interpolated output for both functions across all five locales.

**Before** (zh-CN): `(value)后重置` and `今日 (cost)`
**After** (zh-CN): `8/12 05:09后重置` and `今日 $1.23`

## Scope

## Checklist

- [x] Commits follow conventional style (`fix:`)
- [x] No new user-facing strings (existing strings corrected, no `copy.csv` change needed)

## Verification

- `node --test test/pet-parity.test.js` — 16/16 pass (no cross-platform parity regression).
- The new `test/native-pet-limit-reset.test.js` compiles `Strings.swift` + a harness via `xcrun swiftc` and asserts all five locales. It is macOS-only and skips cleanly on other platforms / when `swiftc` is missing.

## Limitations

- **The macOS app was not built or run locally.** Verification is limited to the native test harness, which compiles the relevant Swift sources (`Strings.swift`, `NativeLocalization.swift`, `WidgetSnapshot.swift`, `MenuBarDisplayPreferences.swift`, `UsageLimits.swift`) directly with `swiftc` rather than building the full Xcode project. A full `xcodegen generate` + Xcode build was not performed, so visual confirmation of the pet bubble in the running app is still pending from a maintainer (or a local build on macOS).
- The fix is a two-line string-interpolation correction with no logic change, so the regression surface is intentionally narrow.

## Repro

Open the macOS menu-bar pet while a provider has an active usage limit. The reset line reads `(value)后重置` / `in (value)` instead of the real reset time, and the today-cost line reads `今日 (cost)` instead of the real cost.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected pet limit reset and daily pet cost messages to display the relevant values instead of placeholder text.

* **Tests**
  * Added macOS coverage for localized pet reset and cost messages in English, Simplified Chinese, Traditional Chinese, Japanese, and Korean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,9 @@ jobs:
           npm ci
           node --test test/openclaw-parser.test.js
 
+      - name: Test macOS native pet limit reset interpolation
+        run: node --test test/native-pet-limit-reset.test.js
+
       - name: Install xcodegen
         run: brew install xcodegen
 
--- a/TokenTrackerBar/TokenTrackerBar/Utilities/Strings.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Utilities/Strings.swift
@@ -212,10 +212,10 @@ enum Strings {
     static var petLimitLow: String { t("low", "余量偏低", "餘量偏低", "残りわずか", "얼마 안 남음") }
     static var petLimitSomeLeft: String { t("some left", "还有余量", "還有餘量", "残りあり", "남아 있음") }
     static func petLimitReset(_ value: String) -> String {
-        t("in (value)", "(value)后重置", "(value)後重置", "(value)でリセット", "(value) 후 초기화")
+        t("in \(value)", "\(value)后重置", "\(value)後重置", "\(value)でリセット", "\(value) 후 초기화")
     }
     static func petCostToday(_ cost: String) -> String {
-        t("(cost) today", "今日 (cost)", "今日 (cost)", "今日 (cost)", "오늘 (cost)")
+        t("\(cost) today", "今日 \(cost)", "今日 \(cost)", "今日 \(cost)", "오늘 \(cost)")
     }
 
 
--- a/test/native-pet-limit-reset.test.js
+++ b/test/native-pet-limit-reset.test.js
@@ -30,14 +30,11 @@ test("native pet reset strings interpolate the supplied reset value", (t) => {
     harnessPath,
     `import Foundation
 
-func fail(_ message: String) -> Never {
-    FileHandle.standardError.write(Data(message.utf8))
-    exit(1)
-}
+var failures: [String] = []
 
 func requireEqual(_ actual: String, _ expected: String, _ locale: String) {
     if actual != expected {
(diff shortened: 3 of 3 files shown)


fflake33:
Addressed both review findings in 6e28dd2.

**1. Restore preferences before calling `exit(1)` (Data Integrity)**

Verified the finding: `fail(_:)` called `exit(1)`, which terminates the process before the top-level `defer` could restore `NativeLocalization.preferenceKey` in both `UserDefaults.standard` and the shared App Group defaults. A failed assertion would leave the host environment mutated.

Fix: extracted the restoration into `restorePreferences()` and call it both from `defer` (normal exit path) and explicitly before `exit(1)` (failure path). This makes restoration unconditional regardless of how the process exits, rather than relying on whether `defer` unwinds before `exit`.

Verification: built a deliberately-failing harness (flipped one expectation to `WRONG`) and read `NativeLocalization.preferenceKey` back from both UserDefaults scopes before and after the failing binary ran. `BEFORE` and `AFTER` are identical, confirming preferences are restored on the failure path.

```
BEFORE: standard=<nil> shared=zh-CN
FAIL exit: 1 stderr: zh-CN: expected WRONG, got 8/12 05:09后重置
AFTER:  standard=<nil> shared=zh-CN
```

The normal (passing) path also still passes: `node --test test/native-pet-limit-reset.test.js` ✔.

**2. Run the Darwin-only regression test in macOS CI (P2)**

Verified: the Ubuntu job runs `npm test`, where this test skips (`process.platform !== "darwin"`); the macOS job only ran `test/openclaw-parser.test.js` and the Xcode `TokenTrackerBarTests` scheme, so this test never executed in CI.

Fix: added a step to the `macos-tests` job to run `node --test test/native-pet-limit-reset.test.js`. The `macos-26` runner ships with Xcode, so `xcrun --find swiftc` resolves and the test runs rather than skipping.

Note: the repo has several other `*-native.test.js` files (`codex-reset-bank-native`, `kiro-cli-native-contract`, `native-bridge-sync-feedback`, `serve-native-background-sync`, `windows-native-glass-selector`) that also only run on their respective platforms and are not currently invoked by the macOS/Windows CI jobs. That is a pre-existing coverage gap outside this PR's scope; this PR only ensures the test it introduces is covered.

PR #443, opened 2026-08-10
fix(macos): preserve floating panels when dashboard closes

paveezheng:
## Summary

Prevent closing the Dashboard from globally hiding the app, so Dynamic Island and desktop pet panels remain visible while the Dock icon still disappears correctly.

## Scope

## Checklist

- [x] macOS unit tests pass
- [x] PR description explains why

## Tests

- `xcodebuild -quiet -project TokenTrackerBar.xcodeproj -scheme TokenTrackerBarTests -configuration Debug -derivedDataPath /private/tmp/tokentracker-dock-tests test`

## Summary by CodeRabbit

* **New Features**
  * Improved Dashboard presentation, closing, and reopening behavior.
  * Dashboard content and state are preserved when the window closes.
  * Independent panels remain available after closing the Dashboard.

* **Bug Fixes**
  * Improved activation behavior when switching between menu-bar and Dashboard modes.
  * Prevented the Dashboard from reopening during app termination.
  * Improved termination handling when closing the Dashboard with Cmd+Q.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Models/DashboardPresentationPolicy.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Models/DashboardPresentationPolicy.swift
@@ -1,0 +1,75 @@
+import Foundation
+
+/// Pure state machine for the app's two presentations:
+/// a background menu-bar agent and a foreground Dashboard window.
+///
+/// Keeping this free of AppKit makes the Dock/close ordering testable without
+/// launching an application. `DashboardPresentationCoordinator` applies the
+/// returned commands to `NSApp` on the main actor.
+enum DashboardPresentationState: Equatable {
+    case menuBar
+    case dashboard
+    case dashboardClosing
+    case terminating
+}
+
+enum DashboardPresentationCommand: Equatable {
+    case useRegularActivationPolicy
+    case useAccessoryActivationPolicy
+    case showDashboardWindow
+}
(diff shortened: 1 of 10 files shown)


xiufengsun:
There is one blocking termination regression in the latest head (`5c5e3b50`).

`applicationShouldTerminate` always returns `.terminateCancel` for Command-Q after calling `closeDashboard()`, but `closeDashboard()` is a no-op when the Dashboard window does not exist or is already closed. In that state, Command-Q can leave TokenTracker running indefinitely in the background.

Please make `closeDashboard()` report whether it actually started closing a visible Dashboard window. Return `.terminateCancel` only when a close began; otherwise call `prepareForTermination()` and return `.terminateNow`. Please also add coverage for Command-Q when the Dashboard is absent/already closed.

The overall floating-panel lifecycle approach looks sound, and the current macOS tests pass locally (126/126), but they do not cover this integration path.

paveezheng:
Fixed in `3b140844`.

`Cmd+Q` now cancels termination only when it actually begins closing a visible Dashboard window. If the Dashboard is absent or already closed, TokenTracker prepares for termination and exits normally.

Added regression coverage for this path.

Validated with:

- `node --test test/macos-dashboard-window-space.test.js`
- Full macOS Xcode unit tests
- `npm test`

PR #469, opened 2026-08-15
feat(settings): add manual outbound proxy configuration

Roomen:
## Summary

Adds an explicit **Network** section in Settings where the CLI's outbound traffic can be pointed at a proxy — protocol (`http` / `https` / `socks5`), host and port.

Today a user behind a corporate or personal proxy depends on `HTTPS_PROXY` being exported in the right shell, or on the macOS system proxy being picked up. Neither is reachable from the menu bar app, where there is no shell to export into. This makes it configurable from the UI.

## Design

**Config** — `config.json` gains one top-level `proxy` key:

```json
"proxy": { "mode": "manual", "protocol": "socks5", "host": "127.0.0.1", "port": 1080 }
```

`mode` is `system` (default) / `manual` / `off`. A missing key, or any value that cannot be interpreted (out-of-range port, host with a protocol prefix, unknown mode), falls back to `system` without throwing — **existing installs behave exactly as before**.

**Resolution order** — `manual` > environment variables > system proxy. `off` forces a direct connection and ignores both.

**socks5 needs no new dependency.** undici's `ProxyAgent` has handled `socks5://` natively since v8.0.0 (`lib/dispatcher/socks5-proxy-agent.js`), and this repo already depends on `undici ^8.7.0`. All three schemes go through the same code path.

## Two decisions worth calling out

**1. Manual mode never injects `HTTPS_PROXY`.** The obvious implementation is to reuse the existing `resolveSystemProxyEnv` → relaunch path. That turns out to be wrong: the injected variable lands in the child's `process.env`, so after the user switches back to `system`, `pickProxyUrl(env)` reads the *manual* URL back and keeps using it — the mode change silently does nothing, and can leave the user with no connectivity. Manual mode is therefore applied only through the undici dispatcher, which covers it completely. `system` / `off` relaunch behavior is untouched.

**2. A failed manual proxy is fail-closed.** If the dispatcher cannot be constructed, outbound traffic is blocked and the error is surfaced in Settings, rather than falling back to a direct connection. Silently going direct would mean the user believes they are proxied while their real IP reaches the cloud endpoint. This applies to `manual` only — `system` / `env` keep their existing tolerant, silent behavior.

Runtime switches also `close()` the dispatcher they replace, since `setGlobalDispatcher` only swaps the global symbol and does not recycle the previous agent.

## Local API

- `GET /functions/tokentracker-proxy-config` — normalized config + which source is effective + last apply error. Unauthenticated, matching `tokentracker-cloud-sync-pref`; the settings page also uses it to probe availability.
- `POST /functions/tokentracker-proxy-config` — guarded by `isAuthorizedLocalMutation`. Invalid payloads are rejected before anything is written; the write is read-modify-write and atomic (temp + rename), so other keys in `config.json` survive.
- `POST /functions/tokentracker-proxy-test` — connectivity check against the existing `dashboardUrl` with a 5s cap. It builds a throwaway dispatcher and passes it to `fetch({ dispatcher })`; it never calls `setGlobalDispatcher` and never persists the values under test.

The Network section only renders when `GET .../proxy-config` succeeds, so the hosted dashboard at www.tokentracker.cc is unaffected — no section, no console noise, no retry loop.

## Testing

- `npm test` — 2198 pass / 0 fail / 2 skip
- `npm run validate:copy`, `validate:ui-hardcode`, `validate:guardrails` — pass
- `npm run dashboard:build` — pass
- `npm --prefix dashboard test -- --run` — 85 files / 502 tests pass

New tests cover config normalization (valid, dirty and legacy-missing shapes), both endpoints including the 401 path and config-key preservation, the manual→system transition, and fail-closed installation. They were mutation-verified: reintroducing the env-injection bug, dropping the fail-closed install, removing the auth guard, clobbering other config keys, and accepting out-of-range ports each make the suite fail.

All user-facing strings go through `copy.csv` with the five locale files updated.

## Notes for review

- **The connectivity test is the easiest thing to cut** if you'd prefer a narrower PR. It's the only part beyond "protocol/host/port in Settings". It binds loopback-only and requires the local mutation token, but it does give an authenticated local caller a TCP reachability probe for arbitrary `host:port`. Happy to drop it.
- **Pre-existing, deliberately not touched:** `package.json` declares `engines.node: ">=20"` while undici 8.x declares `>=22.19.0`. On a runtime where `new ProxyAgent('socks5://…')` throws, this change fails closed with a visible error rather than silently connecting direct — but the underlying version mismatch is worth a separate look.
- **Also pre-existing:** `sync` gets no proxy when only a macOS system proxy is set and no env vars are exported, because only `serve` goes through the relaunch path. Manual mode incidentally fixes this for `sync`, since the config is read directly. Not otherwise changed.
- Proxy authentication (`user:pass`) is intentionally out of scope; the schema leaves room for it.

## Summary by CodeRabbit

* **New Features**
  * Added network settings for system, manual, or disabled proxy modes.
  * Added proxy host, port, protocol configuration, validation, connection testing, and effective proxy details.
  * Persisted proxy settings for dashboard and runtime connections.
  * Added localized network settings in German, Japanese, Korean, Simplified Chinese, and Traditional Chinese.

* **Bug Fixes**
  * Improved handling of invalid proxy settings and application failures, including safer traffic blocking.
  * Restored direct connections when proxy settings are disabled or unavailable.
  * Improved request timeouts, redirects, and temporary-file cleanup.


xiufengsun (changes requested):
Quality is well above average for this codebase — the socks5 support is real (undici 8.9's socks5-proxy-agent, zero new deps, so the --ignore-scripts desktop bundles are unaffected), loopback traffic is correctly untouched, the mutation endpoint follows isAuthorizedLocalMutation, and the fail-closed integration test actually binds a real server. Two things block the merge, both variants of the failure mode this PR itself calls unacceptable — UI says proxied, traffic goes direct:

1. **Runtime switch manual → system silently goes direct while the UI claims the system proxy is active.** In `applyUndiciProxyIfNeeded`, `const proxyUrl = manual ? buildProxyUrl(normalized) : pickProxyUrl(env)` — a serve process that started in manual mode never went through the relaunch env injection, so on switching to `system`, `pickProxyUrl(env)` is null and `installDirectAgent` runs. Meanwhile `resolveEffectiveProxySource` shells out to `scutil`, finds a macOS system proxy, and the UI renders "Currently using the system proxy." That's exactly the target user of this PR (menu bar app, no shell env). Two of the PR's own tests assert this behavior as expected, so CI won't catch it. Either resolve the system proxy URL (you already parse scutil output) and build the dispatcher from it, or make the POST response report what was actually installed plus a restart-required flag.

2. **Cursor account usage fetches bypass the manual proxy entirely.** `src/lib/cursor-config.js:149` and `:226` use `node:https` directly to www.cursor.com; the undici global dispatcher doesn't govern `https.request`. env/system modes cover these today via `NODE_USE_ENV_PROXY` on the bundled Node 22.22.2 — manual mode, which deliberately avoids env injection, does not. So manual's coverage is strictly weaker than the modes it replaces. Convert those two calls to `fetch` or pass an explicit agent.

Worth fixing in the same pass:

- `writeConfig` in local-api.js drops config.json from 0600 to 0644 on every save — the file holds `deviceToken`, and `machine-id.js` deliberately chmods it 0600. `src/lib/fs.js` already exports `writeFileAtomic` + `chmod600IfPossible`; reuse them (and clean up the tmp file on failure).
- A corrupted manual config block silently degrades to system/direct — fail-open, contradicting the PR's own fail-closed rationale. Surface it via the existing `applyError` field at minimum.
- `NetworkSection.jsx` sets `saveState("error")` but only ever renders `"saved"`, and renders no feedback at all in System/Off modes — a failed save is invisible. Needs a copy.csv key for the error state.

Minor, non-blocking: `resolveEffectiveProxySource` runs `scutil` via spawnSync inside request handlers (cache it or go async), and `use-proxy-settings.js` should gate on `isLocalDashboardHost()` like cloud-sync-prefs.ts does rather than relying on the JSON parse of the SPA fallback failing.

Ping me once the two blockers are addressed and I'll re-review promptly — I want this feature in.

Roomen:
Thanks for the detailed review — both blockers are fixed, plus everything under "worth fixing" and the two minor items. Pushed as three commits on top of the original (`7149615e`, `597dc8c9`, `d59dda0f`).

## Blocker 1 — manual → system silently went direct

Took your first option: **resolve the system proxy URL and build the dispatcher from it**, rather than a restart-required flag. The reason for preferring it is that a flag leaves two independent resolution paths that can drift again; this makes them one.

`resolveEffectiveProxyUrl()` is now the single resolver, consumed by all three call sites — `applyUndiciProxyIfNeeded` (dispatcher), `resolveEffectiveProxySource` (UI label), and `resolveSystemProxyEnv` (relaunch). System-mode order is `process env` → `scutil`. The dispatcher and the "Currently using the system proxy" label can no longer disagree, because they are the same function call.

You were right that the PR asserted the bug as expected behavior. It was two tests plus one more you didn't name:

- `test/proxy-env.test.js` — `applyUndiciProxyIfNeeded is a no-op when no proxy env var is set`
- `test/proxy-env.test.js` — `switching from manual to system does not keep using the manual proxy` (asserted a `DirectAgent`)
- `test/local-api-proxy-config.test.js` — `POST system after manual` asserted `afterSystem instanceof ProxyAgent === false`, which on a machine that *has* a system proxy also encoded the bug

All three now assert the correct behavior against an injected `commandRunner`. Mutation-verified independently of the implementing pass: reverting the line to `manual ? buildProxyUrl(normalized) : pickProxyUrl(env)` fails with `actual: null, expected: { ok: true, proxyUrl: 'http://10.0.0.9:9' }`.

Side effect worth noting: this also fixes the "`sync` gets no proxy when only a macOS system proxy is set" gap called out in the original PR description, since `sync` no longer depends on having been relaunched with injected env.

## Blocker 2 — Cursor usage fetches bypassed the manual proxy

Converted both `node:https` call sites (`fetchCursorUsageCsv`, `fetchUrlRaw`) to injectable `fetch`; `require("node:https")` is gone from the file. Headers, the one-hop 301/302/308 follow, the missing-`Location` error, the 401/403 message and the `Cursor API request timed out` string are all preserved byte-for-byte — `AbortSignal.timeout`'s native `TimeoutError`/`AbortError` is remapped so the user-visible message doesn't change.

`fetchUrlRaw` deliberately keeps `redirect: "manual"` rather than `"follow"`: the old code threw `Cursor API returned ${status} from ${urlStr}` on a *second* hop, and auto-follow would have turned that into a silent success (and may drop the explicit `Cookie` on a cross-host hop).

Verified at runtime rather than by inspection — with a fail-closed undici global dispatcher installed:

```
fetchCursorUsageCsv({ cookie: 'x' })
  → caught: fetch failed | cause: BLOCKED_BY_DISPATCHER
```

i.e. the CSV path is now governed by the global dispatcher, which is exactly what manual mode installs. `src/` was swept for other dispatcher-bypassing outbound calls: none remain that talk to the public internet (the leftovers are `serve.js` / `browser-auth.js` inbound servers, the Antigravity `127.0.0.1` probe in `usage-limits.js`, and the Qoder unix-socket RPC).

## Worth-fixing items

- **`writeConfig` 0600** — now uses `writeFileAtomic` + `chmod600IfPossible` from `src/lib/fs.js`, and `writeFileAtomic` gained tmp-file cleanup on failure. Confirmed your premise first: `machine-id.js:223` does chmod 0600, and the old local `writeFileSync` + `rename` landed at 0644 under the default umask. A write failure now returns 500 instead of reporting success.
- **Corrupted manual config fail-open** — `applyUndiciProxyIfNeeded` now detects "raw config declared `manual` but normalization rejected it" and routes it through `failManual()`: fail-closed dispatcher installed, reason surfaced via `applyError` (`invalid port`, etc.). `resolveEffectiveProxyUrl` returns `source: "none"` for that case so the UI never claims a system proxy is active while apply is fail-closing. Same treatment when `mode === "manual"` but `buildProxyUrl` returns null (e.g. a malformed colon host) — no silent direct connection.
  - Related: if the fail-closed dispatcher itself cannot be installed, `applyUndiciProxyIfNeeded` now returns `unprotected: true` and `bin/tracker.js` aborts with `process.exit(1)` rather than continuing on the default global dispatcher.
- **`NetworkSection.jsx` invisible failures** — saved/error feedback moved outside the manual-mode branch, so System/Off saves report too. Errors render as `role="alert"` carrying the backend reason, via two new `copy.csv` keys (`settings.network.save_error`, `settings.network.save_error_unprotected`) with all five locales filled in.
- **POST now reports reality** — the response's `ok` is derived from the apply result (HTTP stays 200 when the write succeeded but apply failed — "saved but not in effect"), and carries `unprotected`. The hook treats `200 + ok:false` as a save failure so it surfaces in the UI.

## Minor items

- **`scutil` spawn** — 30s TTL cache (negative results included) shared by the apply path and the UI path; the POST handler invalidates it so a save is never followed by a stale `effective`. One correction to the finding: `resolveEffectiveProxySource` was not running on *every* API request — its only production caller is `toResponse` on `/proxy-config` GET/POST. The cache is still worth having now that apply uses the same probe.
- **`use-proxy-settings.js`** — gated on `isLocalDashboardHost()`; no request is made on a public deploy. Verified this doesn't break the native apps first: macOS loads `localhost`/`127.0.0.1` (`DashboardWindowController.swift:444`) and Windows uses `http://127.0.0.1:{Port}` (`ServerManager.cs:38`) — the helper ignores the port, so the dynamic-port Windows case is covered.
- **CodeRabbit i18n** — de `apply_error` switched to the informal imperative, ko `error.host` space removed, zh-TW `mode.label` → `對外代理` and `互聯網` → `網際網路`.
- **CodeRabbit hardening** — `normalizeHost` now also rejects `@ ? # \` and internal whitespace; `buildProxyUrl` only brackets a host when `net.isIPv6()` says so and returns `null` for other colon-bearing hosts; `runProxyConnectivityTest` defaults to `undici.fetch`; the real-`ProxyAgent` test closes its dispatcher and resets module state.
- **CodeQL 227** (`Outbound network request depends on file data`) — investigated rather than suppressed. The tainted value is the proxy hop the user explicitly configured, and it reaches `createProxyDispatcher` only via `parseProxyPayload` or `normalizeProxyConfig` + `buildProxyUrl` (manual) / `pickProxyUrl` + `parseMacProxyOutput` (system). No unvalidated path exists, so behavior is unchanged and the reasoning is documented in a comment at the call site. Happy to add a suppression instead if you'd prefer the alert cleared.

## Verification

```
npm test                      2222 tests / 2220 pass / 0 fail / 2 skip
npm run validate:copy         ok (1458 entries, 1156 keys used)
npm run validate:ui-hardcode  ok
npm run validate:guardrails   ok
npm run validate:versions     ok (v0.88.8, unchanged)
npm run dashboard:build       ok
```

Every fix was mutation-verified (revert the fix, confirm the new test goes red, restore).

One thing to flag that is **not** from this PR: `npm --prefix dashboard test -- --run` currently reports 107 failing tests across 17 files (`ActivityHeatmap`, `LeaderboardPage`, `SessionsPage`, `SkillsPage`, `use-limits-display-prefs`, `cloud-sync-prefs`, …). I checked out `origin/main` into a clean worktree and got **the same 107 failures**, so this branch neither causes nor fixes them — it adds 2 files / 16 tests, all passing. Worth a separate look if it also reproduces on your machine.

Nothing under `src/` or `dashboard/` was version-bumped; per the release rules this will need the usual npm + DMG + Windows + Linux release when it lands, but I left that to you.

Roomen:
Follow-up on the second review round (`696e11a2`).

**CodeRabbit's redirect finding is valid and is now fixed.** The `Location` header of a 301/302/308 from the CSV endpoint was passed straight to `fetchUrlRaw`, which then forwards the session cookie to it — an off-origin redirect would have disclosed the cookie. This was pre-existing behavior (the old `node:https` version did the same), but the `fetch` conversion put it back in scope, so it should be fixed here rather than deferred.

`resolveCursorRedirect()` now resolves the header against `CURSOR_CSV_URL` and requires `https:` on `cursor.com` / `www.cursor.com`, otherwise throwing `Cursor API redirect to an untrusted origin`. Resolving also fixes a *relative* `Location`, which bare `fetch(location)` rejects — that was listed as a residual risk in the previous round and is now closed.

Covered by two new tests: relative `Location` resolves to `https://cursor.com/api/moved-csv`, and four hostile targets (`evil.example.com`, the `cursor.com.evil.example.com` suffix trick, an `http:` downgrade, and a protocol-relative `//evil.example.com`) are each rejected with **no second request issued**. Mutation-verified: deleting the protocol/host check turns that test red.

**On the three new CodeQL alerts** — all three are the same design-intended data flow rather than new defects, so I've left the behavior alone and would rather you decide how to close them out:

- `229` / `230` (`src/lib/cursor-config.js` — outbound request depends on file data): the "file data" is the Cursor session cookie read from Cursor's own local storage, which is the entire point of the account-usage fetch. The URL itself is now a compile-time constant plus an allowlisted redirect target, so the redirect fix above narrows this as far as it can go without dropping the feature.
- `231` (`src/lib/fs.js:14` — network data written to file): this is `POST /functions/tokentracker-proxy-config` persisting the proxy settings. The payload is guarded by `isAuthorizedLocalMutation` and fully validated by `parseProxyPayload` (mode/protocol enums, hostname charset, port range) before it reaches `writeFileAtomic`, and the file is chmod'd 0600. Flagging it is expected for any settings endpoint that writes to disk.

Happy to add dismissals or a query filter if you'd prefer the alerts cleared rather than argued.

`npm test` is 2224 tests / 2222 pass / 0 fail / 2 skip; guardrails clean. Also confirmed the branch is still current with `main` — no rebase needed.

xiufengsun:
Re-reviewed all four commits. Both blockers are genuinely fixed:

- **manual → system**: converging on `resolveEffectiveProxyUrl()` as the single resolver was the right call over a restart-required flag. I scanned `src/` — no second resolution path remains, and the switch test asserts the scutil-derived proxy gets installed rather than the leftover manual URL, with an end-to-end repro through the real undici global dispatcher.
- **redirect cookie disclosure**: the exact-Set hostname check (`cursor.com` / `www.cursor.com` after WHATWG URL normalization) closes the off-origin cases including `cursor.com.evil.example.com` and userinfo tricks, single-hop only, and non-3xx behavior is byte-for-byte unchanged. Since the old `node:https` version had the same hole, this is a net security win over main.

Also verified: no remaining public-network call in `src/` bypasses the global dispatcher (the `127.0.0.1` Antigravity probe correctly stays direct), proxy credentials never reach logs or API responses, and all 30 new copy keys are in `copy.csv` across 5 locales. I ran the dashboard suite against your head locally: 86 files / 513 tests, all green.

Two small things before merge:

1. **The mode hint overpromises.** "…reaches the internet for cloud sync **and updates**" — the desktop updaters (macOS `URLSession.shared`, Windows `HttpClient`) only honor the *system* proxy, never this manual setting. The core audience for manual mode is exactly the user whose system proxy isn't set, and their update check will still fail. Narrow the hint to cloud sync (or state that updates follow the system proxy), across all 5 locales.
2. **Contradictory UI in the fail-closed state.** When a corrupt manual config blocks traffic, the red alert says outbound traffic is blocked while the card subtitle simultaneously renders "Currently connecting directly." The effective state needs a distinct blocked representation.

One correction: the claim that origin/main has 107 dashboard test failures across 17 files doesn't reproduce here — current main runs 84 files / 497 tests, all passing. The real gap is that CI only runs a single dashboard test file (`use-limits-display-prefs.test.js`), so your three new test files never execute in CI. That's a CI change I'll make separately; no action needed from you, but please don't skip dashboard tests on that basis in future PRs.

Minor notes, no action required: zh-TW `save_error_unprotected` still says 「出網」 while the mode label was changed to 「對外代理」— worth unifying while you're in the locales; manual mode not supporting authenticated proxies (`@` rejected) is fine as a known limitation. I'll dismiss CodeQL 227 on the Security page — your reasoning on the data path holds.

Fix the two items above and this is ready to merge.

Roomen:
Both items fixed in `8e9f3347`, kept minimal.

**1. Mode hint no longer promises updates.** You're right that the updaters (`URLSession.shared` on macOS, `HttpClient` on Windows) never see this setting, and that the user manual mode exists for is exactly the one whose system proxy is unset. Took the explicit option rather than just dropping the word, since silence would leave the same wrong assumption:

> Choose how Token Tracker reaches the internet for cloud sync. App updates always follow the system proxy.

Updated across all 5 locales.

**2. Fail-closed now has its own effective state.** `resolveEffectiveProxyUrl()` returns `source: "blocked"` instead of `"none"` for the two cases that fail-close — a persisted `manual` block that fails normalization, and a `manual` config whose URL can't be built (e.g. a malformed colon host). New copy key `settings.network.effective.blocked` ("Outbound traffic is blocked: the manual proxy could not be applied.") in all 5 locales, and `NetworkSection`'s subtitle renders it. The card can no longer say "connecting directly" while the alert says traffic is blocked.

Two tests, both mutation-verified: reverting either source to `"none"` turns `fail-closed manual configs report effective 'blocked', never 'none'` red, and removing the subtitle branch turns the UI test red. I also tightened the existing `GET proxy-config falls back to system for truly invalid persisted proxy values` case from `notEqual(effective, "manual")` to `equal(effective, "blocked")`, so the contract is asserted rather than merely not-contradicted.

**3. Minor, done while in the locales:** zh-TW `save_error_unprotected` now uses 對外連線 to match the 對外代理 label.

---

**On the dashboard test count — you're right and I was wrong.** I dug into why my machine disagrees: the failures are `TypeError: Cannot read properties of undefined (reading 'clear')` on `localStorage`, i.e. the jsdom environment isn't being applied at all. This box is on Node v26.5.0 against vitest 2.1.9 / jsdom 24, and I have no older Node installed to compare. So it's a local toolchain mismatch producing a uniform false failure, not anything about `main` — please disregard that paragraph in my earlier comment.

Noted on not skipping dashboard tests on that basis, and thanks for catching that CI only runs `use-limits-display-prefs.test.js`. That does mean my `NetworkSection.test.jsx` / `use-proxy-settings.test.js` / `cursor-config` dashboard-side additions are unverified by CI until your change lands — worth weighing when you review them.

```
npm test                      2225 tests / 2223 pass / 0 fail / 2 skip
npm run validate:copy         ok (1459 entries, 1157 keys used)
npm run validate:ui-hardcode  ok
npm run validate:guardrails   ok
npm run validate:versions     ok (v0.88.8)
npm run dashboard:build       ok
```

Branch is still current with `main`, no rebase needed. Version deliberately unbumped — happy to bump and cut the npm + DMG + Windows + Linux release on merge if you'd rather I do that.

PR #510, opened 2026-08-24
fix(linux): default to the WebKitGTK compatibility renderer

csmashe:


csmashe:
Closing this — I opened it before actually testing the fix, and the reproduction claim in the description was not something I verified. Will re-open once it's confirmed on real hardware.

PR #511, opened 2026-08-24
feat: show exact Codex subagent usage by thread

kymonr:
## Why

Codex v2 subagents are stored as separate rollout files. The Sessions page currently presents them as unrelated sessions and estimates subagent usage from spawn calls, so a Dynamic Workflow cannot answer how many tokens each child agent and model actually consumed.

## What changed

- preserve local Codex lineage and agent metadata from the first `session_meta` row (forked rollouts may replay parent metadata later)
- build fail-closed parent/child trees and exact recursive token/cost views without changing each session's authoritative `total_tokens`
- replace Codex spawn-call estimates with observed child-session attribution
- fold child sessions under their root session, with agent nickname/role and a combined `Σ` usage value
- add per-root model breakdown/filter chips showing model, child count, and token total
- honor `CODEX_HOME` when resolving the process home
- add English, German, Japanese, Korean, Simplified Chinese, and Traditional Chinese UI copy

## Safety and privacy

The implementation does not read or retain prompts, responses, command output, or diff content. Lineage IDs, agent names/roles, titles, and project paths remain local-only and are stripped from cloud/CSV summaries. Conflicting, cyclic, duplicate, or orphan links are not aggregated.

## Validation

- `node --test test/session-analytics-codex-subagents.test.js test/session-analytics.test.js` — 30/30 passed
- `npm run validate:copy`
- `npm run validate:locale`
- `npm --prefix dashboard run typecheck`

Validated against a local real-world corpus containing 8,128 sessions and 2,094 folded subagent rows. A 24-child Dynamic Workflow was attributed by observed child sessions as:

- `gpt-5.6-sol`: 10 children / 38.4M tokens
- `gpt-5.6-luna`: 13 children / 33M tokens
- `gpt-5.3-codex-spark`: 1 child / 5.5M tokens
- combined: 76.8M tokens

## Summary by CodeRabbit

* **New Features**
  * Added subagent thread grouping with expandable nested session rows.
  * Added filtering by subagent model, result counts, and combined token summaries.
  * Improved session search with agent metadata.
  * Added clearer usage details for parent sessions and subagents.

* **Localization**
  * Added translations for new session controls and usage details in German, Japanese, Korean, Simplified Chinese, and Traditional Chinese.

* **Bug Fixes**
  * Improved session lineage and usage tracking across resumed and nested sessions.
  * Added safeguards for invalid, conflicting, orphaned, or circular session relationships.


xiufengsun (changes requested):
Thanks for the substantial work here. I re-reviewed exact head `8f3c605b` against the current main after today's merges. The backend lineage tests pass 30/30 and the integration is conflict-free, but three blockers remain before merge:

1. **Custom `CODEX_HOME` loses Codex thread titles.** `src/lib/session-analytics.js:1052-1056` derives `session_index.jsonl` only when the rollout path contains a literal `.codex` segment. The new `providerRoots()` support accepts an arbitrary custom directory, so sessions are discovered there but their title index is never found. Derive the index from the discovered Codex root (including `sessions/` / `archived_sessions/`) and add a regression using a custom directory name.

2. **The exported client contract is stale.** `dashboard/src/lib/sessions-api.ts:10-30` does not declare the new thread/lineage and combined-usage fields consumed by `SessionsPage.jsx` (`thread_kind`, parent/root hashes, agent metadata, own/subagent/combined totals and counts). Please update `SessionRow` so TypeScript consumers match the server payload.

3. **The new UI behavior has no regression coverage.** This PR adds roughly 210 lines of grouping, nested rendering, model filtering, combined totals, and standalone-child behavior to `SessionsPage.jsx`, but only backend tests were added. Please add focused page/component tests covering root + child + grandchild folding, model filtering, and a child whose root is filtered out.

The current red `test + validate + build` check is from the existing `fs-lock` timing tests rather than this feature, so rerun CI after the code changes instead of treating that failure as a feature blocker.

kymonr:
## Performance follow-up: session analytics cache TTL is too short for large Codex archives

While validating this PR against a production-sized local corpus, I found that ordinary Sessions page loads can take 50-60+ seconds due to aggressive cache invalidation in `buildSessionAnalyticsInternal`.

### Measured environment

- Tracked rollout files: **8,473**
- `archived_sessions`: ~54 GB / 8,043 files
- `sessions`: ~1.2 GB / 404 files
- Server: TokenTracker 0.93.2 embedded server, Windows 11

### Timings

| Scenario | Latency |
|---|---|
| Cache hit (within TTL) | 0.3-0.8s |
| Incremental rebuild after 5-min TTL expiry | ~63s |
| Forced refresh (`refresh=true`) | ~53s |

### Root cause

`buildSessionAnalyticsInternal` uses `cacheTtlMs = 5 * 60_000`. After 5 minutes, even an ordinary page load triggers discovery + stat + incremental scan of all tracked files. For large archives, this means users wait nearly a minute just opening the Sessions tab.

### Proposed change

Raise default TTL from 5 minutes to 30 minutes, and allow environment override:

```js
const SESSION_ANALYTICS_CACHE_TTL_MS = Number(
  process.env.TOKENTRACKER_SESSION_CACHE_TTL_MS || (30 * 60_000),
);

async function buildSessionAnalyticsInternal({ home = os.homedir(), force = false, cacheTtlMs = SESSION_ANALYTICS_CACHE_TTL_MS } = {}) {
```

This does not change parsing logic, token attribution, privacy behavior, or schema. It only extends how long valid metadata is trusted before re-checking file signatures. The existing signature check still catches changed files immediately within the window.

I have validated this locally: after applying the same change to the installed copy, cached responses remain fast, forced refreshes still work correctly via `force:true`, and no data accuracy issues were observed.

Happy to submit this as a separate small PR if you prefer.

xiufengsun (changes requested):
Thanks for addressing all three items from the previous review at exact head `2b6819c47268aa4e947de1c26670949b3f84db75`; custom CODEX_HOME titles, the SessionRow contract, and the UI coverage are no longer blockers. Two new lineage-integrity blockers remain:

1. In `src/lib/session-analytics.js:1547-1561`, a row whose own parent is missing is marked `orphaned_subagent`, but it remains in `byId` and is still allowed to parent other rows. Repro: A points to a missing parent and B points to A; A is then presented as an orphan/root with B aggregated into its descendant totals even though the branch has no valid root. Exclude orphaned/conflicted nodes from parenting (or invalidate the whole branch) and add this chain as a regression.

2. Duplicate session IDs are marked conflicted at lines 1527-1531, but aggregation is memoized only by `session_id` at lines 1565-1583 and then applied to every original row at lines 1602-1615. With two rows sharing ID A plus one child of A, the second duplicate reuses the first row's memoized descendant aggregate, so both conflicting rows report the same child usage. Isolate/drop duplicate IDs before graph construction and add a duplicate-ID regression that proves no aggregate is reused.

Conclusion: NO-MERGE. Please push a new head with both graph-integrity cases covered.

xiufengsun:
Merged.

The thing I checked hardest is the one that would be invisible until the numbers were already wrong: whether the combined view can leak into the ledger. It can't, and your test says so directly —

```js
assert.equal(result.sessions.reduce((sum, row) => sum + row.total_tokens, 0), 170);
```

with `get("root").combined_total_tokens` also 170. Summing `total_tokens` across rows still equals the global ledger, and `combined_total_tokens` is a derived field beside it rather than written back over it. The cycle and orphan cases assert the same thing, and `local-only identity fields never reach the cloud or CSV surface` covers the lineage ids. 31/31 green on a fresh checkout here.

Dropping the spawn-call estimate is right on its own. A spawn is a control-plane event; billing it as usage was inventing tokens, and observed child rows are the real thing.

---

**On the cache TTL follow-up: the measurement is valuable, the fix isn't the one I want.**

Raising the TTL to 30 minutes doesn't remove the 50-60s wait — it moves it from every 5 minutes to every 30, and whoever lands on the expired tick still sits through the whole thing. It also means a session you just finished can stay invisible for half an hour, on a page whose whole point is showing what you just ran.

Your numbers actually locate the real cost. Past the TTL the code already short-circuits: `filesSignature` is compared, and an unchanged signature returns the cached sidecar without reparsing anything. So the minute is being spent in `discoverSessionFiles` plus statting 8,473 paths — directory traversal over a 54GB archive, not parsing. That's why the TTL bump appears to work: it skips the traversal, it doesn't make it cheaper.

The fix belongs in discovery. `archived_sessions` is append-only and its dated subtrees never change again, so their directory mtime can prune whole subtrees before any file is statted, and the discovery result itself can be cached in the meta with the same signature discipline. That keeps a fresh session visible within seconds while making the steady-state path close to free.

A separate PR on that would be welcome — please include the same before/after timings from your 8,473-file corpus, since that environment is what makes it reviewable.

PR #512, opened 2026-08-24
fix(linux): default to the WebKitGTK compatibility renderer

csmashe:
## Problem

WebKitGTK renders through DMA-BUF by default. On some Wayland setups — most reliably NVIDIA's proprietary driver — that path either paints a permanently blank webview or loses the Wayland connection outright, aborting before the dashboard is ever shown:

```
Gdk-Message: Error 71 (Protocol error) dispatching to Wayland display.
```

The process exits non-zero, so on a desktop launch there is no window and no error the user ever sees. Everything else looks healthy on the way down — the bundled Node server starts and answers its readiness probe, and nothing lands in `server.log`.

## Evidence

Tested against an **unmodified** build of current `main` (`b7ce3aec`), so the variable is isolated as the cause independently of this patch. Fedora 44 / GNOME (Wayland), GeForce GTX 1070 on NVIDIA 580.173.02, WebKitGTK 4.1:

| `WEBKIT_DISABLE_DMABUF_RENDERER` | Result |
| --- | --- |
| unset — upstream's current default | `exit 1`, `Error 71 (Protocol error)` |
| `1` | runs normally |
| `0` | `exit 1`, `Error 71 (Protocol error)` |
| empty string | runs normally |

The last two rows confirm WebKitGTK reads the variable as *set and not `0`*.

## Fix

Set the variable to `1` before GTK initializes, and leave any user-supplied value untouched — which, given the semantics above, means `WEBKIT_DISABLE_DMABUF_RENDERER=0` still genuinely opts back into the accelerated renderer:

```bash
WEBKIT_DISABLE_DMABUF_RENDERER=0 tokentracker-linux
```

`configure_webkit_runtime()` is the first statement in `main` for two reasons: the variable is only read when GTK/WebKit initializes, and `std::env::set_var` is sound only while the process is still single-threaded.

## Separate bug found while testing (not fixed here)

The abort never reaches `stop_server()`, so each crashed launch **orphans the bundled Node server**. I accumulated four of them across test runs, holding ports 17680, 45859, 40603 and 40555. Because the orphan keeps 17680, the next launch hits the `pick_available_port` fallback and lands on a random port — and per `server.rs`'s own comment, OAuth sign-in then silently fails until the orphan is killed. Documented in the README here; happy to file it separately if you want it fixed.

## Scope

- Linux-only. No `src/` or `dashboard/` changes, so the npm CLI, macOS and Windows are untouched.
- No version bump — `scripts/version-files.cjs` drives nine files off `package.json`, so that's yours to make on the next release.
- README gains a troubleshooting section, since the symptom is invisible in the logs and users have nothing to search for otherwise.
- Test in `test/linux-health-monitor.test.js` asserts the call is the first thing `main` does and that an explicit value stays authoritative.

## Verification

- `cargo check --locked --all-targets` — clean, no new warnings
- `cargo test --locked` — 12 + 14 passing
- `node --test test/linux-health-monitor.test.js test/linux-bundle.test.js test/linux-test-safety.test.js test/linux-client-workflow.test.js` — 21 passing
- `npm run validate:guardrails`, `npm run validate:versions`, `node --test test/architecture-guardrails.test.js` — clean
- Both binaries built from this tree and installed side by side; the running process was confirmed via `/proc/<pid>/exe` hashes so the before/after runs could not be mixed up

`cargo fmt --check` and `clippy` could not be run locally (this box has cargo without the rustfmt/clippy components), so CI is the authority on those two.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Linux startup reliability by disabling problematic WebKitGTK DMA-BUF rendering by default on affected Wayland setups.
  - Preserved explicit user configuration for re-enabling hardware acceleration.
  - Added guidance for resolving blank or immediately closing Linux windows and cleaning up stalled processes.

- **Tests**
  - Added coverage to verify renderer configuration is applied before application startup.

### DIFF
--- a/TokenTrackerLinux/README.md
+++ b/TokenTrackerLinux/README.md
@@ -91,6 +91,37 @@ Until the extension is installed, closing the window hides the app with no way t
 get it back from the tray — quit it from the launcher or with `pkill
 tokentracker-linux`.
 
+### The window is blank, or the app exits immediately
+
+WebKitGTK renders through DMA-BUF by default. On some Wayland setups — most
+reliably NVIDIA's proprietary driver — that path either paints a permanently
+blank webview or loses the Wayland connection outright, exiting non-zero with:
+
+```
+Gdk-Message: Error 71 (Protocol error) dispatching to Wayland display.
+```
+
+The client therefore sets `WEBKIT_DISABLE_DMABUF_RENDERER=1` before starting
+GTK. To retry the accelerated renderer, set it explicitly — an explicit value is
+never overridden:
+
+```bash
+WEBKIT_DISABLE_DMABUF_RENDERER=0 tokentracker-linux
+```
+
+WebKitGTK treats the variable as "set and not `0`", so `=0` genuinely restores
+the accelerated path while any other value (including an empty string) disables
+it.
+
+Note that when the webview aborts this way the app never reaches its shutdown
+path, so the bundled Node server is left running and keeps port 17680. Since
+OAuth sign-in requires that exact port, a later launch will fall back to a
+random port and browser sign-in will fail until the orphan is stopped:
+
+```bash
+pkill -f 'tokentracker/bin/tracker.js serve'
+```
+
 ## Sign-in
 
 The app prefers a **fixed loopback port, 17680**, because OAuth redirect URLs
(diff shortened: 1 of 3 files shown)


xiufengsun (approved):
Verified exact head 4f2fbf91 against current main. The WebKitGTK fallback runs before Tauri initialization, preserves an explicit user override, has focused regression coverage, and the full Linux/macOS/Windows/CodeQL checks are green. The Fedora Wayland/NVIDIA before/after evidence demonstrates the user-facing fix.

PR #514, opened 2026-08-25
fix: reset responsive drawer and manage modal focus

LceAn:
## Summary

- Reset an open mobile drawer when the viewport enters the desktop breakpoint so the hidden drawer unmounts and the body scroll lock is released.
- Add modal dialog semantics, move focus into the drawer, trap Tab focus, and restore focus to the mobile menu button after normal user dismissals.
- Use a separate breakpoint-close path that moves focus to the visible main-content region instead of the now-hidden mobile menu button.
- Preserve data-native-sidebar as the desktop aside's first attribute so the native Windows selector guard remains intact.

## Validation

- node --test test/windows-native-glass-selector.test.js
- npm --prefix dashboard test -- src/ui/components/Sidebar.test.jsx --minWorkers=1 --maxWorkers=1
- npm --prefix dashboard run lint
- npm --prefix dashboard run typecheck
- npm run validate:copy
- npm run validate:ui-hardcode
- npm run validate:guardrails
- git diff --check

## Summary by CodeRabbit

- **Accessibility**
  - Improved sidebar navigation labels and keyboard focus handling.
  - Added focus trapping for the mobile drawer and restored focus when it closes.
  - Added dialog semantics and expanded/collapsed state indicators.

- **User Experience**
  - Mobile navigation now closes with Escape, backdrop clicks, or when switching to desktop view.
  - Sidebar collapse preferences persist between sessions.

- **Tests**
  - Added coverage for desktop, mobile, focus, keyboard, persistence, and breakpoint behaviors.

### DIFF
--- a/dashboard/src/ui/components/Sidebar.jsx
+++ b/dashboard/src/ui/components/Sidebar.jsx
@@ -436,9 +436,9 @@ function Sidebar({ collapsed, onToggleCollapsed }) {
 
   return (
     <aside
+      data-native-sidebar
       id="app-sidebar"
       data-sidebar-state={collapsed ? "collapsed" : "expanded"}
-      data-native-sidebar
       aria-label={copy("nav.aside_label")}
       className={cn(
         "hidden lg:flex flex-col shrink-0 h-full min-h-0 transition-[width] duration-200",
@@ -453,7 +453,7 @@ function Sidebar({ collapsed, onToggleCollapsed }) {
 /**
  * Mobile drawer — slides in from the left, full-height overlay.
  */
-function MobileDrawer({ open, onClose }) {
+function MobileDrawer({ open, onClose, onDesktopBreakpointClose }) {
   const drawerRef = useRef(null);
   const closeButtonRef = useRef(null);
 
@@ -497,10 +497,10 @@ function MobileDrawer({ open, onClose }) {
     if (!open || typeof window === "undefined" || typeof window.matchMedia !== "function") return;
     const desktopQuery = window.matchMedia("(min-width: 1024px)");
     const onBreakpointChange = (event) => {
-      if (event.matches) onClose();
+      if (event.matches) onDesktopBreakpointClose();
     };
     if (desktopQuery.matches) {
-      onClose();
+      onDesktopBreakpointClose();
       return;
     }
     if (desktopQuery.addEventListener) {
@@ -509,7 +509,7 @@ function MobileDrawer({ open, onClose }) {
     }
     desktopQuery.addListener?.(onBreakpointChange);
     return () => desktopQuery.removeListener?.(onBreakpointChange);
-  }, [open, onClose]);
+  }, [open, onDesktopBreakpointClose]);
 
   // Lock body scroll when open
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


xiufengsun (changes requested):
I reviewed exact head `9ea808e933d3eba4acfe59a023c0e23a42a26877`. Two blockers remain:

1. The exact-head CI is red. The selector guard expects `/<aside\s+data-native-sidebar/`, while `dashboard/src/ui/components/Sidebar.jsx:438-442` inserts `id` and `data-sidebar-state` before `data-native-sidebar`. Either preserve the guarded native selector contract or update the guard and its consumers intentionally; then rerun the failed Node/macOS checks.

2. The responsive close path restores focus to a control that is hidden at the desktop breakpoint. `MobileDrawer` calls `onClose()` on the media-query transition at lines 494-504, and `closeDrawer` unconditionally focuses `menuButtonRef` at lines 594-599, but that button lives under `lg:hidden` at line 557. After resizing with the drawer open, focus can remain on an invisible element. Separate user-dismissal focus restoration from breakpoint cleanup and assert the active element after crossing 1024px.

Conclusion: NO-MERGE.

xiufengsun:
CI is red on both the Linux and macOS jobs, and the framing of the PR needs a correction before this can go in.

**The failure.** `test/windows-native-glass-selector.test.js:17` asserts `/<aside\s+data-native-sidebar/`. Inserting `id="app-sidebar"` and `data-sidebar-state` between `<aside` and `data-native-sidebar` breaks the match. That test exists because the Windows shell injects `html.native-windows-app .tt-native-glass-shell aside[data-native-sidebar]` (`TokenTrackerWin/DashboardWindow.cs:545`) and an earlier version keyed off `aside[aria-label]`, which also hit the mobile drawer. Keep `data-native-sidebar` as the first attribute on the element and put the new ones after it.

**"Restore" isn't accurate.** On current `main` none of the four things in the summary are broken:

- Collapse/expand and its `localStorage` persistence are live — `useSidebarCollapsed` at `Sidebar.jsx:78`, the chevron button at `:397`, wired at `:546`.
- The drawer already closes via the X button, the backdrop `onClick`, Escape, and `onItemClick`.

Please retitle and rewrite the summary to describe what the diff actually does, so the changelog doesn't claim a regression that never shipped.

**What is genuinely worth having.** Two real defects, and they're the reason I'm not just closing this:

1. `drawerOpen` never resets on the mobile → desktop transition. The container is `lg:hidden` so it disappears visually, but the `document.body.style.overflow = "hidden"` cleanup only runs when `open` flips to false. Widen the window with the drawer open and body scrolling stays locked. Its cleanup is also still reachable by Tab.
2. The drawer is a modal without `role="dialog"` / `aria-modal`, no focus move on open, no focus trap, no focus restore on close.

Fix the attribute order, rescope the description to these two, and I'll review the focus-trap implementation and the 180 lines of tests on the next push.

LceAn:
Updated in 4e64482293ceebd4ce043ab02659b87edeef3b15.

- Kept data-native-sidebar as the first attribute after the desktop aside tag; the native selector guard now passes.
- Split normal user dismissal from desktop-breakpoint cleanup. X, backdrop, Escape, and navigation dismissals still restore focus to the mobile menu button.
- Desktop-breakpoint cleanup now closes the drawer, releases the body scroll lock, and focuses the visible app-main-content target after the drawer unmounts.
- Extended the breakpoint regression test to assert the dialog is removed, body overflow is restored, and document.activeElement is the main-content target rather than the hidden menu button.
- Retitled and reframed the PR around the responsive drawer and modal-focus defects.

All targeted tests, lint, typecheck, production build, copy/UI/architecture validators, and git diff --check pass.

PR #515, opened 2026-08-25
fix: preserve precise Grok token usage and costs

paveezheng:
## Summary

Preserve Grok Build’s reported cache-write tokens and exact provider costs, while syncing the branch with upstream v0.93.3 so usage, pricing, and dashboard mirrors remain consistent.

## Scope

- [x] CLI (`src/`)
- [x] Docs / CI / config

## Checklist

- [x] `npm test` passes

---

<details>

### Risk layer triggers

### Rules / invariants

- Prefer `turn_completed.usage` over context-window watermarks when the provider exposes it.
- Keep input, output, reasoning, cache-read, and cache-write token columns mutually exclusive.
- Prefer a positive provider-reported `total_cost_usd` over local model-price reconstruction.
- Use the legacy watermark and pricing fallback only when exact provider fields are absent.

### Boundary matrix (list at least 3)

- Grok session log → normalized usage row: retain `cacheCreationTokens` and `costUsdTicks`.
- Normalized row → local pricing: use reported provider cost before pricing-table fallback.
- Normalized row → Dashboard edge mirrors: preserve the same cost precedence and DeepSeek tier fallback.
- Existing cursors → v5 migration: rebuild Grok rows so older normalized data adopts mutually exclusive token fields.

</details>

<details>

- **Delta since last Codex review:** Merged upstream v0.93.3 and resolved pricing/edge-function conflicts while retaining the Grok P0 changes.
- **Intended behavior / invariants:** Exact Grok cache-write usage and provider cost are collected when present; all pricing consumers use the same precedence.
- **Edge cases covered:** Missing `turn_completed.usage`, free Grok Build SKU, malformed numeric fields, cursor v4→v5 migration, project attribution, and dashboard pricing parity.
- **Tests run (command + result):** `npm run ci:local` — passed, including full Node test suite, architecture/localization/version checks, and Dashboard production build.
- **Known gaps / out of scope:** Historical or malformed Grok logs without provider usage still use the existing context-watermark fallback; live xAI schema changes may require future parser updates.

Pricing differences between CLI/session analytics and Dashboard edge functions, especially where provider-reported costs coexist with model-tier fallback pricing.

- [x] Full local CI
- [x] Parser and migration tests
- [x] Edge-pricing parity tests
- [x] Dashboard production build

### Uncovered scope

No live end-to-end xAI gateway test was added; verification uses representative recorded Grok usage payloads.

</details>

## Summary by CodeRabbit

* **New Features**
  * Added detailed Grok session usage, cost, context, runtime, tool, and error information.
  * Added nested subagent session views with expandable threads, hierarchy, filtering, and combined totals.
  * Improved session discovery across additional OpenCode, Copilot, and Trae data sources.
* **Bug Fixes**
  * Grok-reported costs are now reflected accurately, while unsupported reported values continue using estimated pricing.
  * Improved token and cache accounting, including reasoning and incomplete-usage indicators.
* **Localization**
  * Added Grok session labels and usage details across German, Japanese, Korean, Simplified Chinese, and Traditional Chinese.


xiufengsun (changes requested):
I reviewed exact head `3db7b728411d2c0330b794818518dff447e934f9`. The direction is useful, but three accounting blockers remain:

1. `src/lib/local-api.js:2270-2306` aggregates all rows for one source/model before calling `computeRowCost`. If one Grok row has provider-reported cost and a peer row has no cost, the aggregate is already non-zero, so the fallback pricing path is suppressed for the unpriced tokens. Compute provider/fallback cost per row before aggregation (and mirror the Vite path), with a mixed reported+missing-cost regression.

2. `src/lib/rollout.js:16024-16034` copies token fields into project deltas but drops `event.total_cost_usd`. Account totals can therefore preserve Grok's precise provider cost while project totals silently fall back or report a different value. Carry the cost through and add project/account parity coverage.

3. `src/lib/session-analytics.js:751-768` does not mark precision incomplete when a `turn_completed` event has missing or malformed `usage` after earlier valid usage. The session is then presented as exact even though a completed turn was omitted. Set the incomplete/partial flags for that case and cover valid-then-missing and valid-then-malformed sequences.

Conclusion: NO-MERGE. Green CI does not cover these mixed-data cases yet.

xiufengsun:
The Grok work here is right, and it's the part most people get wrong. Two changes before I merge, and the first is the blocker.

**1. The reported-cost branch is global, but the comment says it isn't.**

```js
// Some providers (currently Grok) persist an exact server-reported cost
const reportedCost = Number(row?.total_cost_usd);
if (Number.isFinite(reportedCost) && reportedCost > 0) return reportedCost;
```

The comment scopes this to Grok; the code applies it to every row of every source in `computeRowCost` and in all five edge functions. Two consequences:

- The pricing tables stop being the single basis for cost. Any source that ever writes a positive `total_cost_usd` silently bypasses them, and the divergence shows up as a number nobody can trace back to a rate.
- On the cloud side that field is unvalidated client input — `tokentracker-ingest.ts:196` is `total_cost_usd: Number(b.total_cost_usd) || 0`, straight from the request body. Today the leaderboard ranks on tokens, so this isn't a ranking exploit, but it does make the cost shown on profile and account pages a value the client gets to name.

Gate it on the sources that have actually earned it — a `SOURCES_WITH_AUTHORITATIVE_COST` set containing `grok`, checked alongside the positive-value test, mirrored verbatim into the five edges and asserted by `edge-pricing-parity.test.js` the same way you asserted the branch itself. Then the comment and the code say the same thing.

For the record on blast radius: production currently has 4 rows out of 2,707,591 with a positive `total_cost_usd` (3 `command-code`, 1 `tokentracker-local-manual-adjustment`), so this is about the contract going forward, not a historical repricing.

**2. Drop the unrelated changes to `dashboard/src/lib/bot/engine.ts` and `profiles.ts`.** Both diffs only remove trailing blank lines. That engine is vendored, and touching it costs a conflict on the next sync for no benefit here.

Keeping the rest as-is:

- Splitting `inputTokens` (whole prompt, cache included) from snake_case `input_tokens` (already non-cached) instead of subtracting cache tokens on both shapes. Assuming one semantic across both spellings is what inflated Codex cost 6–7× before.
- Returning `null` from `reportedCostUsd` when `costIsPartial || usageIsIncomplete`, so a partial turn falls through to model pricing rather than booking a truncated number.
- The `vite.config.js` fix — the dev mock's model aggregate was dropping `total_cost_usd` entirely.

Push those two and I'll merge.

xiufengsun:
Merged. Both points addressed, and the allowlist is the right shape — `new Set(["grok"])`, byte-identical across all five edges and `src/lib/pricing/index.js`, with the parity test now asserting the allowlist itself rather than just the branch. I checked that assertion isn't decorative: adding a second source to one edge's set turns `edge-pricing-parity` red, and dropping the `SOURCES_WITH_AUTHORITATIVE_COST.has()` guard from `computeRowCost` turns your new `ignores reported costs from non-authoritative sources` test red. Good tests.

**The reconciliation turned up something better than "no regression".** I ran the same Grok session through `main` and through this branch:

```
main:  input 34167  cached 128  output 12  reasoning 11  total 34307
this:  input 34167  cached 128  output  1  reasoning 11  total 34307
```

`main` is violating the column invariant — `34167 + 128 + 0 + 12 + 11 = 34318`, but `total_tokens` says 34307. The 11 reasoning tokens were being counted twice, once inside `outputTokens` and once on their own, because Grok's `outputTokens` is inclusive. This branch lands on `34167 + 128 + 0 + 1 + 11 = 34307` exactly. So this isn't only more precise, it fixes rows that were internally inconsistent — and since cost is computed from the columns and never from the total, `main` was also overbilling the reasoning portion. My sample is one session, but the invariant either holds or it doesn't, and it now holds.

Second sync queues 0 buckets, and `total_cost_usd: 0` on a session Grok reported no cost for correctly falls through to model pricing rather than booking a zero.

**One thing to be aware of, not a blocker.** Existing Grok rows in users' queues keep the old split; only newly parsed sessions get the corrected one. `total_tokens` is identical either way, so nothing in the ledger moves and no repair is needed — but a user's Grok output/reasoning breakdown will have a seam at the upgrade point. Worth remembering if someone reports their Grok output column looking discontinuous.

Thanks for turning the global branch around quickly — that one would have quietly made the pricing tables optional for every source.

PR #521, opened 2026-08-26
fix(linux): reap servers orphaned by a crashed app

csmashe:
Follow-up to #512, which fixed the Wayland abort itself. This fixes the mess that abort leaves behind — it applies to any crash, not just the DMA-BUF one.

## Problem

A GTK/Wayland failure calls `_exit(1)`, so neither `Drop for TokenTrackerServer` nor Tauri's `RunEvent::ExitRequested` runs, and the bundled Node server outlives its parent.

I verified the `_exit` path rather than inferring it from the exit code: I registered an `atexit` hook at startup and instrumented it. The crashing run spawned a child — proving `TokenTrackerServer::start()` ran and the hook was registered — yet the hook never fired. So `atexit`-based cleanup cannot work here, and I dropped that approach.

The orphan is not merely idle. It keeps port 17680, so the next launch hits the `pick_available_port` fallback and lands on a random port — and per `server.rs`'s own comment, OAuth sign-in needs that exact port. It then fails silently until the orphan is killed by hand. While testing #512 I accumulated four orphans across runs, on 17680, 45859, 40603 and 40555.

## Approach

Record each spawned server's identity, and stop a stranded one at startup before `pick_available_port` runs.

Identity is a persisted record, **not a filesystem path**. My first attempt matched `/proc/<pid>/exe` against the resolved `node` path, and that is broken for the artifact you actually ship: an AppImage mounts at a fresh `/tmp/.mount_XXXXXX` on every launch, so an orphan's `node` and `tracker.js` paths never match the ones the next launch resolves.

The record pins one exact process:

| Field | Why |
| --- | --- |
| `start_time` (`/proc/<pid>/stat` field 22) | A recycled PID has a different start time, so signalling can't hit a stranger |
| `boot_id` | `start_time` counts ticks since boot and resets, so a record surviving an unclean shutdown can't match an unrelated process on the same PID at the same tick |
| `owner_pid` + `owner_start_time` | Single-instance locking runs over the session D-Bus, so a second login session on one account starts a second app. "Orphaned" must mean the owner is gone, not that the child is alive — otherwise one session kills the other's healthy server |

Records are one-per-owner (`server-<owner_pid>.json`) so concurrent sessions can't delete or overwrite each other's entry and strand a server permanently. They live under a `0700` state directory as `0600` files, never in `/tmp` (world-writable: a forged record there could make this user signal their own PIDs), and are read with `O_NOFOLLOW` plus `fstat` on the descriptor actually read, so a planted symlink or foreign-owned file is rejected rather than followed.

## Process group

Servers now spawn with `process_group(0)` and are signalled by group.

`bin/tracker.js` re-executes itself through `spawnSync` when a proxy is configured, and `shouldRelaunchForProxy` fires for exactly `serve` — so for proxied users the process holding the port is a **grandchild**, and the recorded pid is a blocked wrapper. Signalling only that pid left the real listener on 17680.

This also fixes `stop_child`, which had the same gap on every normal quit, and the startup-failure and `restart_process` paths, which called `Child::kill()` directly.

## Verification

Each behaviour was exercised against a real build, not just reasoned about:

- crash → orphan on 17680 → relaunch reaps it and **reclaims 17680** (an earlier version reaped but still lost the port to a random one, because signalling doesn't release the listening socket synchronously — hence the bounded wait with SIGKILL escalation)
- live owner → 0 reaps, both processes untouched, **its record preserved**
- stale `boot_id` → 0 reaps, nothing signalled
- symlinked record → rejected
- spawned server is its own group leader (`pgid == pid`); records are `0600` in a `0700` directory

`cargo test --locked` 59 passing (4 new), `cargo check --locked --all-targets` clean, Linux JS tests 21 passing, `validate:guardrails` and `validate:versions` clean.

`cargo fmt --check` and `clippy` could not be run locally (this box has cargo without the rustfmt/clippy components), so CI is the authority on those two.

## Trade-off worth your call

`restart_process` now routes through `stop_child`, which sends SIGTERM and polls up to 1s before SIGKILL. That runs under the `SERVER` mutex, so a health-monitor restart holds the lock up to ~1s longer than the previous immediate `kill()`. Bounded, and only on restart — but it does cut against the comment in `main.rs` about not holding that mutex. Happy to shorten the grace period if you'd rather have the latency back.

## Scope

Linux-only; no `src/` or `dashboard/` changes. No version bump — `scripts/version-files.cjs` drives nine files off `package.json`, so that's yours to make on the next release.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux server startup, shutdown, and restart reliability.
  * Prevented stale or orphaned server processes from blocking new sessions or ports.
  * Enhanced cleanup of related processes, including descendants and proxy-launched processes.
  * Added safeguards to verify server identity and ownership before termination.
  * Added bounded shutdown timeouts with escalation when graceful termination fails.
  * Improved recovery after interrupted sessions and reporting when ports cannot be fully reclaimed.

### DIFF
--- a/TokenTrackerLinux/src-tauri/src/server.rs
+++ b/TokenTrackerLinux/src-tauri/src/server.rs
@@ -2,6 +2,8 @@ use std::ffi::OsString;
 use std::fs::OpenOptions;
 use std::io::{Read, Write};
 use std::net::{Shutdown, TcpListener, TcpStream};
+use std::os::unix::fs::{DirBuilderExt, MetadataExt, OpenOptionsExt};
+use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
 use std::process::{Child, Command, Stdio};
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -12,6 +14,11 @@ use std::time::{Duration, Instant};
 use crate::paths::RuntimePaths;
 
 const READINESS_PATH: &str = "/functions/tokentracker-user-status";
+
+/// How long to wait for an orphaned server to exit before giving up, and how
+/// long to wait before escalating SIGTERM to SIGKILL.
+const REAP_TIMEOUT: Duration = Duration::from_secs(3);
+const REAP_ESCALATE_AFTER: Duration = Duration::from_millis(750);
 
 /// How often the health monitor probes the server.
 pub const HEALTH_CHECK_INTERVAL: Duration = Duration::from_secs(15);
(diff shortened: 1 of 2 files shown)


xiufengsun (changes requested):
I reviewed exact head `ec654f18fe096358fc40df5307b561a2e8ef0c7f`. The PID-reuse and record-hardening work is solid, but the core port-recovery case still has a process-group blocker.

After signalling the group, `wait_for_exit` returns as soon as `record_still_matches(record)` becomes false (`TokenTrackerLinux/src-tauri/src/server.rs:349-368`). That predicate checks only the recorded leader PID. A leader can exit while a grandchild in the same process group continues to hold port 17680; the function then returns before escalation, reports the server as reaped, and startup can still choose a random port. Please wait for the process group (for example, PID-safe `kill(-pgid, 0)`) and/or verify the recorded port is free, and retain a safe escalation path even after the leader exits. Add an integration test where the leader exits on SIGTERM but a descendant keeps listening.

Conclusion: NO-MERGE.

csmashe:
Fixed in 41493c0f804586eee7037429b83c106329f4b72b. You were right, and it reproduced exactly as described.

**Confirming it first.** I extracted a `stop_recorded_server` seam with no behaviour change and wrote the integration test you asked for — a leader in its own process group, orphaned onto init, with a backgrounded subshell that ignores SIGTERM and keeps the listening socket. Against the unfixed wait it failed on precisely your prediction:

```
the descendant kept port 43563; reaping must wait for the whole process group,
not just the recorded leader, or startup silently falls back to a random port
```

Two details decide whether that test means anything, and both bit me while writing it. The tree has to be genuinely orphaned: my first attempt left the leader as the test's own child, so it lingered as a **zombie**, kept its `/proc/<pid>/stat`, and `record_still_matches` stayed true — the old code passed for the wrong reason. And POSIX hands an asynchronous list `/dev/null` for stdin, so the socket needs `exec 3<&0` before backgrounding or the descendant quietly loses it and the test asserts on a port nobody holds.

**The fix.**

- `wait_for_exit` now polls the process **group** via `group_still_alive` (`kill(-pgid, 0)`, `EPERM` counted as alive) and returns whether it emptied. This is strictly stronger than probing the port: a task closes its file descriptors during exit, before it leaves its process group, so an empty group is proof the socket is gone.
- Escalation no longer requires `leads_own_process_group`. That predicate *cannot* hold once the leader exits, so requiring it is exactly what disabled the SIGKILL path in the case that needed it. PID-safety now rests on continuity instead: the loop returns the instant the group is empty, so reaching escalation means the group was non-empty at every observation, and the kernel keeps a pid number reserved while it is still some group's pgid — it cannot have been freed and recycled into a stranger's group underneath the wait.
- Belt and braces on your "verify the recorded port is free": a descendant that called `setsid()` leaves the group and survives the group signal, so `stop_recorded_server` checks the port rather than assuming. `reap_orphaned_servers` returns `ReapOutcome { pid, port, port_recovered }` and startup logs either `port 17680 released` or `still holds port 17680; this launch will fall back to a random port and OAuth sign-in may fail`. Claiming a recovery that did not happen is how this stayed invisible.
- `stop_child` had the same gap on **every normal quit**, which my earlier commit message wrongly claimed to have closed: it waited on `child.try_wait()` and gated escalation on `!exited`, so a leader dying first meant no escalation and a surviving descendant. It waits on the group too now, reaping the leader inside the loop since a zombie is still a group member.

**Tests.** Three added: the group case above; a descendant that escapes via `setsid` and must be *reported* rather than claimed reaped; and `group_still_alive` rejecting pid ≤ 1 so a group signal is never aimed at the whole session. They share an `OrphanTree` harness and skip cleanly where `setsid` is unavailable. 63 tests pass, `cargo clippy --all-targets` clean, `cargo fmt --check` clean.

xiufengsun:
Merged.

**On the trade-off you flagged: keep the 1s grace, don't shorten it.** The `main.rs` comment is about not holding `SERVER` across `probe_server_http`, which runs on *every* health tick with second-scale timeouts. `restart_process` only runs after a probe has already declared the server dead, and the wait is bounded and escalates. Trading up to 1s on a rare restart for a SIGKILL that can strand a descendant holding the port is the wrong direction — that stranded descendant is the exact bug this PR exists to fix.

The parts that make this reviewable rather than merely plausible:

- Verifying the `_exit` path with an instrumented `atexit` hook that never fired, instead of inferring it from the exit code, and then dropping the `atexit` approach because of it.
- Rejecting `/proc/<pid>/exe` matching once you found the AppImage remounts at a fresh `/tmp/.mount_XXXXXX` per launch. That's the shipped artifact, and a path-identity scheme would have passed every local test and never reaped anything in the field.
- `owner_pid` + `owner_start_time`. Single-instance locking is per session D-Bus, so without it a second login session reaps the first one's healthy server.
- `pid <= 1` rejected before any `kill(-pid, …)`. `kill(-1, SIGKILL)` is POSIX broadcast, and that's the failure mode you don't get a second chance at.
- `O_NOFOLLOW` plus `fstat` on the descriptor actually read, `0600` in `0700`, never `/tmp`.

One note, no action needed: `wait_for_exit` not rechecking `leads_own_process_group` before escalating is correct as reasoned, and the residual PID-reuse window is closed in practice because a recycled PID inherits its parent's pgid and would have to call `setpgid(0, 0)` to land in the recorded group.

Reaping runs synchronously ahead of `pick_available_port`, but only when a record exists, so a clean launch pays nothing. Fine.

`cargo fmt` / `clippy` are green on CI. No version bump needed here — that's mine on the next release.

PR #522, opened 2026-08-26
feat(linux): publish .deb and .rpm alongside the AppImage

csmashe:
Linux currently ships a single AppImage. Two reasons to also ship native packages:

**AppImages carry a lot of bloat.** They bundle their own copy of GTK, WebKit and every supporting library so they can run anywhere, and the user pays for that on every download and on disk. Measured on this change: the AppImage is ~120 MB while the same app as a `.deb` or `.rpm` is ~55 MB — **less than half**, for identical functionality. The distro already ships those libraries; the AppImage just refuses to use them.

**Many users prefer their platform's own packaging.** `apt install` / `dnf install` gives them the system's install, upgrade and removal path, a properly registered desktop entry, and no FUSE dependency — instead of a loose executable they have to `chmod +x`, place somewhere sensible, and update by hand. For plenty of people that's the difference between installing the app and not bothering.

This adds `.deb` and `.rpm`, built by the same `tauri build` run.

**This is a policy change, not a bug fix.** The single-target limit was deliberate and pinned by a test whose message says so:

```rust
"Linux ships a single AppImage; adding targets here multiplies release assets"
```

So it's entirely reasonable to say no. I've rewritten that test rather than deleting it — the target list stays an explicit set, with a message pointing at the three places that must change together (build, payload verification, publish asset list).

## Sizes

| Asset | Size |
| --- | --- |
| `TokenTracker-linux-x86_64.AppImage` | ~120 MB |
| `TokenTracker-linux-x86_64.deb` | ~55 MB |
| `TokenTracker-linux-x86_64.rpm` | ~55 MB |

The trade-off: the packages depend on the distro shipping a compatible `webkit2gtk-4.1`, where the AppImage carries its own. That's exactly why this is *additive* — the AppImage stays for anyone who needs the self-contained version.

## Payload verification

The existing check extracted the AppImage and failed if `EmbeddedServer/node` was missing — the thing that catches a broken `bundle.resources` mapping before it reaches users. Each format is produced by a **separate** tauri bundler and can fail independently, so that check now runs against all three rather than treating the AppImage as a proxy. It's factored into one `verify_payload` function asserting the same things it always did: the binary, a complete `EmbeddedServer` (`tracker.js`, `package.json`, `dashboard/dist/index.html`), and an executable `node`.

`rpm2cpio` and `cpio` are not on `ubuntu-latest` by default, so the rpm arm installs them; the test asserts that, since without them the rpm could never be inspected.

## Verified locally

Built and unpacked both packages rather than assuming:

- Both install the runtime at `/usr/lib/TokenTracker/EmbeddedServer/node`, which `candidate_runtime_roots` already probes via `<prefix>/lib/<PRODUCT_DIR_NAME>` — so a packaged install resolves its runtime with **no code change**
- `node` is executable in both; all three required `EmbeddedServer` files present
- I extracted the workflow's verification script and ran it against the real bundles, then rebuilt a `.deb` with `EmbeddedServer` stripped and confirmed it fails with a non-zero exit — the check is not decorative
- Removing the `rpm`/`cpio` apt deps fails the workflow test

The AppImage arm could not be exercised on my machine (`linuxdeploy` needs FUSE, which CI installs) — that arm is unchanged apart from being moved into the shared function, but CI is the authority on it.

## One caveat worth your call

Tauri hardcodes `libappindicator3-1` as the deb dependency. That package exists in Ubuntu 24.04 (universe) but **not** Debian 12, so the `.deb` won't install there. `bundle.linux.deb.depends` only *appends* to Tauri's defaults — I tried overriding it and it produced duplicate entries while leaving the original in place — so it can't be fixed from config. Documented in the README (Debian users are pointed at the AppImage). If you want it properly fixed, it needs a repack step in CI or a fix upstream in Tauri; happy to add the former if you'd prefer.

The Arch `PKGBUILD` is untouched and stays a local-build path.

## Summary by CodeRabbit

- **New Features**
  - Added Linux release packages in AppImage, DEB, and RPM formats.
  - AppImage bundles GTK/WebKit dependencies and requires only a recent glibc.
  - DEB and RPM packages support Debian/Ubuntu and Fedora/RHEL installations.
  - Added guidance for Debian 12 users to choose the AppImage instead of the DEB package.

- **Documentation**
  - Updated installation instructions across supported languages with package options, commands, and dependency details.

- **Tests**
  - Added validation for all Linux package formats and release assets.


xiufengsun (changes requested):
I reviewed exact head `4b42b615f4160186ae41df2d2a16ee476b8f9b87`. The package verification is thorough, but this changes the release contract from one Linux AppImage / four total assets to AppImage + deb + rpm / six total assets without updating the repository source of truth.

`CLAUDE.md` still explicitly requires a single AppImage and exactly four release assets, and the release/read-back procedure is defined from that contract. Please update `CLAUDE.md` and all aligned release documentation/tests in the same PR, including the expected asset inventory and verification steps, so maintainers and automation do not enforce contradictory policies. Then rerun the combined release workflow tests.

Conclusion: NO-MERGE until the canonical release contract is updated.

csmashe:
Fixed in 14bbd3d7f8f720cc992b97fb288ae0f8112ba944. You're right that the contract was left contradicting the workflow — thanks for catching it.

**`CLAUDE.md`**
- The `linux` job no longer "builds a **single AppImage**": it builds all three packages from one `tauri build`, and the payload verification is described as per-format, since each format comes from its own bundler and can fail alone.
- "verifies all four assets" → **six**.
- Added an explicit **Release asset inventory** table (dmg / win zip + installer / AppImage + deb + rpm), with the rule that changing the list means touching the workflow's upload and required-asset steps and `test/linux-client-workflow.test.js` together. The publish job's required-asset check is what actually refuses to flip the draft live, so the list maintainers read is now the same list.
- Step 3 and the embedded-runtime paragraph updated to match.

**Aligned docs**
- `TokenTrackerLinux/README.md`: the earlier commit had left two overlapping AppImage bullets and an `## Install` section that jumped straight into the AppImage with no heading, so the new apt/dnf sections dangled under it. Restructured into one release-packages bullet and three parallel subsections.
- `README.md` + the four translations: `.deb`/`.rpm` install paths added.
- `.github/workflows/ci.yml`: the comment explaining why bundling is left to the release workflow now names all three formats and the rpm tooling.

**Tests** — the two that pinned the old contract were already rewritten rather than deleted in 4b42b615 (`bundle_produces_the_three_supported_linux_targets` keeps the target set explicit; the publish test now asserts all six assets). Reran on the new head:
- `node --test` over `linux-client-workflow`, `release-dmg-workflow`, `release-windows-workflow`, `npm-publish-workflow`, `linux-bundle` — **72 pass, 0 fail**.
- `cargo test --test capabilities` — **9 pass, 0 fail**.
- Full `npm test` — 2423 pass, 5 skipped. One file, `test/legacy-baseurl-migration.test.js`, intermittently dies with the node test-runner's `Unable to deserialize cloned data` IPC error; it passes 14/14 run alone, and this commit is docs-only.

**One correction that came out of this.** Every README claimed the AppImage "needs `webkit2gtk-4.1`, `gtk3` and `libayatana-appindicator` from your distro". The build tree says otherwise: `TokenTracker.AppDir/usr/lib` carries `libwebkit2gtk-4.1.so.0`, `libjavascriptcoregtk-4.1.so.0`, `libgtk-3.so.0` and `libappindicator3.so.1`, plus the `WebKitWebProcess` / `WebKitNetworkProcess` helpers in `usr/libexec`, and AppRun prepends `$APPDIR/usr/lib` and `usr/lib64` to `LD_LIBRARY_PATH`, so those copies are what load. The system stack is what the *packages* need — the `.deb` declares `libwebkit2gtk-4.1-0`, `libgtk-3-0`, `libayatana-appindicator3-1`, `libappindicator3-1` and the `.rpm` the equivalent sonames — which is also the 55MB vs 120MB gap, and why the `.deb` is uninstallable on Debian 12. `librsvg` is dropped from the runtime lists: it's a build dependency, and nothing named rsvg is in the AppDir.

xiufengsun:
The `CLAUDE.md` correction is right and the per-format payload verification is the part I'd have asked for. One change before I merge.

**`pick_one` writes its error to stdout, which is captured.** It's only ever called inside a command substitution, so the `::error::` line lands in `$image` / `$deb` / `$rpm_pkg` instead of the log. I ran it:

```
image="$(pick_one AppImage /nonexistent/*.AppImage)"
→ exit 1, stdout empty, stderr empty
```

The failure itself is correct — `set -e` still aborts the step, because the assignment isn't `local` and so doesn't mask the substitution's status. But the job dies with no line saying why, in the one workflow where a silent failure is most expensive: the draft is already created, two other platforms have uploaded, and whoever picks it up sees a red step with no message. Send it to `>&2`.

While you're there, `verify_payload`'s errors are fine as-is — it isn't called in a substitution, so those reach the log.

**Not a blocker, just noting the risk we're both accepting.** Nothing in CI exercises the deb/rpm bundlers or the extraction path; `linux-client-workflow.test.js` asserts the YAML, not the build. First real proof is a dispatch. The failure mode is safe — a bundler failure leaves the release unpublished rather than half-public — but it will block a release when it happens, so I'll dispatch this on the next version rather than a hotfix.

Push the stderr fix and I'll merge.

PR #523, opened 2026-08-26
fix(serve): only stop a verified TokenTracker server on a busy port

csmashe:
`serve --port <n>` sets `portExplicit`, which wires `ensurePortFree` into `listenOnAvailablePort`. That function kills **every pid `lsof` reports for the port** — SIGTERM, then SIGKILL three seconds later — with no check that the process was ever ours:

```js
const targets = pids.filter((p) => p !== self);
```

The only protection is "not my own pid". A caller-supplied port may be owned by an unrelated application, and failing to bind is far safer than terminating a stranger's process.

Two independent problems, both fixed here.

## 1. No identity check before signalling

A TokenTracker server is always `node <somewhere>/bin/<entry> serve` — either the real script (npm global, embedded desktop runtime, repo checkout) or one of the published npm bin shims, which `ps` reports by the shim path rather than the resolved script.

Path shape alone is **not** identifying, though: plenty of unrelated projects ship a `bin/tracker.js`. So the entry is resolved through `realpathSync` and must have a `package.json` named `tokentracker-cli` within three levels above it. That covers every real layout — repo checkout, npm global, embedded desktop runtime, and the bin shims, which are symlinks into the package and resolve to the same place.

This fails closed: an entry that cannot be resolved (deleted, or inside a mount namespace this process cannot read) is left alone, so the worst case is a failed bind rather than a killed stranger.

Identity is rechecked before the SIGKILL escalation, since the pid may have exited during the three-second wait and been recycled by something else.

The matching is deliberately conservative about `ps -o command=`, which joins argv with spaces and drops all quoting — so the script path is only unambiguous relative to the `serve` argument that follows it. Splitting on whitespace would reject an install under `~/Token Tracker/`.

## 2. The port scan matched clients, not just listeners

`lsof -i tcp:<port>` matches a socket whose **local or remote** port matches. I confirmed this directly: with a listener on 17680 and one client attached, `lsof -ti tcp:17680` returned both pids; adding `-sTCP:LISTEN` returned only the listener. So a browser merely connected to the dashboard was reported as occupying the port, and was equally liable to be killed.

## Windows is unaffected

`findPidOnPort` shells out to `lsof`, which doesn't exist on Windows — the call throws, the catch returns `[]`, and `ensurePortFree` is already a no-op there. This change adds a `ps` call on the same path, so it stays Unix-only and Windows behaviour is identical before and after.

## Verified against live processes

Not just unit-tested:

- An unrelated `node` listener on the port **survives** `ensurePortFree`, and the bind then fails with `EADDRINUSE` — the correct outcome
- A real `node bin/tracker.js serve --port …` on the same port **is** still stopped and the port released, so the cleanup feature still works
- The matcher was exercised against realistic commands: npm global install, embedded desktop runtime, a path containing a space, the npm bin shims (all matched); a lookalike `tracker.js` outside `bin/`, a `.bak` suffix, our own `sync` process, `python3` instead of node, postgres and nginx, and the empty string `ps` prints once a pid is gone (all rejected)
- Both halves are mutation-tested: deleting the identity filter from the kill path fails the new integration test, and dropping `-sTCP:LISTEN` fails the scan test. An earlier version of my tests passed with the filter removed, so the integration test exists specifically to close that gap

`npm test` — 2427 passing, 0 failing. `validate:guardrails`, `validate:copy`, `validate:ui-hardcode`, `validate:versions` and the architecture guardrails all clean.

## Scope

`ensurePortFree` is newly exported so the kill path can be tested against a real process rather than only through the matcher. No version bump — `scripts/version-files.cjs` drives nine files off `package.json`, so that's yours on the next release. This touches `src/`, so by the CLAUDE.md matrix it ships npm plus all three desktop apps whenever you next cut one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server startup when the configured port is already in use.
  * Prevents unrelated processes from being terminated during port cleanup.
  * Ensures only actively listening endpoints are considered when checking port availability.
  * Adds additional safeguards before forcefully stopping a server process.
  * Improves recognition of the correct server process, including when launched through a package-managed command or symlink.

### DIFF
--- a/src/commands/serve.js
+++ b/src/commands/serve.js
@@ -372,19 +372,36 @@ const TRACKER_ENTRY_RE = new RegExp(
   String.raw`(?:^|[\\/])\.?bin[\\/](?:tracker\.js|tokentracker-cli|tokentracker-tracker|tokentracker|tracker)$`,
   "i",
 );
-// `ps -o command=` joins argv with spaces and drops all quoting, so the script
-// path is only unambiguous relative to the `serve` argument that follows it --
-// splitting on whitespace would reject an install under `~/Token Tracker/`.
-const NODE_SERVE_COMMAND_RE = /^(\S+)\s+(.+?)\s+serve(?:\s|$)/i;
 const NODE_EXECUTABLE_RE = /(?:^|[\\/])node(?:\.exe)?$/i;
+// Every `serve` token that could be the subcommand rather than part of a path.
+const SERVE_BOUNDARY_RE = /\s+serve(?=\s|$)/g;
 
-// The script path a `node ... serve` command would run, or null.
+// The tracker entry path a `node ... serve` command would run, or null.
+//
+// `ps -o command=` joins argv with spaces and drops all quoting, so the script
+// path is only delimited by the `serve` argument after it -- splitting on
+// whitespace would reject an install under `~/Token Tracker/`. But `serve` can
+// occur inside the install path too, so every boundary is tried and the one
+// whose prefix is actually a tracker entry wins; taking the first would parse
+// `~/my serve dir/bin/tracker.js` as `~/my`.
 function parseServeScriptPath(command) {
   const value = String(command || "").replaceAll("\0", " ").trim();
   if (!value) return null;
-  const match = NODE_SERVE_COMMAND_RE.exec(value);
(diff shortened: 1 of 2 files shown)


csmashe:
@coderabbitai review

PR #527, opened 2026-08-27
fix: discover Copilot OTel logs and preserve chat usage

FLC-niko:
## Problem

Copilot Chat file-exporter logs can be written under ~/.copilot-otel, while TokenTracker only scanned ~/.copilot/otel. A desktop process that does not inherit the custom exporter environment therefore missed the file entirely. Chat-extension LogRecords can also share one nested spanContext across multiple model requests even when gen_ai.response.id differs, so the old deduplication key under-counted usage.

## Changes

- Discover both known default OTel directories plus the explicit exporter path.
- Use gen_ai.response.id for Chat LogRecord deduplication while retaining top-level traceId/spanId for CLI spans.
- Migrate verified v2 cursor prefixes so previously collapsed Chat records are corrected once without replaying CLI history.
- Show the detected OTel file in status output.
- Add resolver, deduplication, migration, and status regression tests.

## Scope

This PR does not invent token usage when VS Code emits no gen_ai.usage.* fields. That is an upstream VS Code issue tracked at https://github.com/microsoft/vscode/issues/331455; such a source file still needs an upstream exporter fix.

## Validation

- Copilot parser, session-store, App DB, and usage-limits tests: 429 passed, 0 failed.
- npm test was also run; unrelated local environment failures occurred for missing dashboard dependencies, Swift toolchain/module-cache permissions, and a missing dashboard artifact.

## Summary by CodeRabbit

- **New Features**
  - Copilot usage detection now recognizes telemetry files from both standard telemetry locations and configured exporter paths.
  - Status information now reports all detected telemetry locations and provides clearer fallback path details.

- **Bug Fixes**
  - Copilot conversations with distinct response IDs are now counted separately, improving usage totals.
  - Existing usage data is migrated and corrected automatically when upgrading from the previous tracking format.

### DIFF
--- a/src/commands/status.js
+++ b/src/commands/status.js
@@ -1215,8 +1215,10 @@ function formatCopilotLines({ token, otel, sessionStore, appDb }) {
   const appDbState = appDb?.app_db_has_file
     ? `${appDb.app_db_mode || "set"} (${(appDb.app_db_paths || []).join(", ")})`
     : `not found (${appDb?.app_db_path || "unknown"})`;
+  const otelLocation =
+    otel.otel_path || otel.otel_detected_paths?.[0] || otel.otel_default_dir;
   const usageState = otel.otel_has_files
-    ? `set (${otel.otel_path || otel.otel_default_dir})`
+    ? `set (${otelLocation})`
     : otel.otel_enabled
       ? "enabled but no files yet"
       : "unset (OTEL export not enabled)";
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -13313,7 +13313,7 @@ async function parseCraftIncremental({
 //   COPILOT_OTEL_ENABLED=true
 //   COPILOT_OTEL_EXPORTER_TYPE=file
 //   COPILOT_OTEL_FILE_EXPORTER_PATH=$HOME/.copilot/otel/copilot-otel-...jsonl
-// We scan the default directory plus the env-overridden path.
+// We scan both known default directories plus the env-overridden path.
 // ─────────────────────────────────────────────────────────────────────────────
 
 function resolveCopilotOtelPaths(env = process.env) {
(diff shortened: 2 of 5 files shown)


ernarunshan:
Does the new .copilot-otel discovery find actual VS Code Chat LogRecords containing gen_ai.request.model, gen_ai.response.id, and gen_ai.usage.*, or does it only improve handling for installations where those fields were already present?

xiufengsun (changes requested):
I reviewed exact head `d22389102e825bd08887749d5771d539701ebb5f`. The v3 response-ID dedup is correct for available files, but the migration can be permanently blocked by normal log rotation.

`src/lib/rollout.js:13676-13685` requires every historical path in persisted `fileOffsets` to appear in the current discovery set; if one already-consumed OTEL file was deleted or rotated, migration returns `applied:false`. The cursor stays at v2 and retries the same impossible condition on every sync, so new Chat LogRecords continue using the old shared-span fallback and may remain undercounted indefinitely. Define a safe policy for retired files (for example, settle/prune an unavailable historical offset without blocking reachable prefixes) and add a regression with one deleted tracked file plus one live file containing new records.

Conclusion: NO-MERGE.

xiufengsun:
Merged.

@ernarunshan asked whether the new `~/.copilot-otel` discovery finds Chat LogRecords that actually carry `gen_ai.usage.*`, or only improves handling where those fields were already present. The answer is the second one, and the Scope section says so: this reads what the exporter writes and invents nothing. I verified the upstream reference — microsoft/vscode#331455 is open, and its title is precise about the limit: *"Copilot Chat OTel export only emits utility-model spans, never the active chat model (GPT-5/Claude)"*. So for the main chat model this changes nothing until VS Code ships the export; what it fixes is the directory that was never scanned, and the records that were being collapsed.

The migration is the part I looked at hardest, because a repair that rewrites already-counted history is how this project has lost user data before. This one is built the right way:

- It rescans and computes both the old-key and new-key totals, then applies the **difference** (`subtractTotals` / `addTotals`) instead of replaying into the bucket.
- `copilotTotalsCover` gates it across all eight fields — if any existing bucket doesn't cover the old contribution, the whole migration is abandoned and the cursor stays at v2 to retry. So an unexpected state defers rather than half-applying, and no field can go negative.
- CLI spans are deliberately untouched: their v2 `traceId:spanId` key was already correct.

I checked the migration test isn't a false green, since `eventsAggregated === 0` alone would pass even if the repair did nothing. It isn't — it asserts the queued bucket at `input 1300 / output 140 / total 1440 / conv 2` against the collapsed `550 / conv 1` prior state. Stubbing out the `subtractTotals` + `addTotals` pair turns it red immediately. 359/360 green locally on a fresh checkout.

Ships with the next release.

PR #529, opened 2026-08-28
feat(omo): add OmO as a first-class usage source

ingyeomnote:
## Summary

Adds **OmO** as a first-class local source, tracked separately from oh-my-pi (`omp`) and pi.

OmO writes the same append-only JSONL sessions as oh-my-pi:

```
~/.omo/agent/sessions/--<cwd-encoded>--/<timestamp>_<sessionId>.jsonl
```

This change generalizes the existing omp reader (`parseOmpLikeIncremental`) instead of duplicating it. omp and omo differ only in:

| | oh-my-pi | OmO |
|---|---|---|
| Home | `~/.omp` | `~/.omo` |
| Cursor / source | `omp` | `omo` |
| Reasoning field | `usage.reasoningTokens` | `usage.reasoning` |
| `PI_CONFIG_DIR` / `PI_CODING_AGENT_DIR` | honored (shared with pi) | **not** honored |

`PI_*` env vars stay with pi/omp. Routing them to a third provider would reintroduce the collision `decidePiCodingAgentDirOwner` exists to prevent. Overrides are `TOKENTRACKER_OMO_AGENT_DIR` / `TOKENTRACKER_OMO_HOME` / `OMO_HOME`.

## Costing

OmO reports reasoning as a **subset of `output`**, and `totalTokens` excludes it (`input + output + cacheRead + cacheWrite === totalTokens`). Its own `usage.cost` has no reasoning component.

So OmO follows the Codex convention: `reasoning_output_tokens` is informational and must not be billed on top of output (`computeRowCost` + the token-audit invariant).

## Dashboard

- Display name `OmO` (copy key + zh / zh-TW)
- Brand mark at `dashboard/public/brand-logos/omo.svg`
- Listed in `AGENT_LOGOS` so the marketing tool strip picks it up

## Out of scope

- No notify extension (`omp-hook` equivalent). Passive session scan is the source of truth; `tokentracker serve` already periodic-syncs.
- No change to `resolveKilocodeRoots` (`split(":")` on Windows). That is a pre-existing test issue on drive-letter paths, unrelated to this source.

## Test plan

- [x] `test/omo-parser.test.js` — field spelling, totalTokens exclusive of reasoning, independent omp/omo cursors, path overrides, scan layout, no double-billing
- [x] Provider display + ProviderIcon tests
- [x] `validate:copy`, `validate:locale`, `validate:ui-hardcode`, `validate:guardrails`
- [x] Live parse of a real `~/.omo` install (separate `omo` buckets, omp totals unchanged)

## Summary by CodeRabbit

- **New Features**
  - Added OmO as a supported AI coding tool, including dashboard display, localized naming, welcome-screen listing, and documentation updates.
  - Added OmO session discovery and synchronization, including passive-session status reporting.
  - Added separate OmO usage tracking and source-specific parsing.

- **Bug Fixes**
  - Prevented duplicate OmO/OmP session counting when directories overlap.
  - Corrected OmO token accounting to avoid double-counting reasoning tokens.

- **Tests**
  - Added coverage for OmO parsing, discovery, synchronization, pricing, and directory handling.


xiufengsun (changes requested):
I rechecked exact head `59b15124fd76436a2fd2fb360b428b1f41e3c3f2`. The later commits appear to address the earlier OmO/OMP collision and reasoning-fallback findings, so I am not repeating those items.

The current branch is now `DIRTY` against the latest main after today's merges and v0.93.5 work. GitHub cannot construct a merge result, and the green checks are for the pre-integration head rather than a current-main merge. Please rebase or merge the latest main, resolve the conflicts, and push the resulting head; CI must rerun on that exact integration before semantic merge review can finish.

Conclusion: NO-MERGE until the branch integrates cleanly with current main.

xiufengsun:
The approach is right and the accounting homework is the part most new-provider PRs skip, so this is close. It needs a rebase before I finish the review — the branch conflicts with `main`.

**What's already correct.**

- The reasoning question is answered with evidence, not assumption. `input + output + cacheRead + cacheWrite === totalTokens` across 2,586 messages, with no separate reasoning component in `usage.cost`, is what justifies `reasoningIncludedInOutput: true`. Getting this wrong is how Codex/every-code inflated cost 6–7× before, so the corpus check is the right bar.
- Factoring `parseOmpLikeIncremental` out instead of copying the omp parser, with the reasoning spelling as a per-provider `reasoningFields` list rather than a both-spellings guess.
- `omoAgentDirCollidesWithOmp`, and deliberately not honoring `PI_CONFIG_DIR` / `PI_CODING_AGENT_DIR`. Routing a third provider through the dir that `decidePiCodingAgentDirOwner` exists to disambiguate would have been a real bug.

**Rebase.** The provider count in the five READMEs, `dashboard/index.html`, and `llms.txt` moves whenever a source lands on `main`, so recount rather than resolving to your number — it's an exhaustive search, the count appears in more places than the conflict markers will show you.

**One non-blocking note for the follow-up.** `entryId` is OmO's 8-hex `id`, and `seenIds` is capped at 10,000. Within one window that's roughly a 1% chance of a birthday collision, and a collision means the message is silently *skipped*, not double-counted. This is inherited from the existing omp parser, not introduced here, so it isn't a blocker for this PR — but since you're the one touching this code path, it's worth checking whether OmO exposes anything longer, or whether keying on `sessionId + id` is cheap here.

Push the rebase and I'll review the parser and the 323 lines of tests against the exact head.

xiufengsun:
I rebased this onto main for you rather than leaving it stale, since the branch was 203 commits behind and the conflicts were all in the tool-count surfaces rather than in your parser. Your commits are untouched; mine is a merge commit on top.

What I resolved and why, in case you disagree with any of it:

- **Tool count is now 39.** It is derived from `SUPPORTED_PROVIDERS` in `src/commands/init.js`, and main has since refactored `test/discovery-metadata.test.js` to compute `TOOL_COUNT` from that list instead of hardcoding it, so I took main's form of those assertions. Adding OmO to the list is what moves every surface.
- **`reasoningIncludedInOutput`** keeps main's `acode` alongside `codex` / `every-code`, plus your `omo`, and I kept your comment explaining the OmO evidence.
- **`oh-my-pi` stays out of the passive-reader sentence.** Your branch listed it there, which was correct when you wrote it, but main now documents that `tokentracker init` writes a managed notify extension for OMP, so "nothing is installed into those tools" no longer applies to it. OmO is in that list, since your integration installs nothing.
- **`validate-locale-coverage.cjs`** takes both sides: main's new `limits.*` patterns and your `omo` in the `provider.display` alternation.
- Counts and the OmO entry propagated through five READMEs, `dashboard/index.html` (including the JSON-LD ItemList and the FAQ text), `dashboard/public/llms.txt` and `agent-logos.js`.

Verified on the merged tree: `test/omo-parser.test.js` 9/9, `test/discovery-metadata.test.js` 5/5, and `validate:copy` / `validate:ui-hardcode` / `validate:guardrails` all exit 0. Full `npm test` is 2853/2859 with 4 failures that are all missing local dev dependencies in my worktree, confirmed by symlinking `dashboard/node_modules` and watching the two Vite ones go green.

I am waiting on CI, then I will finish the review. The accounting homework in the original PR is still the part that made this worth picking up: proving `input + output + cacheRead + cacheWrite === totalTokens` across 2,586 messages is what justifies `reasoningIncludedInOutput`, and it is exactly the step that has burned this project before.

xiufengsun:
CI is green on the rebased branch, and I have finished the review. This is good to merge.

What I checked, since the risk in a new source is always double counting or a wrong token normalization:

**The parser is a thin wrapper, not a copy.** `parseOmoIncremental` delegates to `parseOmpLikeIncremental` with `stateKey: "omo"`, `source: "omo"` and its own resolvers. So dedup, subagent transcript handling and the cursor mechanics are the already-exercised omp/pi implementations rather than a second copy that can drift. That is the right shape, and it is why this needed much less scrutiny than a from-scratch parser would have.

**The omp collision is handled.** `omoAgentDirCollidesWithOmp` compares the resolved dirs and, on a match, `sync.js:2415-2420` empties both file lists so OmO is skipped entirely and omp keeps the transcripts. A user pointing `TOKENTRACKER_OMO_AGENT_DIR` at `~/.omp/agent` gets one count, not two. It follows the existing pi precedent.

**Error isolation is correct.** `parseOmoIncremental` is inside the try/catch that routes to `warnProviderParseFailure`, so an OmO failure cannot take the rest of sync down.

**The idempotency test is real.** It reuses the same `cursors` object across both calls and asserts the queue line count, not just the returned counters, so a second pass that re-queued an identical bucket would fail it. 9/9 locally.

**`reasoningIncludedInOutput: true`** is the one claim that could not be verified from the code alone, and your corpus check is what carries it: `input + output + cacheRead + cacheWrite === totalTokens` across 2,586 messages with no separate reasoning component in `usage.cost`. That is exactly the homework Codex and every-code needed and did not get the first time, when a naive copy inflated cost 6-7x.

One non-blocking gap for a follow-up: the test covers an unchanged file on the second pass, but not a file that gained new lines between passes. The append path is where an off-by-one in the cursor would show up as either lost or double-counted turns, and it is the more likely regression shape for an append-only log.

Thanks for the integration, and sorry it sat as long as it did.

PR #533, opened 2026-08-28
fix(linux): guard bundle cleanup and write the Tauri icon atomically

csmashe:
## What

Hardens the two destructive file operations in the Linux bundle scripts.

**`bundle-node-linux.sh`** runs `rm -rf "$EMBED_DIR"` on a path overridable via `TOKENTRACKER_LINUX_EMBED_DIR`, with no validation. It now refuses:

- relative or empty paths,
- `/`, the script directory, `TokenTrackerLinux/`, and the repo root,
- any path whose final component is not named `EmbeddedServer`,
- any path containing a symlinked component (so a link can't redirect the delete elsewhere).

A mistyped or unexpected override now fails loudly instead of deleting whatever it pointed at. Two smaller fixes in the same script: `TARGET_ARCH` is validated against the only supported value (`x64`) before anything downloads, and the icon sync moved below the `--clean` early-exit — cleaning no longer regenerates the icon it's cleaning up around.

**`sync-tauri-icon.mjs`** used a bare `writeFileSync`, which is non-atomic and writes *through* a symlink at the destination. It now writes to an exclusive (`wx`) temp file and renames it into place: the icon is replaced atomically, and a symlinked destination is replaced rather than followed.

## Tests

- Three new cases in `test/linux-bundle.test.js`: unsafe path refused, symlinked parent refused, bad `TARGET_ARCH` refused — each asserting a sentinel file survives. All verified to fail against the unhardened script.
- One new case in `TokenTrackerLinux/test/sync-tauri-icon.test.mjs`: a symlinked destination is replaced without writing through it, plus strengthened source assertions for the atomic write.
- Test temp dirs now go through `realpathSync`: macOS `os.tmpdir()` sits under the `/var → private/var` symlink, which the new guard would (correctly) refuse in the existing bundle test on the macOS CI job.

Behavior of the normal CI build path is unchanged — the workflow's plain `npm --prefix TokenTrackerLinux run bundle:node` invocation uses the default absolute, symlink-free path, and `--clean` has no CI callers.

Full `npm test`: 2492 tests, 0 failures. Guardrails validator clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux bundle safety by rejecting unsupported architectures and unsafe cleanup paths.
  * Prevented cleanup operations from following symlinks or removing protected locations.
  * Updated icon synchronization to safely replace files without overwriting symlink targets.
  * Ensured temporary files are removed even when icon synchronization encounters errors.

* **Tests**
  * Added coverage for unsafe architecture and cleanup path scenarios.
  * Added tests confirming symlink-safe icon replacement and temporary-file handling.

### DIFF
--- a/TokenTrackerLinux/scripts/bundle-node-linux.sh
+++ b/TokenTrackerLinux/scripts/bundle-node-linux.sh
@@ -16,8 +16,44 @@ EMBED_DIR="${TOKENTRACKER_LINUX_EMBED_DIR:-$LINUX_DIR/EmbeddedServer}"
 DASHBOARD_DIST="${TOKENTRACKER_DASHBOARD_DIST:-$REPO_ROOT/dashboard/dist}"
 TAURI_ICON="${TOKENTRACKER_TAURI_ICON:-$LINUX_DIR/src-tauri/icons/icon.png}"
 
-node "$LINUX_DIR/scripts/sync-tauri-icon.mjs" "$REPO_ROOT/dashboard/public/icon-512.png" "$TAURI_ICON"
 TARGET_ARCH="${TARGET_ARCH:-x64}"
+if [[ "$TARGET_ARCH" != "x64" ]]; then
+  echo "Refusing to bundle unsupported Linux architecture: $TARGET_ARCH" >&2
+  exit 1
+fi
+
+assert_safe_embed_dir() {
+  local normalized="$EMBED_DIR"
+  while [[ "$normalized" != "/" && "$normalized" == */ ]]; do
+    normalized="${normalized%/}"
+  done
+  if [[ -z "$normalized" || "$normalized" != /* ]]; then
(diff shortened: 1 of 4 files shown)


xiufengsun:
Merged. I checked out your test file against the unhardened scripts on `main` and all three new bundle cases fail there — unsafe path, symlinked parent, bad `TARGET_ARCH` — so the tests actually exercise what they assert rather than passing either way.

Also confirmed `--clean` has no callers (no CI step, no npm script), so moving the icon sync below the early exit changes nothing for the release build, which invokes `bundle:node` with no arguments.

The atomic icon write is the part I'd have wanted regardless of the path guards: a bare `writeFileSync` follows a symlinked destination, and `wx` + rename fixes both that and the torn-write window.

Thanks.

PR #535, opened 2026-08-28
feat(linux): report the linux app shell in telemetry heartbeats

csmashe:
## ⚠️ Deploy ordering

**The updated `tokentracker-telemetry` edge function must be deployed before this change ships in any Linux client build.** The current edge validator rejects `shell: "linux"` with HTTP 400 (`tokentracker-telemetry.ts:59`), so a client released first would have every Linux heartbeat discarded (silently, with hourly backoff — no user-facing harm, but no data). Deploying the edge first is fully backward-compatible: it only widens the allowlist.

## What

Every Linux install currently reports `shell=cli`, making Linux invisible in platform stats: the Linux app never sets `TOKENTRACKER_APP_SHELL`, and both allowlists (`VALID_SHELLS` in `src/lib/telemetry.js`, `SHELLS` in the edge patch) would map/reject the value anyway.

- **`TokenTrackerLinux/src-tauri/src/server.rs`** — new `node_command()` helper sets `TOKENTRACKER_APP_SHELL=linux` and is used by all three Node spawn sites: serve start, health-monitor restart, and the background sync. Marking the background sync matters on Linux specifically: unlike macOS (where sync comes from hooks and the marker rides only on the app-spawned serve), the Linux app spawns its own `sync` process, and both processes share the daily heartbeat throttle state — an unmarked sync that fires first would pin the day's row to `cli`. Hook-spawned syncs from a separate CLI install still report `cli` and are handled by the edge's existing upgrade rule.
- **`src/lib/telemetry.js`** — `VALID_SHELLS` accepts `linux` (previously `resolveShell` silently mapped it to `cli`).
- **`dashboard/edge-patches/tokentracker-telemetry.ts`** — `SHELLS` accepts `linux`; the upgrade-rule comment updated to reflect that app-spawned processes (not just serve) carry the marker.
- **Docs kept in sync** as the telemetry header requires: `docs/PRIVACY.md` and the shell list in all five READMEs (en/de/ja/ko/zh-CN).

## Tests

- New Rust unit test pins the env marker on `node_command()` via `Command::get_envs`.
- `resolveShell` test extended to assert `linux` is accepted.
- Rust: 56 tests pass, `clippy --all-targets` clean, `fmt --check` clean. JS: full `npm test` — 2484 pass, 0 fail.

## Summary by CodeRabbit

* **New Features**
  * Linux app sessions are now identified as `linux` in anonymous usage statistics.
  * Linux-launched Node processes consistently report the Linux app shell.

* **Documentation**
  * Updated privacy documentation and localized README files with clearer heartbeat data descriptions, including separate app version, platform, and shell fields.

* **Tests**
  * Added coverage confirming Linux shell detection and reporting.


xiufengsun:
Merged.

Checked the live `tokentracker-telemetry` function against `dashboard/edge-patches/` before merging — they were byte-identical, so there was no out-of-band drift to lose. Deploying the updated function now; without it the `SHELLS` check returns `400 invalid shell` and Linux app heartbeats would be rejected outright rather than falling back to `cli`, which is worse than the state you're fixing.

Correct on the privacy side: `shell` was already an existing field, so this adds a value rather than a dimension. Fixing `cli/mac/win` → `cli/macos/windows/linux` across the five READMEs and `PRIVACY.md` is a real correction on its own — those strings never matched what `resolveShell` actually accepts.

This ships with the next release, since it touches `src/` and `TokenTrackerLinux/`.

PR #537, opened 2026-08-28
perf(opencode): sync SQLite history incrementally

kvyb:
## Summary

Make OpenCode SQLite sync incremental so routine hook/background syncs no longer materialize the entire message history in memory.

The existing path read every assistant row on every sync. On a real 24 GB OpenCode database with 179,489 assistant rows, that full read took about 10 seconds and peaked at 630,767,616 bytes RSS. With a persisted database cursor, a resumed read returned only the timestamp-boundary row, took 3.81 seconds, and peaked at 55,246,848 bytes RSS (about 91% lower reader RSS).

The cursor is scoped to the database identity and to each v1/v2 table generation. It combines rowid and inclusive `time_updated` watermarks so appended rows, old-timestamp imports, in-place updates, and same-millisecond updates remain visible. An immutable row anchor forces a replay when a database is replaced or truncated in place, including equal/larger replacements; the read is bracketed by the same anchor to avoid advancing across a concurrent replacement.

This also keeps fork dedup memory proportional to the incremental batch rather than duplicating the complete historical fingerprint index. While exercising authoritative updates, the parser now handles cache-only rows, downward/mixed token corrections, stable conversation counts, and model/time/project bucket moves correctly.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Checklist

- [x] `npm test` passes
- [x] New user-facing strings go through `dashboard/src/content/copy.csv` (no new user-facing strings)

## Validation

- `npm test` — 2,491 passed, 2 skipped, 0 failed
- `node --test test/opencode2-parser.test.js test/opencode-fork-dedup.test.js test/multi-install-parser.test.js` — 35 passed
- `npm run validate:copy`
- `npm run validate:locale`
- `npm run validate:versions`
- `npm run validate:bot-frames`
- `node --test test/architecture-guardrails.test.js`
- `node --check src/lib/rollout.js`
- `node --check src/commands/sync.js`
- `git diff --check`

## Compatibility / edge cases covered

- v1, v2, and transitional OpenCode schemas
- new rows and old-timestamp imports
- in-place and same-millisecond updates
- cache-only usage
- downward and mixed-column corrections
- model/time bucket moves without conversation inflation
- path/inode database replacement
- same-file lower/equal/larger replacement and replacement races
- fork dedup across syncs
- per-install cursor namespace isolation

## Known trade-offs

- The first sync after upgrade, or a detected database replacement, intentionally performs one full replay to establish/repair the cursor safely.
- `time_updated` is not indexed in current OpenCode databases, so the resumed query still scans on disk; this change targets the much larger JS allocation/RSS cost. The measured resumed read was still faster (3.81s vs ~10s).

---

<details>

### Risk layer triggers

### Rules / invariants

- A cursor is persisted only after parsing succeeds.
- Database identity or generation-anchor mismatch causes replay, never silent advancement.
- Inclusive update boundaries prefer harmless rereads over missed same-millisecond updates.
- Fork copies remain deduplicated across incremental batches.

### Boundary matrix

- Append with newer rowid and timestamp → incremental read
- Append with old timestamp → rowid path still reads it
- In-place update at the saved timestamp → inclusive timestamp path rereads it
- Replaced/truncated DB → identity/head-anchor mismatch triggers full replay

</details>

## Summary by CodeRabbit

* **New Features**
  * OpenCode synchronization now processes only newly added or updated database records, improving sync efficiency.
  * Synchronization automatically detects database replacements or resets and replays affected data safely.
  * Updated messages can move between time periods, models, or projects without being double-counted.
  * Forked conversations are deduplicated more reliably, including across later synchronizations.

* **Bug Fixes**
  * Corrected token totals when message details change.
  * Preserved accurate conversation counts during message corrections and database resets.

### DIFF
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -26,6 +26,7 @@ const {
   listGeminiSessionFiles,
   listOpencodeMessageFiles,
   readOpencodeDbMessages,
+  readOpencodeDbMessagesIncremental,
   readMimoDbMessages,
   readZcodeDbMessages,
   hasZcodeNativeUsageSchema,
@@ -1164,11 +1165,15 @@ async function cmdSync(argv, context = {}) {
         let dbResult = { messagesProcessed: 0, eventsAggregated: 0, bucketsQueued: 0 };
         if (dbDir) {
           const dbPath = path.join(dbDir, "opencode.db");
-          const dbMessages = readOpencodeDbMessages(dbPath);
-          if (dbMessages.length > 0) {
+          const dbRead = readOpencodeDbMessagesIncremental(
+            dbPath,
+            cursors?.opencode?.dbCursor,
+          );
+          if (dbRead.messages.length > 0 || dbRead.cursor) {
             dbResult = await parseOpencodeDbIncremental({
               ...options,
-              dbMessages,
+              dbMessages: dbRead.messages,
+              dbCursor: dbRead.cursor,
               dbPath,
             });
           }
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -3285,6 +3285,7 @@ function normalizeOpencodeState(raw) {
   const messages = state.messages && typeof state.messages === "object" ? state.messages : {};
   return {
     messages,
+    dbCursor: state.dbCursor && typeof state.dbCursor === "object" ? state.dbCursor : null,
     updatedAt: typeof state.updatedAt === "string" ? state.updatedAt : null,
   };
 }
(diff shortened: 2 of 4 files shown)


xiufengsun:
Merged.

A persisted cursor plus a "replay once to establish it" path is exactly the shape that has cost this project real user data before — a bounded state paired with a full re-read of the source is how totals inflate — so I reconciled it against my own OpenCode database (1,817 messages, 1,559 conversations, 124,363,924 tokens) in four `HOME`-overridden sandboxes rather than reading the diff alone.

| Scenario | Result |
|---|---|
| `main`, fresh | 124,363,924 / conv 1,559 / 81 buckets |
| PR, fresh | identical |
| PR, second run | 0 files, 0 buckets, total unchanged |
| `main` first, then PR (upgrade) | replays 1,559 files, queues **0** buckets, total unchanged |
| PR, third run after upgrade | 0 files, 0 buckets |

The upgrade row is the one that matters. The cursor-establishing replay re-reads everything and queues nothing, so it overwrites its own buckets instead of adding to them — the failure mode I was looking for isn't there.

Then the incremental path itself, since none of the above exercises it: I dropped the 200 most recent messages, synced to establish a cursor against the truncated database (109,760,354 / 1,372 conv / 77 buckets), inserted the 200 rows back, and synced again. It read 188 files and landed on **124,363,924 / 1,559 conv / 81 buckets** — byte-identical to the full-read baseline, conversation count included. A resumed read that silently skipped or re-counted a boundary row would have shown up there.

The measurements in the description are the reason this was reviewable at all: 630MB → 55MB reader RSS with the read still correct is a real result, and stating the `time_updated` non-indexed trade-off up front instead of claiming a pure win is the right way to present it.

Ships with the next release.

PR #544, opened 2026-08-30
fix: unblock Copilot v2 migration for recovered buckets

FLC-niko:
## Problem

After #527, the Copilot cursor migration from v2 to v3 rebuilds historical Chat deduplication totals using both the legacy and corrected dedup keys.

Two valid states could leave the migration retrying forever:

1. Corrected deduplication recovers a request whose model/hour bucket exists only in the new totals. Validating the union of old and new buckets calls `copilotTotalsCover(undefined, oldUsage)` for that new bucket and aborts the migration.
2. A v2 cursor retains an offset for an OTEL file that has since been rotated or deleted. The path is no longer in the discovered file list, so the availability check aborts before the cursor can be persisted or new files can be processed.

In both cases the cursor remains at v2 and the same historical prefix is retried on every sync, leaving Copilot usage missing or stale.

## Fix

- Validate coverage only for buckets that contributed to the existing v2 totals, while retaining the union of old and new buckets for subtract/add reconciliation.
- Prune tracked file offsets whose paths are absent from the current discovered file list before deciding whether migration can proceed. Existing historical totals are preserved, and newly discovered files can advance the cursor to v3.

## Regression tests

Adds synthetic migration cases for:

- a recovered request creating a new model/hour bucket;
- a deleted old OTEL file alongside a newly discovered file.

Verified with:

`node --test test/rollout-parser.test.js` — 257 passed
`git diff --check` — passed

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Copilot usage tracking across native and WSL environments by recognizing additional telemetry locations.
  - Prevented duplicate Copilot activity from being counted and improved recovery when telemetry files are moved or removed.
  - Improved migration of existing Copilot usage data, including recovery into updated model categories.
  - Improved rollout metadata processing for unfinished requests, legacy records, network paths, and repeated ZCode usage updates.
  - Updated usage-data migrations for the latest Copilot and Grok formats.

### DIFF
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -14360,7 +14360,10 @@ async function migrateCopilotChatLogRecordDedup({
   }
 
   const keys = new Set([...oldTotals.keys(), ...newTotals.keys()]);
-  for (const key of keys) {
+  // Only keys that existed in the v2 contribution need coverage validation.
+  // A key that exists only in newTotals is a newly recovered Chat request; its
+  // old contribution is zero, so an absent old bucket is expected.
+  for (const key of oldTotals.keys()) {
     const [model, bucketStart] = JSON.parse(key);
     const oldUsage = oldTotals.get(key) || initTotals();
     if (!copilotTotalsCover(
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -14269,16 +14269,18 @@ async function migrateCopilotChatLogRecordDedup({
   touchedBuckets,
   seenIds,
 } = {}) {
+  // A v2 cursor can retain an offset for a rotated or deleted OTEL file that
+  // is no longer returned by discovery. Its historical contribution cannot be
+  // reconciled, but the stale offset must not block migration of new files.
+  const availableFiles = new Set(Array.isArray(files) ? files : []);
+  for (const filePath of Object.keys(fileOffsets || {})) {
+    if (!availableFiles.has(filePath)) delete fileOffsets[filePath];
+  }
   const trackedFiles = Object.keys(fileOffsets || {}).filter(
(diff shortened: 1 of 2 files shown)


xiufengsun (changes requested):
Exact head `6cbf2ce0d102467991f485f9332e0a338faeeaa8` still conflicts with current `main`, so the existing green checks do not represent an integrated result. Please rebase first and rerun CI on the resulting head.

There is also a migration correctness issue to address during the rebase: for a v2 Copilot cursor, `migrateCopilotChatLogRecordDedup` checks availability before removing offsets for tracked files that no longer exist. A deleted OTEL file can therefore keep the migration unavailable/repeating and prevent new files from being processed correctly. Prune missing tracked-file offsets before the availability decision, persist the upgraded cursor, and add a deleted-old-file plus new-file regression.

FLC-niko:
Addressed the requested migration edge case and rebased onto the current `main`.

- Prune v2 `fileOffsets` whose paths are no longer in the discovered file list before migration proceeds, so a deleted/rotated old file cannot block cursor persistence or processing of new files.
- Added a regression test covering a deleted old OTEL file alongside a newly discovered file.
- Preserved the existing v2-to-v3 recovered-bucket fix.

Validation:

- `node --test test/rollout-parser.test.js` — 257 passed
- All current PR checks pass, including Analyze, CodeQL, macOS, Linux, Windows, test/build, and CodeRabbit.

Updated head: `d6163739b2e04cafa3429468ffdd1ca690d96356`.

xiufengsun (approved):
Re-reviewed exact head d6163739b2e04cafa3429468ffdd1ca690d96356. The v2 migration now recovers newly discovered buckets and prunes deleted-file state without replaying CLI history. The rollout suite and all GitHub checks pass.

PR #553, opened 2026-08-31
feat(limits): fetch Antigravity quota via Google OAuth without the IDE

paveezheng:
## Summary

Antigravity remaining quota (Claude/Gemini 5h + weekly) now uses the same credential-and-remote-API path as Claude/Codex, so Limits stays current without launching the IDE.

Previously TokenTracker could only scrape the local Antigravity language server. Closing the app left a last-good cache that expired at each window reset, so quota disappeared after a few hours or days. This PR reads the on-disk Google token (`~/.gemini/jetski-standalone-oauth-token` or the `gemini`/`antigravity` Keychain item), refreshes it when expired, and calls Cloud Code `retrieveUserQuotaSummary` (daily host first). The local language server and last-good cache remain fallbacks when Cloud Code is unavailable. Gemini CLI `oauth_creds.json` is not treated as Antigravity credentials.

Verified live against a Google AI Pro account: Cl 7d/5h and Gm 7d/5h all returned without the IDE running.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

Dashboard change is copy/i18n only: live/cached tooltips no longer say the IDE must be running.

## Checklist

- [x] `npm test` passes

Targeted tests: `node --test test/usage-limits.test.js` (125 pass, 1 skip). Also `validate-locale-coverage` and `validate-copy-registry`.

---

<details>

### Risk layer triggers

- [x] Auth / session / token handling

### Rules / invariants

- Tokens stay local: read from Antigravity's existing files/Keychain, never logged, never uploaded.
- Refresh writes back only to the same on-disk file we loaded; Keychain entries are not written.
- Gemini CLI `oauth_creds.json` is not an Antigravity credential source.
- Prefer `daily-cloudcode-pa.googleapis.com`; the unprefixed host can report Gemini `remainingFraction=1`.
- Remote success is canonical; language-server scrape and last-good cache are fallbacks only.
- Cache windows still drop after `reset_at` / 7-day max age.

### Boundary matrix (list at least 3)

- Valid on-disk token, IDE closed → live Cl/Gm 5h+weekly from Cloud Code.
- Expired access token + refresh token → refresh, persist, then quota fetch.
- Cloud Code 5xx/timeout + IDE running → local language-server quota.
- Cloud Code unavailable + IDE closed + last-good cache still in window → cached bars.
- No Antigravity credentials and no install evidence → `configured: false`.

### Public exposure checklist (if applicable)

</details>

<details>

- **Delta since last Codex review:** Initial implementation on `feat/antigravity-remote-quota` (`7b0d8aca`).
- **Intended behavior / invariants:** Antigravity Limits uses local Google OAuth + `retrieveUserQuotaSummary` without the IDE; local RPC and last-good cache are fallbacks; tokens never leave the machine except to Google's OAuth/Cloud Code endpoints.
- **Edge cases covered:** Nested jetski token file, `antigravity-cli` token file, Keychain via `securityRunner`, ignore Gemini CLI creds, expired refresh, 401 retry, remote 503 → local RPC, remote 503 → cache, no creds/no process → not running.
- **Tests run (command + result):** `node --test test/usage-limits.test.js` — 125 pass, 1 skip. Live fetch on this machine returned Google AI Pro Cl/Gm windows with the IDE closed.
- **Known gaps / out of scope:** Windows Credential Manager `gemini:antigravity` is not implemented (macOS Keychain + files only). No change to transcript-based usage estimates.

- Limits card empty/error when Antigravity is closed (if remote path fails open).
- Wrong Gemini remaining % if the unprefixed Cloud Code host is used first.
- Gemini CLI quota card accidentally keyed off Antigravity tokens, or the reverse.
- Keychain prompt / failed `security` call blocking the limits poll on macOS.

- Live: `fetchAntigravityLimits()` with this machine's `jetski-standalone-oauth-token`, IDE closed — four windows returned from `daily-cloudcode-pa.googleapis.com`.
- Automated: `test/usage-limits.test.js` remote OAuth + cache/fallback cases.

### Uncovered scope

- Windows Credential Manager token discovery.
- End-to-end dashboard click-through against a running `tracker serve` (copy-only UI change).

</details>

## Summary by CodeRabbit

* **New Features**
  * Antigravity usage limits can now refresh from the associated Google account without requiring the Antigravity app or CLI to be open.
  * Usage information remains available through fallback methods when remote limit retrieval is unavailable.
  * Limit updates are handled within a defined timeout to improve responsiveness.
* **Localization**
  * Updated German, Japanese, Korean, Simplified Chinese, and Traditional Chinese tooltips to accurately describe live and cached limit refresh behavior.


xiufengsun (changes requested):
Exact head `7b0d8aca507b3a188ed5535bb3ab28e6d4e6420c` is not ready to merge.

The provider timeout is currently applied independently to the remote OAuth attempt and the local IDE fallback. When the remote path consumes its full budget, the fallback can consume another full budget, exceeding the caller's provider deadline. Split one total deadline between the remote attempt and the remaining local probe/request time, and add a regression where the remote attempt nearly exhausts the budget.

CodeQL is also failing on this head. Please resolve the failure and rerun CI/CodeQL on the updated exact head.

paveezheng:
Pushed `6522c7ff`.

**Provider timeout.** `providerTimeoutMs` arrived as a total budget but was used as a per-call timeout, so one 15s budget bought remote + `ps` + `lsof` + one per probed port + one per local RPC (~83s for a single port). `main` was already over budget — the local chain has never been bounded — so a remote/local split alone wouldn't have been enough.

The whole serial chain now shares one deadline, the way `fetchArkCodingPlanLimits` does: each step gets `min(step ceiling, remaining − guard)`, and a step with no budget left is skipped rather than issued. The guard is `min(1500ms, 15% of the budget)` so the disk-cache fallback still resolves inside the race. The call site is wrapped in `withAbortableProviderTimeout`, with `signal` threaded into the scans, local sockets and remote fetches.

Measured with a hanging Cloud Code host, the IDE running and 3 listening ports at a 1000ms budget: `main` 3005ms → `7b0d8aca` 4008ms → `6522c7ff` 855ms. 7 new regression tests, including "remote nearly exhausts the budget" (asserts the cache is still served, not an error); 1/3/10 ports all land at ~854ms. `node --test test/usage-limits.test.js` 133 pass, validators green.

**CodeQL 242** — fixed. The test file already had a `urlHostMatches()` helper for that rule; the new assertion uses it.

**CodeQL 243** (`persistAntigravityCredentials`) — needs your call, I can't dismiss alerts here. It flags the refresh response being written back to the credential file, which matches `codex-token-refresh.js` (required there — ChatGPT rotates the refresh token), Gemini and Kimi. This is the strictest of the four: atomic tmp + rename, `0600`, rotation-aware, file-sourced credentials only. Dismiss as a false positive, or say so and I'll drop the write-back and refresh per poll instead.

Left out on purpose, happy to open separately: a shared vendor-credential store (the Gemini and Kimi write-backs aren't atomic), and a meta-test asserting every provider in `getUsageLimits` is wrapped in a provider race.

xiufengsun (approved):
Re-reviewed exact head 6522c7ff35feb4c16061c4dd8a352baa7832d8c0. One provider deadline now bounds the complete remote/local Antigravity chain with abort-aware fallbacks; focused timeout tests and all GitHub checks pass.

PR #558, opened 2026-09-01
feat: attribute Codex usage per turn model

paveezheng:
## Summary

Codex `token_count` events carry no model id, so a session's usage was billed entirely to the last model seen in the file — this attributes every usage event to the model named by the most recent `turn_context`, which Codex emits at the start of every turn and which tracks mid-session model switches exactly.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Checklist

- [x] `npm test` passes

---

## Why

A session that switched models mid-thread had **all** of its tokens attributed to whichever model happened to be selected last. On a local history of 6617 Codex sessions, **501 (7.6%)** are multi-model; at the half-hour-bucket level it is **747 of 1213 codex buckets (61.5%)**. One measured session billed 52.1M tokens entirely to `gpt-5.6-terra` when 12.8M of them belonged to `gpt-5.6-sol` — a ~$5.6 undercount in a single session.

## What changed

**Per-turn attribution.** New `src/lib/codex-model-attribution.js` tracks the selected/effective model across a rollout; `codex-rollout-parser.js` accumulates a per-model usage table, and `rollout.js` persists the attribution state across incremental syncs. Sessions now carry `model_usage[]`; multi-model sessions report as `mixed` and are priced per model.

**Pricing corrections.** GPT-5.6 Sol moves to the official $4/$20 per MTok (from $5/$30), and the bare `gpt-5.6` alias now points at sol rather than terra. All 5 edge-function copies updated; `test/edge-pricing-parity.test.js` passes.

**Partial costs are labelled.** A session whose model has no public rate (mainly `codex-auto-review` — 1733 sessions, 27.5% of local sessions) rendered a bare `≥` in the cost column with nothing explaining it, because the badge that says so was gated behind `isGrok && usage_precision`. That gate now covers any partial cost, with its own copy key (`Partial cost` is a distinct claim from `Partial usage`: every token is observed, only the rate is missing) plus a tooltip. Both keys are localized in zh / zh-TW / ja / ko / de.

**Single-pass Codex scanning.** Delivery-signal collection folds into the token parser's pass via an `onObject` callback instead of reading every rollout file twice.

**Sidecar trimmed.** `SIDECAR_VERSION` 12 → 13. The sidecar is re-read on every dashboard request and a third of it was fields `normalizeModelUsageRows()` reconstructs anyway (zero counters, a `selected_models` that repeats the model, the default attribution, a cost recomputed on read). Omitting them is lossless: 11.9 MB → 9.4 MB on a 6816-session history.

## Two paths are forward-looking, and say so in code

- **`model/rerouted` attribution** — no such event exists in any of **5832** local rollouts (codex-cli 0.151.0). Exact-key search for `"model/rerouted"` and `"model_rerouted"` returns zero files; every loose match is conversation content. The envelope shapes and `fromModel`/`toModel` names are inferred from the app-server protocol, not confirmed against logged data.
- **The 272K long-context tier** — unreachable while Codex reports `model_context_window = 258400` (max raw input observed: 238853). `long_context_*` stays at zero and the repricing never fires.

Also documented: `event_msg/thread_settings_applied` is deliberately **not** used for attribution. It carries `thread_settings.model` but fires when the user picks a model in the UI, seconds to minutes before the switch takes effect, so usage still streaming from the in-flight turn belongs to the previous model. Across 600 rollouts, 12 files / 66 usage events fall in that window. `turn_context` is the only signal that marks where a turn actually begins.

## Verification

`npm run ci:local` passes end to end (exit 0):

```
npm test                 2519 pass / 0 fail  (108 suites)
validate:copy            1536 entries ok
validate:locale          zh + zh-TW 100%
validate:ui-hardcode     ok
validate:guardrails      ok
validate:versions        all match v0.94.1
validate:bot-frames      ok
architecture-guardrails  4 pass
dashboard build          ok

dashboard vitest         675 pass / 96 files
tsc --noEmit             clean
```

Against real logs (6.6 GB, 5832 rollouts), in an isolated `HOME` so no upload happened and no local state was touched:

- Cold sidecar rebuild 22.7s → 6816 sessions; warm read 54ms, summarize 35ms.
- Cold build and warm re-read agree to the cent (**$16135.25** both), and `model_usage` round-trips field for field.
- **Sync stress test, three consecutive runs** (per the repo's data-migration rule): run 1 cold 29.9s / 394 files / 80 buckets; runs 2 and 3 both 1.5s with **0 brand-new buckets**, the only delta being the live half-hour bucket still being written.
- **Three incremental runs == one from-scratch cold sync, exactly**: 2428/2428 settled buckets identical, 20,694,070,852 tokens on both sides. No state pollution.

## Reviewer notes

- **`SIDECAR_VERSION` 12 → 13 forces a full rebuild** on first run after upgrade (~23s on a 6.6 GB history; larger histories will take longer).
- **The 5 edge functions must be deployed** after merge (account-daily / account-summary / account-model-breakdown / leaderboard-profile / leaderboard-refresh).
- **No version bump here.** This touches `src/` and `dashboard/`, so per CLAUDE.md it needs a version bump plus the full npm + DMG + Windows + Linux release; left to the release owner.
- **`by_model.sessions` deliberately overlaps.** A session contributes a row to every model it used, so summing that column overstates the total (7330 vs 6816 locally) while tokens, cost and edit turns reconcile exactly. Nothing sums it today; a comment and a test now pin both halves.
- **4.8% of tokens remain unpriced** — `codex-auto-review` (870M) and `unknown` (473M). Honest rather than silently $0, but still uncovered.

## Summary by CodeRabbit

* **New Features**
  * Session analytics now show usage, tokens, edits, and costs broken down by model, including rerouted models.
  * Added support for tracking usage from Qoder, LM Studio, and Unsloth Studio.
  * Added long-context pricing adjustments for GPT-5.6 usage.
  * Sessions with unavailable model pricing now display clear lower-bound cost indicators.
* **Bug Fixes**
  * Corrected GPT-5.6 Sol pricing and default model mapping.
  * Improved model attribution and ignored invalid zero-total usage records.
* **Localization**
  * Added partial-cost labels and explanations in German, Japanese, Korean, Simplified Chinese, and Traditional Chinese.

### DIFF
--- a/src/lib/session-analytics.js
+++ b/src/lib/session-analytics.js
@@ -30,7 +30,7 @@ const path = require("node:path");
 const readline = require("node:readline");
 const { listClaudeProjectFiles, listRolloutFilesDeep, claudeMessageDedupKey } = require("./rollout");
 const { parseCodexRolloutFile } = require("./codex-rollout-parser");
-const { computeRowCost } = require("./pricing");
+const { computeRowCost, getModelPricing } = require("./pricing");
 const { USD_TICKS_PER_USD, normalizeGrokUsage } = require("./grok-usage");
 const wsl = require("./wsl-probe");
 
@@ -57,9 +57,10 @@ const wsl = require("./wsl-probe");
 // their union, deduping shared records while retaining divergent tails.
 // v12 fixes Grok's mutually exclusive token columns, retains exact reported
 // cost, and adds metadata-only usage diagnostics to the local session browser.
-// v12 retains Codex parent/subagent metadata locally and derives exact child
-// token totals from observed child rollouts instead of spawn-call estimates.
-const SIDECAR_VERSION = 12;
+// v13 stores per-model Codex usage (including observed reroutes and the exact
+// long-context subset) and folds delivery signals into the token parser's
+// single pass instead of parsing every Codex file twice.
+const SIDECAR_VERSION = 13;
 const EDIT_TOOLS = new Set([
   "apply_patch",
   "edit",
@@ -116,6 +117,113 @@ function addTotals(target, delta) {
 
 function emptyTotals() {
   return tokenTotals({});
+}
+
+const MODEL_USAGE_SUM_FIELDS = [
+  "input_tokens",
+  "cached_input_tokens",
+  "cache_creation_input_tokens",
+  "output_tokens",
+  "reasoning_output_tokens",
+  "total_tokens",
+  "long_context_input_tokens",
+  "long_context_cached_input_tokens",
+  "long_context_cache_creation_input_tokens",
+  "long_context_output_tokens",
+  "long_context_reasoning_output_tokens",
+  "usage_events",
+  "rerouted_usage_events",
+  "long_context_usage_events",
+];
+
+function normalizeModelUsageRows(rows) {
+  const byModel = new Map();
+  for (const value of Array.isArray(rows) ? rows : []) {
+    const model = normalizeSessionModel(value?.model) || "unknown";
+    let row = byModel.get(model);
+    if (!row) {
+      row = {
+        model,
+        ...Object.fromEntries(MODEL_USAGE_SUM_FIELDS.map((field) => [field, 0])),
+        selected_models: new Set(),
+        reroute_reasons: new Set(),
+        model_attribution: "selected",
+      };
+      byModel.set(model, row);
+    }
+    for (const field of MODEL_USAGE_SUM_FIELDS) row[field] += finite(value?.[field]);
+    for (const selected of Array.isArray(value?.selected_models) ? value.selected_models : []) {
+      const normalized = normalizeSessionModel(selected);
+      if (normalized) row.selected_models.add(normalized);
+    }
+    for (const reason of Array.isArray(value?.reroute_reasons) ? value.reroute_reasons : []) {
+      if (typeof reason === "string" && reason.trim()) row.reroute_reasons.add(reason.trim());
+    }
+    if (value?.model_attribution === "effective" || finite(value?.rerouted_usage_events) > 0) {
+      row.model_attribution = "effective";
+    }
+  }
+  return [...byModel.values()]
+    .map((row) => ({
+      ...row,
+      selected_models: [...row.selected_models].sort(),
+      reroute_reasons: [...row.reroute_reasons].sort(),
+    }))
+    .sort((a, b) => b.total_tokens - a.total_tokens || a.model.localeCompare(b.model));
+}
+
+function repriceSessionRecord(record) {
+  const modelUsage = normalizeModelUsageRows(record?.model_usage);
+  if (modelUsage.length > 0) {
+    let hasUnpricedUsage = false;
+    for (const row of modelUsage) {
+      row.cost_usd = computeRowCost({ source: record.source, ...row });
+      if (record.source === "codex" && row.total_tokens > 0) {
+        const pricing = getModelPricing(row.model, { source: record.source });
+        if (![pricing.input, pricing.output, pricing.cache_read, pricing.cache_write]
+          .some((value) => finite(value) > 0)) {
+          hasUnpricedUsage = true;
+        }
+      }
+    }
+    record.model_usage = modelUsage;
+    record.model = modelUsage.length > 1 ? "mixed" : modelUsage[0].model;
+    record.cost_usd = modelUsage.reduce((sum, row) => sum + finite(row.cost_usd), 0);
+    record.cost_source = "model_pricing";
+    // Internal labels such as `codex-auto-review` expose token usage but not
+    // an underlying public model/rate. Keep their tokens, leave their cost at
+    // zero, and explicitly mark the session estimate as a known lower bound.
+    record.cost_is_partial = hasUnpricedUsage;
+  } else {
+    const providerCost = Number(record.provider_cost_usd);
+    record.cost_usd = record.cost_source === "provider_reported"
+      && Number.isFinite(providerCost)
+      && providerCost >= 0
+      ? providerCost
+      : computeRowCost({ source: record.source, model: record.model, ...record.tokens });
+  }
+  record.cost_per_edit = record.edit_turns > 0 ? record.cost_usd / record.edit_turns : null;
+  return record;
+}
+
+function modelUsageForAggregation(record) {
+  const rows = normalizeModelUsageRows(record?.model_usage);
+  if (rows.length > 0) {
+    return rows.map((row) => ({
+      ...row,
+      cost_usd: Number.isFinite(Number(row.cost_usd))
+        ? Number(row.cost_usd)
+        : computeRowCost({ source: record.source, ...row }),
+    }));
+  }
+  return [{
+    model: normalizeSessionModel(record?.model) || "unknown",
+    total_tokens: finite(record?.total_tokens || record?.tokens?.total_tokens),
+    cost_usd: finite(record?.cost_usd),
+    selected_models: [],
+    reroute_reasons: [],
+    model_attribution: "selected",
+  }];
 }
 
 function safeTimestamp(value) {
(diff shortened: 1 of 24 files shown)


xiufengsun (changes requested):
I rechecked exact head `ebf900559b340b02f742311bdf7b2262fb7a8fe7`. The feature direction is useful, but the attribution/consumer contract is not safe to merge yet:

- `applyCodexModelEvent` must read `session_meta.payload.model` when no `turn_context` model is available; otherwise early usage is attributed to the fallback model.
- `SessionModelUsage` / `SessionRow` currently require fields that `modelUsageForAggregation` does not always emit, and `model_usage` must remain optional for older local servers. Align the TypeScript contract with the producer rather than asserting fields into existence.
- Hourly and session parsing normalize `total_tokens` differently when reported totals disagree with component sums, so the same rollout can show different totals across views.
- Mixed-model `tokens_per_edit` attributes all edit turns to the primary model but only part of the tokens, producing a misleading ratio.

Please fix these contracts and add fixtures covering a session-meta-only initial model, an older response without `model_usage`, and a mixed-model session before requesting another review.

paveezheng:
Pushed 5f4ebaa4, 4c4aeb16, 1417551b. `ci:local` green.

**Fixed**
- `tokens_per_edit` — edit turns now split per model, so numerator and denominator match. `sum(by_model.edit_tokens)` reconciles with `summary.edit_tokens` again. Retries stay session-level: `user_message` / `turn_context` ordering isn't stable (1332 vs 1139 across 400 rollouts).
- TS contract — fixed on the producer side. `modelUsageForAggregation` emits a dense row on both paths, so the type is true without `?? 0` in every consumer. `model_usage` optional for the older-server skew only.
- German umlauts (`zahlen` → `zählen` also inverted the meaning).

Fixtures added for a mixed-model session and a response without `model_usage`, plus a contract test that both producer paths emit the same key set.

**`session_meta.payload.model` doesn't exist** — 0 of 10,310 `session_meta` lines across 5,849 local rollouts; the payload carries `model_provider`. The session-meta-only fixture isn't writable, and reading the field changes nothing. The `unknown` attribution is real but caused by fork replay: all 102 sessions with non-zero `unknown` are forked, and their pre-`turn_context` events are exact duplicates of the parent's. `parseRolloutIncremental` already excludes those (`isForkedReplayToken`), `parseCodexRolloutFile` doesn't — own PR, since it moves ~1.85% of session tokens.

**`total_tokens` normalization** predates this branch — both `normalizeUsage` copies are byte-identical to `main`. 175M tokens (0.67%) counted hourly and dropped by the session view. Unifying them also changes headline numbers, so also its own PR.

Happy to open both.

xiufengsun (changes requested):
I re-reviewed the exact current head `1417551b47872251d413eefc1454eb6843549f05`. The dense API contract and per-model edit-turn changes are improvements, but two attribution blockers remain:

1. `applyCodexModelEvent()` ignores the model carried by `session_meta.payload.model`. Sessions without a later `turn_context` model can therefore remain attributed to `unknown` even though authoritative model metadata was present.
2. A `model_rerouted` event updates token attribution but not the active delivery collector's `currentTurnModel`. Edits made after a mid-turn reroute can consequently be charged to the old model.

Please update both state machines consistently and add regressions for session-meta-only model attribution and a mid-turn reroute followed by edit events.

xiufengsun (changes requested):
I re-reviewed the exact current head `1417551b47872251d413eefc1454eb6843549f05`. The dense API contract and per-model edit-turn changes are improvements, but two attribution blockers remain:

1. `applyCodexModelEvent()` ignores the model carried by `session_meta.payload.model`. Sessions without a later `turn_context` model can therefore remain attributed to `unknown` even though authoritative model metadata was present.
2. A `model_rerouted` event updates token attribution but not the active delivery collector's `currentTurnModel`. Edits made after a mid-turn reroute can consequently be charged to the old model.

Please update both state machines consistently and add regressions for session-meta-only model attribution and a mid-turn reroute followed by edit events.

paveezheng:
Pushed `e049b1ab`, rebased onto `main` (`0f7066c8`). `ci:local` green.

**2. `model_rerouted` → `currentTurnModel`** — confirmed and fixed. `currentTurnModel` was only ever written in `beginTurn()`, so a reroute event fell through `consume()` untouched and post-reroute edits stayed on the old model:

```
before:   terra 3300 tokens / 0 edit turns    sol 1100 / 1
after:    terra 3300 tokens / 1 edit turn     sol 1100 / 0
```

The collector now follows the model the parser hands it whenever that model moves inside an open turn. The sync sits *after* the `turn_context` branch on purpose — before `beginTurn()` it would retag the previous turn at the moment `closeTurn()` settles it.

**1. `session_meta.payload.model`** — fallback added, as a strictly weaker signal: fills `selectedModel` only when nothing else has, never overrides a `turn_context` model, never lets a forked rollout's replayed parent metadata drag attribution backwards. The field itself is still unobserved (0 of 10,310 `session_meta` rows across 5,849 local rollouts; the payload carries `model_provider`), so the fixtures are synthetic by necessity and the comments say so.

It does not fix the `unknown` attribution, though: all 102 sessions with non-zero `unknown` are forked, and `parseCodexRolloutFile` lacks the `isForkedReplayToken` filter `parseRolloutIncremental` already has. That one, and the hourly/session `total_tokens` normalization gap (predates this branch — both `normalizeUsage` copies are byte-identical to `main`), stay as their own PRs.

Three new tests, all failing without the fixes.

src/lib/session-analytics.js
```diff
@@ -1569,6 +1824,7 @@ function mergeSessionFragments(rows) {
     row.duration_ms = finite(row.active_ms);
     row.first_pass = finite(row.edit_turns) === 1 && finite(row.retry_turns) === 0;
     row.one_shot = row.first_pass;
+    repriceSessionRecord(row);
```
xiufengsun, line 1827: [P1] Preserve already-priced Claude fragment costs when no model_usage exists. This unconditional reprice reaches the fallback branch for Claude and bills all merged tokens at `row.model`; the removed busiest-fragment model assignment also leaves that model dependent on input order. A scanned Haiku 4.5 fragment with 100k input tokens plus a Sonnet 4.5 fragment with 1M input tokens should total $3.10. Current head returns $1.10 (Haiku first) or $3.30 (Sonnet first), whereas main returns $3.10 in both orders. Retain the summed cost and legacy representative-model behavior for such records, or preserve complete per-model fragment usage before repricing, and add both order permutations as regressions.

xiufengsun (changes requested):
Re-reviewed exact head `e049b1ab8c3cdcff852547d4f881212c96cb5011` against current main `6a5685d208e478dd55931a1acb494b2f46f9b48b`; integration is conflict-free. The session-meta fallback, mid-turn reroute/edit fix, dense API contract, and per-model edit regressions now pass. I also checked the Sol rates against the current official pricing page. The pre-existing fork-replay and total-normalization gaps can stay separate; please open the follow-up PRs you proposed.

One newly reproduced regression still blocks merge: `mergeSessionFragments()` now reprices Claude fragments using the first fragment's model, replacing their already-correct summed costs. These Claude records do not have `model_usage`.

Reproduction uses two synthetic Claude files with the same `sessionId`, distinct message IDs, no message content, and input-only usage: 100,000 Haiku 4.5 tokens ($0.10) plus 1,000,000 Sonnet 4.5 tokens ($3.00). Parse both with `scanClaudeSession()`, then call `listSessionsForBrowser()` in both input orders:

| Result | main | this head |
| --- | --- | --- |
| Small fragment first | Sonnet, $3.10 | Haiku, $1.10 |
| Large fragment first | Sonnet, $3.10 | Sonnet, $3.30 |

`summarizeSessions()` still reports $3.10, so the browser disagrees with the session summary. Both regression cases pass on main and fail on the integrated PR.

Please preserve per-fragment cost sums and the busiest-fragment representative model for records without per-model usage, or construct complete per-fragment model rows before repricing. Add a scan-based regression covering both orders and browser/summary cost equality. Existing focused validation is green (359 Node tests and 10 SessionsPage tests); it does not cover this mixed-Claude-fragment case.

xiufengsun:
Thanks for the contribution! We have taken over the final integration and merged this PR, preserving your commits.

The Claude fragment-cost regression from the latest review is fixed in 48403b6b: legacy Claude fragment costs remain additive, while per-model repricing is limited to sessions with attributed model usage. Both fragment orders now retain the correct $3.10 total. We also added a three-pass full-source Codex sync regression to check queue stability.

Local validation passed: 2,637 Node tests (2 skipped), 694 dashboard tests, 162 macOS tests, plus the newly added locale-validator regression and all release validators. The exact v0.96.0 commit is now running through cross-platform CI; the version is not yet published. No further author action is needed for this PR.

xiufengsun:
Released in [v0.96.0](https://github.com/xiufengsun/TokenTracker/releases/tag/v0.96.0). The exact release commit passed cross-platform CI; npm and all six macOS/Windows/Linux artifacts are available. All five cloud pricing functions are deployed and match the release source. The published npm package was also downloaded and verified against the maintainer fixes. Thanks again!

PR #561, opened 2026-09-01
fix: stop a transient cloud failure from downgrading account views to one device

paveezheng:
## Summary

A cloud read that merely *failed* was indistinguishable from "this user only has one device", so a single timed-out account request silently replaced the native popover's cross-device Activity heatmap with this-machine data until the next manual sync.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Why

`?account=1` answers with local single-machine data in two completely different situations:

1. that really is the user's scope — signed out, or cloud sync off;
2. the cloud read failed — timeout, offline, rejected token refresh.

Both came back as an indistinguishable `HTTP 200` + `X-TokenTracker-Account-View: 0`, and the popover's heatmap / daily / monthly / model-breakdown fetches went through the generic `fetch()`, which dropped the header entirely. So case 2 overwrote a good account snapshot with one device's data.

An amplifier made it fire more often than it should: a full refresh issues six account reads concurrently, and on a cold token cache every one of them POSTed `/api/auth/refresh` with the same refresh token. When the backend rotates refresh tokens, the losers of that race present an already-consumed token and fail — inside a 4s budget shared by the token mint *and* the account fetch.

Reproduces on any flaky network; wake-from-sleep just makes it likely, because the wake catch-up fires before Wi-Fi/VPN is back.

## What changed

**`src/lib/local-api.js`** — classify the fallback (`signed-out`, `cloud-sync-off`, `transient-timeout|auth|network|upstream`) and report it in a new `X-TokenTracker-Account-Fallback` header. Every transient reason carries the `transient-` prefix, so new reasons need no client change.

**`src/lib/cloud-account.js`** — share one in-flight `/api/auth/refresh` per `(baseUrl, refreshToken)`. Also throw a classified error instead of returning `null` when a *signed-in* refresh fails, so it can no longer be mistaken for "not signed in".

**`TokenTrackerBar/`** — every account-capable fetch returns `AccountFetchResult<T>` carrying the response authority. `AccountViewStateStore` refuses to let a transient fallback replace an account snapshot that is already on screen (full load and hidden menu-bar refresh alike), retries on a 1s/3s/10s backoff, and logs the reason only. Cold start with no snapshot still shows local data, labelled "This Mac only · reconnecting".

**`TokenTrackerWin/`** — the tray poller had the same defect: it stored `X-TokenTracker-Account-View` in an `AccountViewActive` field that **nothing in the assembly read**, then published the payload regardless. Net deletion plus one guard — on a transient downgrade, skip the publish and let `TrayApplicationContext._lastStats` keep rendering the account snapshot. No new types and no cached copies, because `UsageStats` is published atomically.

## Behaviour matrix

| Situation | Before | After |
|---|---|---|
| Cloud read times out, account snapshot on screen | Replaced with this-machine data | Snapshot kept, retried on backoff |
| User signs out / turns cloud sync off | Switches to local | Switches to local (unchanged) |
| Cold start, cloud already down | Local data, unlabelled | Local data, labelled as temporary |
| Server too old to send the reason header | — | Treated as authoritative-local (pre-fix behaviour) |

## Checklist

- [x] `npm test` passes — 2526 passed. The 5 failures in my sandbox are environment-only (this worktree has no `node_modules`, so the vitest / `tsc` / vite-dev-server tests can't resolve their deps); none touch this change.
- [x] New user-facing strings go through the copy registry — n/a for `dashboard/`; the one new string is a macOS `Strings.swift` entry with all five locales filled in.

`TokenTrackerBar/project.yml` is untouched (sources are directory-based, so the new Swift file is picked up by `xcodegen generate` with no project edit).

**Maintainer note:** this touches `src/`, `TokenTrackerBar/` and `TokenTrackerWin/`, so per `CLAUDE.md` it is release-eligible. I did not bump any version file — that is the release workflow's call.

---

<details>

### Risk layer triggers

- [x] Auth / session / token handling

### Rules / invariants

- A `transient-*` fallback must never replace a payload whose recorded authority is `account`.
- An authoritative local view (signed out, cloud sync off) must *always* replace an account payload — otherwise a signed-out user keeps seeing another session's totals.
- With no account snapshot to keep, local data is still published; an empty panel is worse.
- Concurrent account reads sharing a `(baseUrl, refreshToken)` must produce exactly one refresh request.
- A rotated refresh token reaches every waiter of that shared flight.
- The reason string is diagnostic only: no tokens, cookies, or usage figures cross the boundary or reach logs.

### Boundary matrix

| Boundary | Input | Expected |
|---|---|---|
| `X-TokenTracker-Account-View` absent (pre-0.46 server) | no headers | authoritative-local; pre-fix behaviour, no throw on the popover's summary path |
| `Account-View: 0`, no reason header | older server | authoritative-local, not transient |
| `Account-View: 0`, unknown `transient-xyz` | future server | transient, matched by prefix |
| Refresh returns 401 (token consumed) | rotation race | `transient-auth`, **not** `signed-out` |
| Cloud unreachable (`TypeError: fetch failed`) | offline | `transient-network`, body still local data, HTTP 200 |
| Failed in-flight mint | first caller throws | entry cleared; the next call retries rather than inheriting the failure |
| Dataset emptied by a period switch (`hourly`/`monthly`) | period change | degraded flag cleared, so retry budget resets |

### Regression tests

- `test/local-api-account-view.test.js` — six classification cases + "one popover refresh mints ONE token across concurrent reads"
- `test/cloud-account.test.js` — single-flight, failure doesn't poison later attempts, error classification, signed-in refresh failure throws
- `test/macos-account-view-authority.test.js`, `test/windows-account-view-authority.test.js` — no Swift/C# test target exists, so the invariants are asserted against the source

</details>

<details>
<summary><strong>Verification</strong></summary>

**Executed, not just asserted:**

- The Swift state machine is Foundation-only, so it was compiled standalone and run against the reported scenario plus the reverse case (sign-out must still switch), cold start, an old server with no reason header, and the degraded-flag reset — all pass.
- A real `http.createServer` round-trip (the existing tests use a fake `res`, which can't prove `writeHead` doesn't drop `setHeader` values) confirms both headers arrive over a socket and the body is still local data rather than an error:

  ```
  HTTP/1.1 200 OK
  X-TokenTracker-Account-View: 0
  X-TokenTracker-Account-Fallback: transient-network
  ```

**Compiled on both platforms:**

- macOS: Xcode 27.0 beta (27A5252f), Swift 6.4, `xcodebuild -configuration Release` → `** BUILD SUCCEEDED **`. No warnings from any changed file (the one warning in the log is the pre-existing `Copy EmbeddedServer` script-phase notice). Verified the new type actually reached the product binary:

  ```
  TokenTracker.AccountViewStateStore.shouldAdopt(
      _: TokenTracker.AccountViewSource,
      for: TokenTracker.AccountViewStateStore.Dataset,
      hasExistingValue: Swift.Bool) -> Swift.Bool
  ```

- Windows: .NET SDK 10.0.400, `dotnet build -p:EnableWindowsTargeting=true` (cross-compiled from macOS), Debug and Release → **0 warnings, 0 errors** in both.

`TokenTrackerBar/project.yml` needed no edit; `xcodegen generate` picks up the new Swift file from the directory-based source list.

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented temporary local-only data from replacing trusted account data during cloud sync interruptions.
  * Prevented stale data from a previous date or period from overriding current results.
  * Added automatic recovery retries when account data is temporarily unavailable.
  * Improved refresh authentication errors and handling of concurrent token refreshes.
  * Preserved dashboard and menu-bar values during temporary account request failures.

* **User Experience**
  * Activity heatmaps now indicate when they show data from this Mac only, with localized messaging.

* **Tests**
  * Added coverage for account authority, fallback states, recovery, and authentication failures.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Models/AccountViewSource.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Models/AccountViewSource.swift
@@ -81,7 +81,34 @@ struct AccountViewStateStore {
         case modelBreakdown
     }
 
-    private var sourceByDataset: [Dataset: AccountViewSource] = [:]
+    /// Identifies the query a dataset's authority was recorded for.
+    ///
+    /// Authority is only meaningful for the exact scope it was observed on.
+    /// `.periodSummary` / `.modelBreakdown` / `.monthly` follow the selected
+    /// period, and the day-scoped views follow the calendar day. Without this,
+    /// a retained account snapshot from the previous period outranks a fresh
+    /// response for the new one, and the panel keeps last period's numbers
+    /// under this period's label.
+    ///
+    /// Windows that merely slide (rolling 30d, all-time, the year heatmap) use
+    /// a constant scope on purpose: there the one-day shift is a rounding
+    /// difference, and rekeying them at midnight would hand a transient cloud
+    /// failure a fresh chance to downgrade an account view to this machine.
+    enum Scope {
+        static let rolling30 = "rolling30"
+        static let total = "total"
+        static let daily30 = "daily30"
+        static let heatmap = "heatmap"
+        static func day(_ day: String) -> String { "day:\(day)" }
+        static func range(_ from: String, _ to: String) -> String { "range:\(from)..\(to)" }
+    }
+
+    private struct Record {
+        let source: AccountViewSource
+        let scope: String
+    }
+
+    private var recordByDataset: [Dataset: Record] = [:]
     private(set) var degradedDatasets: Set<Dataset> = []
 
     /// True while at least one dataset is running on (or holding onto data
@@ -92,32 +119,42 @@ struct AccountViewStateStore {
     /// because the cloud is unreachable — i.e. cold start with no account
     /// snapshot to fall back on.
     func showsTransientLocalData(_ dataset: Dataset) -> Bool {
-        sourceByDataset[dataset]?.isTransientFallback ?? false
+        recordByDataset[dataset]?.source.isTransientFallback ?? false
     }
 
     /// Decide whether an incoming payload should replace what is on screen.
+    /// - Parameter scope: identifies the query this response answers. A record
+    ///   held for a different scope describes data the view no longer asks for.
     /// - Parameter hasExistingValue: whether the view model already holds a
     ///   rendered payload for this dataset.
     /// - Returns: true to publish the new payload, false to keep the old one.
     mutating func shouldAdopt(
         _ source: AccountViewSource,
         for dataset: Dataset,
+        scope: String,
         hasExistingValue: Bool
     ) -> Bool {
+        // Drop authority recorded for a scope the view has since left, so the
+        // retained-snapshot rule below can never answer for the wrong period.
+        if let record = recordByDataset[dataset], record.scope != scope {
+            recordByDataset.removeValue(forKey: dataset)
+            degradedDatasets.remove(dataset)
+        }
         guard source.isTransientFallback else {
             degradedDatasets.remove(dataset)
-            sourceByDataset[dataset] = source
+            recordByDataset[dataset] = Record(source: source, scope: scope)
             return true
         }
         degradedDatasets.insert(dataset)
-        if hasExistingValue, sourceByDataset[dataset]?.isAccount == true {
+        if hasExistingValue, recordByDataset[dataset]?.source.isAccount == true {
             // Keep the account snapshot and its authority; this failure is
             // temporary and a retry is scheduled.
             return false
         }
-        // Nothing better to show (cold start, or we were already local):
-        // this-machine data beats an empty panel, but stays marked degraded.
-        sourceByDataset[dataset] = source
+        // Nothing better to show (cold start, a scope we have no account
+        // snapshot for, or we were already local): this-machine data beats an
+        // empty panel, but stays marked degraded.
+        recordByDataset[dataset] = Record(source: source, scope: scope)
         return true
     }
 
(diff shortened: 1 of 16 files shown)


xiufengsun (changes requested):
Thanks for the thorough authority model and regression coverage. I rechecked exact head `2474f75a459e90a9bfb4fed1c40296d89f600c9b`; two transition races still need fixes before merge:

1. On macOS, `sourceByDataset` outlives the query scope. When the selected period/day changes, `.periodSummary`, `.modelBreakdown`, and `.hourly` can still be marked as account-authoritative for the previous scope. A transient response for the new scope is then rejected because `hasExistingValue` and the stale authority are both true, leaving old-period data on screen. Clear the affected dataset authority/degraded state when the period or selected day changes, before evaluating the new response, and add scope-change regressions.
2. On Windows, the rich-stat helpers retain local-transient results only when `_showingAccountData` was already true. During a local-to-account transition, the current summary can be account-authoritative while that previous-state flag is still false, so a transient heatmap/model response is accepted and mixed into the account `UsageStats`. Pass `summarySource == AccountSource.Account || _showingAccountData` into both helper decisions and cover this transition.

Please address both authority transitions and rerun the cross-platform tests on the new exact head.

paveezheng:
Both addressed in 276381f3.

1. **macOS** — instead of clearing on switch, `AccountViewStateStore` now records
   the query scope alongside each dataset's authority and discards a record whose
   scope the view has left. That covers the period switch and the day rollover
   with one rule. Sliding windows (rolling 30d, all-time, heatmap) keep a constant
   scope on purpose, so midnight isn't a fresh opening for a transient downgrade.

2. **Windows** — both rich-stat helpers now take
   `summarySource == AccountSource.Account || _showingAccountData`.

Regressions added for both. `TokenTrackerWin.Tests` turned out to be usable here
(net8.0, already run by CI), so the Windows transition is covered end-to-end
against a loopback server rather than by source assertions.

PR #565, opened 2026-09-02
fix(macos): keep limit-reset baseline when a reading has no reset_at

### LINKED ISSUE #564: Limit-reset celebration misses Claude rollovers: nil resetAt reading overwrites the percent baseline
## Summary

The limit-reset celebration (toast + confetti) fires reliably for Kimi but usually misses Claude 5h / 7d rollovers. Root cause is in `WeeklyLimitResetDetector.evaluate`: a reading whose `resetAt` is `nil` still overwrites `lastPercent[key]`, which wipes the baseline the next rollover needs.

## Repro / evidence (macOS TokenTrackerBar, v0.94.x)

`defaults export com.tokentracker.bar -` → `WeeklyLimitResetSnapshot` on my machine:

```
claude.5h        pct=36  resetAt=09-02 20:49  lastEvent=09-02 11:33   (fired once — I happened to be mid-session at rollover)
kimi.secondary   pct=0   resetAt=09-02 22:15  lastEvent=09-02 17:15   (fires every rollover)
```

Kimi returns a `reset_at` even when the window is empty (pct=0 with a future resetAt above), so every rollover satisfies "resetAt advanced + drop ≥ 5". Claude's 5h window, once it has rolled over and there is no active session, reports `utilization: 0` with `resets_at: null` (`ClaudeWindow.resetsAt` is already optional for this reason).

## Mechanism

In `evaluate`:

```swift
updated.lastPercent[key] = reading.usedPercent          // ← runs unconditionally
if let resetAt = reading.resetAt { updated.lastResetAt[key] = resetAt }
guard let prevPercent, let prevReset, let curReset = reading.resetAt else { continue }
```

Timeline for Claude 5h:

1. Poll A (before rollover): `pct=60, resetAt=T1` → baseline 60 / T1.
2. Poll B (after rollover, idle): `pct=0, resetAt=nil` → `lastPercent` becomes **0**, `lastResetAt` stays T1, no event (correct so far).
3. Poll C (user starts a new session): `pct=3, resetAt=T2` → `curReset > prevReset` ✓, but `prevPercent - used = 0 - 3 < minDrop` ✗ → no celebration.

The celebration is only produced when Poll B never happens, i.e. the first poll after rollover already sees a fresh `resets_at` — which requires being mid-session at the exact rollover moment. That matches the observed behaviour (Claude fires rarely, Kimi always).

## Proposed fix

Treat a reading without `resetAt` as "no information about this window" and keep the previous baseline instead of overwriting it:

```swift
guard let curReset = reading.resetAt else { continue }   // don't touch lastPercent / lastResetAt
updated.lastPercent[key] = reading.usedPercent
updated.lastResetAt[key] = curReset
```

(Move the two `updated.*` assignments below the `resetAt` guard.) Existing tests in `WeeklyLimitResetDetectorTests` should keep passing; a new case "nil resetAt between two readings does not suppress the following rollover" would lock it in. Happy to open a PR if you'd like.

arvakme:
## Summary

Skip limit-window readings that carry no `reset_at` before touching the persisted baseline, so Claude's idle `0% / null` reading after rollover no longer wipes the percent needed to celebrate the next reset. Closes #564.

## Scope

- [ ] CLI (`src/`)
- [ ] Docs / CI / config

## Checklist

- [x] `npm test` passes — all pass except `installLocalTrackerApp replaces stale installed runtime…`, which needs a prebuilt `dashboard/dist` locally and is unrelated (no JS touched here)
- [x] New user-facing strings go through `dashboard/src/content/copy.csv` — n/a, no strings

## Why

`WeeklyLimitResetDetector.evaluate` wrote `lastPercent[key] = usedPercent` for every reading, including ones with `resetAt == nil`. Claude's 5h window sitting idle after rollover reports `utilization: 0, resets_at: null`, so the baseline became 0; the next real reading (new `resets_at`, low usage) then failed the `minDrop` check and the toast/confetti never fired. Kimi returns a `reset_at` even for an empty window, which is why it celebrated every rollover while Claude almost never did (snapshot evidence in #564).

## Behaviour

- Reading with `resetAt == nil` → baseline untouched, no event (it still never celebrates by itself, as before).
- Reading with `resetAt` → baseline updated and rollover detection unchanged.

## Test

Added `testIdleReadingWithoutResetTimeKeepsBaseline`: 60% → (0%, nil) → (3%, new reset) fires with `previousPercent == 60`. No Xcode on this machine, so the detector struct was compiled standalone with `swiftc` and run through every scenario in `WeeklyLimitResetDetectorTests` including the new one (all pass); CI should run the real XCTest target.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Models/WeeklyLimitResetDetector.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Models/WeeklyLimitResetDetector.swift
@@ -93,15 +93,19 @@ struct WeeklyLimitResetDetector {
 
         for reading in readings {
             let key = reading.windowKey
+            // A reading without a reset timestamp (e.g. Claude's 5h window sitting idle
+            // after rollover reports 0% / null) carries no window identity, so it must
+            // not touch the baseline: overwriting the percent with 0 here would make the
+            // next real reading fail the minDrop check and silently lose the celebration.
+            guard let curReset = reading.resetAt else { continue }
             let prevPercent = snapshot.lastPercent[key]
             let prevReset = snapshot.lastResetAt[key]
             updated.lastPercent[key] = reading.usedPercent
-            if let resetAt = reading.resetAt { updated.lastResetAt[key] = resetAt }   // keep last value when missing
+            updated.lastResetAt[key] = curReset
 
-            // Need a full prior baseline (percent + reset time) and a current reset
-            // time. First observation, or a window without a reset timestamp, only
+            // Need a full prior baseline (percent + reset time). First observation only
             // records a baseline — never celebrates.
-            guard let prevPercent, let prevReset, let curReset = reading.resetAt else { continue }
+            guard let prevPercent, let prevReset else { continue }
             // The window must have actually rolled over: its reset_at advanced to a
             // new period, not merely a percentage that dipped.
             guard curReset > prevReset + resetAdvanceTolerance else { continue }
--- a/TokenTrackerBar/TokenTrackerBarTests/WeeklyLimitResetDetectorTests.swift
+++ b/TokenTrackerBar/TokenTrackerBarTests/WeeklyLimitResetDetectorTests.swift
@@ -58,6 +58,20 @@ final class WeeklyLimitResetDetectorTests: XCTestCase {
         XCTAssertTrue(events.isEmpty)
     }
 
+    func testIdleReadingWithoutResetTimeKeepsBaseline() {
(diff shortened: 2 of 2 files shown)


xiufengsun (approved):
Reviewed exact head 07563382f2ad52537b9a2d4a4a1d517630fe291d. A reading without reset_at now leaves both the reset timestamp and percentage baseline unchanged, with regression coverage; all GitHub checks pass.

PR #569, opened 2026-09-03
fix(copilot): track VS Code custom endpoint chat usage

FLC-niko:
## Problem

VS Code Copilot Chat requests routed through a custom OpenAI-compatible endpoint can persist only in workspaceStorage chatSessions files. Existing Copilot readers do not collect these records, so their usage is missing.

These records report aggregate input without a reliable fresh/cache split. The PR introduces a disjoint `unclassified_input_tokens` column and partial-cost responses to retain those tokens without billing all input as uncached.

## Changes

- Discover VS Code Stable, Insiders, and VSCodium sessions on macOS, Windows, and Linux; parse JSON snapshots and JSONL patches for `customendpoint/*` requests.
- Reconcile corrections and legacy cursors, bound retained request metadata, and preserve repeated-sync idempotency. Official Copilot requests stay on the existing readers.
- Carry unclassified input through local accounting, dashboard/native displays, cloud ingestion, account aggregations, and leaderboard rollups.
- Add the SQL migration and edge changes required by the new column. The sync capability check retains queued data until the cloud supports accounting version 3.
- Preserve partial cost in leaderboard rows and profile total/daily cost; complete zero-cost profiles still render zero.
- Keep parser fixture writes on a descriptor opened before parser reads, and isolate the local API test from the user's home directory.

## Validation on 4c686d581c350a457abda961954ca1c5ab6aa83e

- Parser suite: 277 passed, including six VS Code Copilot scenarios.
- Parser → pricing → real local API regression: 1 passed.
- Cloud contract and migration suite: 13 passed, none skipped. Executed the optional migration test against temporary local PGlite, including deduplication, session corrections, rollups, and function access checks. No deployed database was modified.
- Leaderboard page/profile tests: 18 passed. The partial-cost profile regression failed before the rendering fix.
- All final-head GitHub CI checks and CodeQL passed; the current PR merge-ref has zero open CodeQL alerts.

## Remaining before merge and release

- `src/lib/outcomes-engine.js:computeQualityPerDollar` still uses `computeRowCost(row) || 0`. A synthetic row with 1,000,000 unclassified input tokens and 100 output tokens has full cost `null` and known cost `$0.005`, but the quality-per-dollar response reports `cost_usd: 0` and `effective_cost_usd: 0`. Propagate incomplete cost through this consumer and its UI, including mixed complete/partial rows; do not turn unknown spend into zero or a complete denominator.
- Reconcile and validate against the latest main branch before merge; the full accounting change has not yet passed maintainer integration review.
- Before a client release, apply the migration and deploy the edge functions in the documented order, then verify the live ingestion capability and account/leaderboard responses. The local migration test does not establish production readiness.


xiufengsun (changes requested):
I re-reviewed the exact current head `f76cc09bd7b9c7805429dfb1ca361ff8cbc9dcc6`. The platform-test fix is present, but these accounting and state-safety blockers remain:

1. `extractVsCodeCopilotUsage()` classifies every `promptTokens` value as uncached input and sets cached input to zero without provider evidence. That can materially overstate cost; please preserve only semantics proven by the source (or leave the split unknown) and add billing regression coverage.
2. Snapshot change detection relies on size/mtime, so a same-size rewrite within the filesystem timestamp granularity can be skipped. Use a durable content/version identity or a conservative rescan path.
3. Deleted or moved chat-session entries remain in the cursor indefinitely. Prune confirmed missing files while retaining state for transient read errors.
4. The cursor persists full request arrays, and incremental JSONL reads still load the whole file. Please bound persisted state and memory use to the overlap/identity data required for deduplication.

Please push a new head with focused tests for billing semantics, same-size rewrites, deletion pruning, and bounded growth.

xiufengsun (changes requested):
The input_split_unknown marker is emitted, but pricing and the dashboard do not consume it. Reproduced this head with a customendpoint/vendor/gpt-6-astra request containing promptTokens=1,000,000 and completionTokens=100: the queue has total_tokens=1,000,100 but the component sum is 100, and computeRowCost returns only $0.005 for the output. This silently presents an incomplete amount as a normal cost and violates the queue normalization contract.

Please preserve the reported aggregate input through an explicit supported representation and carry the unknown/partial-cost state through local API, dashboard and cloud consumers, or defer ingestion until the required accounting semantics are supported. Do not guess a cache split or silently zero the input. Add an end-to-end assertion through the real cost/API consumer, beyond parser-only tests.

The CodeQL failure is in the same-size rewrite test's stat/write/stat fixture, not a confirmed production exploit; it still needs a narrowly scoped test fix. The bounded cursor and rewrite handling improvements are acknowledged.

FLC-niko:
Pushed cb14d62b with the requested accounting and state-safety fixes.

- Copilot custom-endpoint prompt totals now use an explicit unclassified_input_tokens column. The parser preserves exact total usage, does not guess fresh versus cached input, and reports partial known cost instead of overstating billing.
- JSONL cursors now use durable content hashes, conservative legacy-cursor rescans, bounded request overlap, compact file aggregates, and confirmed-missing-file pruning while retaining transient read failures.
- Added focused coverage for billing semantics, same-size rewrites, legacy cursor migration, bounded cursor growth, deletion pruning, ingest capability negotiation, device merge preservation, rollups, and total shards.

Local validation: Copilot parser 6/6, full rollout parser 277/277, cloud accounting 13/13 including PGlite migration execution, dashboard build, and related pricing/backend/rollup checks 96/96. GitHub Actions is now running on this head.

FLC-niko:
Follow-up f8424d6e removes the test fixture TOCTOU pattern that CodeQL flagged as a high-severity file-system race. The production parser behavior is unchanged; the same-size rewrite test now uses the persisted cursor identity for its size/mtime assertions. Focused Copilot parser tests remain 6/6 passing.

FLC-niko:
Follow-up 4aeffd37 addresses the remaining CodeQL file-system-race annotation in the late-patch regression fixture. The append now uses a pre-opened file handle, so the test models an append without a path-based check-then-use sequence. Focused parser tests remain 6/6 passing.

xiufengsun:
The parser work here is good, and you have been responsive through three rounds, so I want to be direct about where this stands rather than send it round a fourth time.

**The scope has outgrown the problem.** The diff is now 47 files and 3,581 lines, including a 1,216-line SQL migration, changes to nine edge functions, three Swift files, three Windows files, and a change to `computeRowCost`'s return type. That is a schema-level change to how TokenTracker accounts for tokens, arriving inside a PR titled "track VS Code custom endpoint chat usage". I am not willing to merge the accounting change on the strength of the parser being useful.

Three reasons, then a proposal.

**1. It breaks the token invariant without updating it.** CLAUDE.md states `total_tokens = input + output + cache_creation + cache_read + reasoning_output`, and that file loads into every session. With `unclassified_input_tokens` the sum no longer holds, and the PR does not touch CLAUDE.md. A documented invariant that is false is worse than a restrictive one.

**2. One partial row blanks the cost for everything.** `computeRowCost` now returns `null` whenever a row carries unclassified input, and `UsageOverview.jsx:337` does `allCostPartial = providers.some(p => p.costPartial)`. So a single custom endpoint request turns the dashboard total into "Cost incomplete" for every provider, and `estimated_cost_usd` goes null for the user's whole leaderboard period. Showing the known cost with a "≥" marker is defensible; erasing the number is a product decision I would rather make myself.

There is also a worse case downstream. `ActivityHeatmap.jsx:378-384` falls back to `totalTokens / 1500000` when `total_cost_usd` is not usable, so the local heatmap now shows a fabricated estimate instead of the real cost. A null that some consumers read as "unknown" and others silently replace with an invented number is the part that worries me most about this design.

**3. The upload path fails closed for everyone.** `sync.js:3840` throws when the cloud reports `accounting_version < 3`. The queue drains in offset order, so one row with unclassified input blocks every later row, including other providers'. Edge deployment here is manual and separate from releases, so the client would ship before the functions land and those users' cloud sync would freeze, with no visible error on the `--auto` path.

**Two data bugs independent of all that**, both reproducible:

`rollout.js:17164` prunes `fileStates[filePath]` on confirmed-missing but deliberately keeps the bucket tokens. When the file comes back there is no baseline left, so it replays in full: 5,895 before deleting, 5,895 after deleting (correct), 11,265 after restoring. iCloud-synced home directories, VS Code profile switches and restores from backup all reach this.

`rollout.js:16861`'s reconciliation only protects a `(model, bucket)` pair that disappears entirely. If the JSONL is compacted but the bucket still has one request left, the new value overwrites the old one: 1,200 tokens becomes 770. VS Code lets you delete a single turn, so ordinary use reaches this. The JSON snapshot branch has the right guard; the JSONL reset branch is missing it.

**What I would merge.** Please split this. A first PR that is parser-only: read the custom endpoint JSONL, fold the prompt total into `input_tokens`, record the precision caveat on the row, no new column and no null cost. It under-reports the cache split for those rows, which is a known and bounded inaccuracy, and it ships behind no migration. Then the accounting column can be discussed on its own merits, with the cross-platform story settled up front. Right now macOS decodes `unclassified_input_tokens` and never renders it, Windows does not decode it at all, and five dashboard token-key lists were not updated (`details.ts`, `daily.ts`, `use-trend-data.ts`, `TrendMonitor.jsx`, `ProjectDetailModal.jsx`), so composition bars and stacked charts will not add up to their own totals.

Before either lands I would also like evidence for the premise at `rollout.js:16147`, that the existing OTEL parser does not already see custom endpoint requests. `rollout.js:15373` says the OTEL reader covers the VS Code Chat extension's LogRecords. If both readers see the same request, this double counts. A redacted `~/.copilot/otel/*.jsonl` showing a custom endpoint turn absent from it would settle it.

Two more things riding along that should not be in a Copilot PR: `TokenFormatter.swift:77` changes `formatCostFromString` to return "—" instead of `$0.00` on a parse failure, which affects the menu bar cost in every state including first paint, and `WidgetSnapshot.swift` changes `costUsd` to `Double?`, which an older widget process cannot decode.

Last thing, and it is not a criticism of the work: the executable migration test at `test/cloud-unclassified-input.test.js:109` is gated behind `TOKENTRACKER_TEST_PGLITE`, so CI only regex-matches the SQL text. A 1,216-line migration that has never actually run is not something I can merge on review alone.

PR #573, opened 2026-09-03
fix(linux): open external links in the system browser

csmashe:
## Problem

On the Linux client, clicking an external link inside the app window does nothing at all — no popup, no browser, no error. Most visible on the **Service Status** page, where every provider card is a link to that provider's status page, but it affects every `target="_blank"` link in the dashboard (leaderboard profiles, the sidebar GitHub link, docs links).

Cause: wry only connects WebKitGTK's `create` signal when a new-window handler is registered (`wry/src/webkitgtk/mod.rs`), and the Tauri window in `main.rs` registered none. WebKitGTK's default `create` returns no webview, so the click is silently dropped.

The other two desktop clients already handle this explicitly:

- macOS — `WKUIDelegate.createWebViewWith` → `NSWorkspace.shared.open` (`DashboardWindowController.swift:507`)
- Windows — `CoreWebView2.NewWindowRequested` → `OpenInBrowser` (`DashboardWindow.cs:364`)

## Fix

New `src-tauri/src/external.rs`:

- `is_internal_url()` — loopback hosts, `*.localhost`, and non-`http(s)` schemes (`tauri://`, `about:`, `blob:`) stay in the window. Port-agnostic, since the bundled server picks its port dynamically. Lookalike hosts such as `https://127.0.0.1.evil.com/` are treated as external.
- `open_in_browser()` — hands the URL to `xdg-open`, matching how `oauth.rs` already launches the browser.

Wired into the window builder:

- `.on_new_window(...)` → open externally, `NewWindowResponse::Deny`. This is the reported bug.
- `.on_navigation(...)` → mirrors the Windows navigation guard. The app window has no browser chrome, so a top-level navigation off the dashboard would otherwise strand the user with no way back.

## Testing

- `cargo test` — all suites pass, including the new `tests/external.rs` and the unit tests in `external.rs`.
- `cargo fmt --check` clean.
- Ran the app on Fedora / GNOME Wayland and clicked through the Service Status cards: each opens the provider's status page in the default browser and the app window stays put.

## Summary by CodeRabbit

- **New Features**
  - External links now open in the system browser on Linux.
  - Internal dashboard links, loopback addresses, and supported application URLs remain inside the app.
  - Links opened in new windows and navigation away from the dashboard are routed appropriately.

- **Bug Fixes**
  - Prevents third-party pages and potentially misleading domains from opening within the app.

- **Tests**
  - Added coverage for internal dashboard navigation, provider status pages, and external domains.

### DIFF
--- a/TokenTrackerLinux/src-tauri/src/external.rs
+++ b/TokenTrackerLinux/src-tauri/src/external.rs
@@ -1,0 +1,88 @@
+//! Routing for links that should leave the app window.
+//!
+//! The dashboard renders external links as `target="_blank"` (service status
+//! cards, leaderboard profiles, GitHub links). WebKitGTK's default `create`
+//! handler returns no webview, so without an explicit handler those clicks are
+//! a silent no-op — the macOS (`WKUIDelegate.createWebViewWith`) and Windows
+//! (`CoreWebView2.NewWindowRequested`) clients both hand the URL to the system
+//! browser instead, and this is the Linux counterpart.
+
+use std::process::Command;
+
+use tauri::Url;
+
+/// URLs the dashboard is allowed to load inside the app window.
+///
+/// Only `http`/`https` can escape to the browser: the loading screen is served
+/// over Tauri's own `tauri://` scheme and the dashboard over the loopback
+/// server, and any other scheme (`about:`, `blob:`, `data:`) is webview
+/// bookkeeping that must not be handed to `xdg-open`.
+pub fn is_internal_url(url: &Url) -> bool {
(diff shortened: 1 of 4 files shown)


xiufengsun (approved):
Reviewed at exact head 0fa6ab55065ff0fb70999c0b697af7499db75f54 and rechecked with merge-tree against current main after #574. The handlers match the desktop clients' external-link behavior, preserve loopback/Tauri navigation, reject lookalike hosts, and the exact-head Linux/full CI and CodeQL checks are green.

PR #582, opened 2026-09-05
fix(sync): preserve ZCode history that repairs cannot re-replicate

LceAn:
## Why

The ZCode queue repair migrations (`zcodeNativeUsageRepair_2026_08`, `zcodeInclusiveTokenRepair_2026_09`) rewrite every removed `zcode` bucket key with a **zero retraction**, then re-parse the source DB and re-queue whatever it still holds.

That is lossy for real users: ZCode itself prunes old sessions from `db.sqlite` over time, so the fresh parse cannot always re-replicate what was originally parsed. Because consumers keep the latest row per `(source, model, hour_start)`, the zero retractions permanently erased real uploaded history — both in the local dedup view and on the cloud after the replay — for every key the fresh parse could not cover.

Measured on a real multi-week install: June GLM-5.2 buckets held ~402M tokens after the `2026_08` repair rebuilt them from a then-complete DB, while the same DB now returns ~323M for those rows (ZCode pruning). A further repair pass would zero those buckets outright. On multi-device accounts the per-device repairs made account totals visibly shrink after each device updated.

## What

- `dropZcodeQueueRows` now writes back the **last uploaded row per bucket key verbatim** (including project rows) instead of zero retractions. Keys the fresh parse covers are overwritten by the new values as usual; keys it cannot cover keep their last known value. Nothing is zeroed.
- The migration record logs `preservedRows` instead of `retractions`.
- Tests updated to assert preservation, plus a regression test proving: covered keys yield to fresh values, uncovered keys keep history, and no zero rows are emitted.

Follow-up to #545, which fixed the parser side of ZCode usage; this fixes the destructive side of the repair migration itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * ZCode usage synchronization now preserves the latest available cloud usage when historical sessions cannot be re-created.
  * Freshly re-parsed usage takes precedence where available, while unrecovered usage remains intact instead of being reset to zero.
  * The same preservation behavior now applies across main and project usage queues.

### DIFF
--- a/src/commands/sync.js
+++ b/src/commands/sync.js
@@ -4283,16 +4283,22 @@ async function resetUploadOffsetForZcodeRepair(queueStatePath, note) {
   return true;
 }
 
-async function dropZcodeQueueRows(filePath, { retainRetractions = false } = {}) {
+// Drops zcode rows ahead of a from-scratch re-parse. The re-parse can only
+// cover rows still present in the source DB, and ZCode prunes old sessions —
+// so keys the fresh parse cannot re-replicate must keep their last uploaded
+// value. Write those rows back verbatim instead of zero retractions: a
+// replay of zero rows permanently erases real history on the cloud (latest
+// per bucket key wins) for every key the fresh parse does not cover.
+async function dropZcodeQueueRows(filePath, { preserveLatestRows = false } = {}) {
   let raw;
   try {
     raw = await fs.readFile(filePath, "utf8");
   } catch (error) {
-    if (error?.code === "ENOENT") return { removed: 0, retractions: 0 };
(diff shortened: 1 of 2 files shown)


LceAn:
Closing this in favor of #584.

#584 establishes that the retroactive rewrite is the intended correction (the pre-repair values double-counted cache/reasoning tokens), so preserving pre-repair values for keys the fresh parse cannot re-replicate would keep known-inflated data — the opposite of what's needed. The actionable part (surfacing the rewrite in release notes / the UI) lives in #584.

Thanks — keeping the branch around in case any of the analysis is useful.

PR #587, opened 2026-09-06
fix(macos): recover usage limits after a system clock rollback

lie5860:
## Summary

I ran into this edge case while working on an agent scheduling feature and temporarily setting the macOS system time into the future. After restoring the correct time, TokenTracker's native menu-bar usage limits continued showing an old Codex percentage and reset time, even though the local endpoint and dashboard were already returning current data.

This is not required for normal operation, but refresh and relaunch did not recover the native view without manually clearing the cache. This PR proposes a small defensive fix in case the maintainers think the edge case is worth handling.

## Why it happens

A response fetched while the clock is in the future can be persisted by URLCache with a future timestamp. Once the clock is moved back, URLSession may continue treating that response as fresh. The last-good usage-limits snapshot in UserDefaults can also be future-dated, so it may be shown again at launch.

## Suggested change

- Bypass URLCache for the macOS client's dynamic localhost API reads.
- Ignore a persisted last-good usage-limits response when its fetched_at value is more than five minutes ahead of the current time.
- Add regression coverage for both the URLCache policy and the clock-rollback case.

## Scope

- [ ] CLI (`src/`)
- [ ] Docs / CI / config

## Verification

- `npm test` — 2,639 passed, 0 failed, 2 skipped
- `node --test test/macos-usage-limits-timeout.test.js` — 3 passed
- Swift type-check of `UsageLimits.swift` and the new cache-loading path

## Checklist

- [x] `npm test` passes
- [x] No new user-facing strings were added
- [x] Commit follows the repository's conventional style
- [x] The description explains both the observed behavior and why the proposed fix helps

Thanks for taking a look. I'm happy to adjust the approach if you would prefer a narrower cache policy or a different clock-skew threshold.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented stale or future-dated usage-limit data from being displayed after system clock changes.
  - Ensured usage-limit requests retrieve fresh data instead of relying on locally cached responses.
  - Improved handling of invalid or unreadable cached data.

- **Tests**
  - Added coverage for clock rollbacks, future-dated cached responses, and cache-bypass behavior.

### DIFF
--- a/TokenTrackerBar/TokenTrackerBar/Models/UsageLimits.swift
+++ b/TokenTrackerBar/TokenTrackerBar/Models/UsageLimits.swift
@@ -111,10 +111,21 @@ extension UsageLimitsResponse {
 
 enum UsageLimitsCache {
     static let defaultsKey = "UsageLimitsLastGoodResponse"
+    private static let maximumFutureSkew: TimeInterval = 5 * 60
 
-    static func load(defaults: UserDefaults = .standard) -> UsageLimitsResponse? {
+    static func load(
+        defaults: UserDefaults = .standard,
+        now: Date = Date()
+    ) -> UsageLimitsResponse? {
         guard let data = defaults.data(forKey: defaultsKey) else { return nil }
-        return try? JSONDecoder().decode(UsageLimitsResponse.self, from: data)
+        guard let limits = try? JSONDecoder().decode(UsageLimitsResponse.self, from: data) else {
+            return nil
+        }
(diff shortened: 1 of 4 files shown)


xiufengsun:
Merged through #589 with your commits preserved. I added a runtime URLSessionConfiguration test alongside the clock-rollback cache regression. The combined macOS suite and cross-platform CI pass. No further author changes are needed; this is not in a new desktop release yet. Thank you!

PR #588, opened 2026-09-07
fix(antigravity): read sqlite context size without inferring cache hits

obsesivegamer:
## Why
Antigravity's local SQLite metadata records prompt context size and model name. It does not record cache reads or output counts. Inferring cache hits understated cost.

## Scope
`parseAntigravityFile` in `src/lib/rollout.js` bills SQLite context growth as `input_tokens`, leaves `cached_input_tokens` at 0, and resets the snapshot on a model switch. Tests in `test/rollout-parser.test.js` cover repeated context, model switch, incremental vs full scan, and a legacy estimated cursor.

## Tradeoffs
Input from SQLite is still a session-incremental estimate, not per-request billed tokens. That matches the existing Antigravity heuristic. Without a cache counter, a true cold cache mid-session still cannot be distinguished.

## Blast radius
Only the Antigravity parser and its tests. Other providers are untouched. The dashboard Estimated tokens notice is unchanged.

## Verification
`node --test --test-name-pattern 'Antigravity' test/rollout-parser.test.js` reported 19 passed.

## Summary by CodeRabbit

- **New Features**
  - Antigravity usage tracking now incorporates metadata from associated conversation databases.
  - Usage data identifies whether values are estimated or derived from stored metadata.
  - Incremental conversations can resume using database-backed progress.

- **Bug Fixes**
  - Corrected token counts for repeated context, incremental conversations, and model switches.
  - Prevented cache-hit tokens from being incorrectly inferred or added to total usage.
  - Improved reconciliation when transitioning from estimated usage to database-derived usage.

### DIFF
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -18722,6 +18722,113 @@ async function parseAntigravityIncremental({
   return { filesProcessed, eventsAggregated, bucketsQueued, projectBucketsQueued };
 }
 
+function decodeAntigravityVarint(buf, offset) {
+  let res = 0;
+  let shift = 0;
+  while (offset < buf.length) {
+    const b = buf[offset++];
+    res += (b & 0x7f) * 2 ** shift;
+    if (!(b & 0x80)) break;
+    shift += 7;
+  }
+  return [res, offset];
+}
+
+function findAntigravityProtoFields(buf) {
+  const fields = [];
+  let offset = 0;
+  while (offset < buf.length) {
+    const [tag, next] = decodeAntigravityVarint(buf, offset);
+    offset = next;
+    const fieldNum = tag >> 3;
+    const wireType = tag & 7;
+    if (wireType === 0) {
+      const [val, vNext] = decodeAntigravityVarint(buf, offset);
+      offset = vNext;
+      fields.push({ num: fieldNum, val });
+    } else if (wireType === 2) {
+      const [len, lNext] = decodeAntigravityVarint(buf, offset);
+      offset = lNext;
+      fields.push({ num: fieldNum, val: buf.subarray(offset, offset + len) });
+      offset += len;
+    } else if (wireType === 1) {
+      offset += 8;
+    } else if (wireType === 5) {
+      offset += 4;
+    } else {
+      break;
+    }
+  }
+  return fields;
+}
+
+function extractAntigravityGenInfo(buf) {
+  const root = findAntigravityProtoFields(buf);
+  const f1 = root.find((f) => f.num === 1)?.val;
+  if (!f1) return null;
+
+  const inner = findAntigravityProtoFields(f1);
+  let model = null;
+  const f19 = inner.find((f) => f.num === 19)?.val;
+  if (f19) model = Buffer.from(f19).toString("utf8").trim();
+
+  let contextTokens = 0;
+  const f9 = inner.find((f) => f.num === 9)?.val;
+  if (f9) {
+    const f10 = findAntigravityProtoFields(f9).find((f) => f.num === 10)?.val;
+    if (f10) {
+      const tok = findAntigravityProtoFields(f10).find((f) => f.num === 1)?.val;
+      if (Number.isFinite(tok)) contextTokens = tok;
+    }
+  }
+
+  let lastStepIndex = null;
+  for (const f of inner) {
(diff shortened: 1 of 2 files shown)


xiufengsun (changes requested):
Reviewed exact head `16f45d480c83532bdc8baa2a4125e7c70ecb9f3a`. The 15 focused Antigravity parser tests pass, but the new accounting cannot be treated as accurate usage yet.

In `src/lib/rollout.js`, `extractAntigravityGenInfo()` reads a context-size field, while `parseAntigravityFile()` invents `cached_input_tokens = min(curTokens, previousContextTokens)` and charges only the growth as fresh input. A repeated context is not evidence of a cache hit: a cold cache, expiration, a model switch, or changed prefix can bill the entire input. The synthetic SQLite test contains no cache-hit counter, yet it produces 25,000 cached tokens. This would materially understate cost for those cases.

Output and reasoning also still come from `antigravityValueTokens(content/thinking)`, so those remain text-length estimates. Please provide a redacted, counts-only first-party metadata sample/schema establishing per-request input/output/cache counters and the protobuf field meanings, and read those counters directly. Until such fields are available, preserve the existing explicit estimation status without presenting inferred cache savings as measured usage. Add cold-cache/model-switch and incremental-versus-full-scan regressions, including the transition from the existing estimated cursor.

No prompts, responses, credentials, or full databases are needed. Keeping this PR open pending accounting evidence.

obsesivegamer:
Thanks for the review. 
You are right, repeated local context is not evidence of a server cache hit, and output+reasoning is still a text-length estimate.

 I checked the local Antigravity databases.

I decoded 1,323 `gen_metadata` rows across 15 files in `~/.gemini/antigravity/conversations/*.db`. Counts and field numbers only. No prompts, responses, or credentials.

Redacted protobuf map under the Field 1 payload:

```text
1.9.10.1  varint   prompt context size
                   present on every row
                   observed range 172 .. 255975
1.9.10.4  varint   model context window
                   present on every row
                   values 128000 or 256000
1.9.10.3  message  rare prompt-section breakdown (15/1323 rows)
                   labels: System Prompt, Tools, Chat Messages
                   per-section sizes only
                   not a billed cache or output counter
1.11      message  timing
                   1.11.2 looks like request duration in microseconds
1.19      string   model id
                   examples: gemini-3.8-flash, claude-sonnet-4-6
1.20      pairs    request_id, last_step_index, model_enum,
                   trajectory_id, used_claude, used_non_gemini_model
```

Field 1.2 is the planner payload (bot id, response text, a local length). It is not a server usage record. There is no cache-read counter and no billed output or reasoning counter in this schema.

This PR now follows that evidence:

- `1.9.10.1` supplies the prompt context size for input deltas.
- `1.19` supplies the model name.
- `cached_input_tokens` stays 0.
- Output and reasoning stay transcript text-length estimates.
- The existing Estimated tokens notice stays.
- Tests cover repeated context, a model switch, incremental vs full scan, and a legacy estimated cursor.

Pushed in `b2b0af11`. CodeRabbit's docstring warning matches the existing CommonJS parser style in `src/lib/rollout.js`. I did not add JSDoc for this change.

obsesivegamer:
Follow-up in `27cad9d8`. The previous patch disabled resume whenever a conversation DB existed, including for sqlite-backed cursors. The file cursor now stores `usageSource`. Estimated or missing cursors still re-walk once. Later sqlite syncs resume.

obsesivegamer:
CodeRabbit caught a real bug: when we replay old turns to catch up with the database, we were dropping the previous reply from the next turn’s count. That is fixed, and the tests pass.

The other CodeRabbit notes we are leaving alone. One is “we don’t go back and rewrite old estimates when the database shows up later,” which was already the plan. The docstring warning is just their JSDoc checker. This repo doesn’t use that.

xiufengsun (approved):
Re-reviewed cd4514f9decc9841ad1a9b2d33f326a24c59299f. The previous blocker is resolved: the parser no longer invents cache hits and retains the explicit estimation boundary. The 22 focused Antigravity tests pass, including metadata-only updates, sparse metadata replay, model switches, and incremental/full-scan consistency. Integration PR #598 preserves these commits and passes the full cross-platform CI. This remains an IDE estimate improvement, not a fix for the separate CLI cache-counter report #595.

PR #590, opened 2026-09-07
fix(parser): preserve Claude reasoning usage without double billing

Kooooojun:
## Summary

Preserve reported reasoning usage from Claude Code / Anthropic-compatible session records instead of always storing zero.

- Read `output_tokens_details.thinking_tokens`, falling back to `reasoning_tokens` when absent (not when explicitly zero).
- Clamp to a non-negative integer within the inclusive output count.
- Split reasoning out of queue `output_tokens`: Claude rows price reasoning separately, so simply adding the detail would double bill it. Keep total tokens and cost unchanged.

## Validation

- Regression reproduced before patch: reported 163 became 0.
- 17 focused tests pass: new regression, Claude categorizer, WSL dual-install tests.
- New numeric-only fixture covers ten detail variants, token conservation, cost conservation and a second incremental parse with no queue changes.
- `validate:guardrails`, `validate:versions`, `git diff --check` pass.
- Full `npm test` is not green locally: Copilot SQLite-sidecar and sync-store tests failed (unmodified areas); full results will be appended when the run finishes. Initial run before installing dependencies also failed on missing packages; that is not claimed as validation.
- Self-reviewed diff for missing versus explicit zero, input/cache conservation, and billing invariants. Maintainer review is still required.

## Scope / limitations

No conversation content or credentials included. No release, version bump, deployment, or local service changes; release coordination belongs to the maintainer.

This fixes newly parsed queue usage, not historical backfill, session discovery, or context-category estimation. Existing logs with the detail omitted cannot reconstruct it. Upstream proxies must preserve the field first: https://github.com/router-for-me/CLIProxyAPI/issues/5587.

## Summary by CodeRabbit

- **Bug Fixes**
  - Claude usage reporting now separately accounts for reasoning tokens while preventing double billing.
  - Output token counts and cost calculations are more accurate when reasoning details are provided.
  - Incremental usage parsing remains consistent across repeated processing.

### DIFF
--- a/src/lib/rollout.js
+++ b/src/lib/rollout.js
@@ -4458,6 +4458,9 @@ function claudeMessageDedupKey(obj) {
 function normalizeClaudeUsage(u) {
   const inputTokens = toNonNegativeInt(u?.input_tokens);
   const outputTokens = toNonNegativeInt(u?.output_tokens);
+  const reasoningTokens = Math.min(outputTokens, toNonNegativeInt(
+    u?.output_tokens_details?.thinking_tokens ?? u?.output_tokens_details?.reasoning_tokens,
+  ));
   const cacheCreation = toNonNegativeInt(u?.cache_creation_input_tokens);
   const cacheRead = toNonNegativeInt(u?.cache_read_input_tokens);
   const totalTokens = inputTokens + outputTokens + cacheCreation + cacheRead;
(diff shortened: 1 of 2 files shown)


Kooooojun:
Full-suite baseline comparison completed with the same installed dependencies: unmodified base 3367ddd has 2,632 passed / 14 failed / 7 skipped (2,653 total); patched run has 2,633 passed / 14 failed / 7 skipped (2,654 total). The sets of failing test names match exactly; no newly failing tests observed. Final focused run (including 10 numeric fixture variants) passes all 17 tests. Self-review checked token/cost conservation, explicit-zero precedence, malformed values, and incremental idempotency. Maintainer review is pending; no service was changed.

PR #599, opened 2026-09-09
fix(antigravity): read cached input and reasoning tokens from sqlite usage metadata

### LINKED ISSUE #595: [bug]: Antigravity CLI 统计的token有BUG
### Preflight checklist

- [x] I searched [existing issues](https://github.com/xiufengsun/TokenTracker/issues) and didn't find a duplicate.
- [x] I'm on the [latest version](https://github.com/xiufengsun/TokenTracker/releases).

### What happened?

最近在用几乎免费的Antigravity cli, 发现里面用的token统计不准确，只有未命中输入和输出token，缓存输入token没有读取到，我发现其他开源项目能正常统计，比如tokscale和toktrack 这两个工具可以正常统计的，都是cmd终端的，还是习惯使用tokentracker请修复该bug谢谢。

### Steps to reproduce

tokentracker-cli和tokentracker桌面版都一样有该问题，版本都是最新版0.96.1

### Expected behavior

_No response_

### Is this a regression?

None

### TokenTracker version

0.96.1

### Node.js version

_No response_

### How did you install it?

npm / npx (CLI)

### OS / version

Windows 11

### Output of `tokentracker doctor` (if applicable)

```shell

```

### Anything else?

_No response_

xMinhx:
Fixes #595

Resolves the issue where Antigravity CLI token usage only tracked uncached inputs and text outputs while completely missing cached input tokens (`cached_input_tokens`) and reasoning output tokens.

### Background & Root Cause
In PR #588, `extractAntigravityGenInfo` attempted to read generation info from `gen_metadata`, but inspected field 9 (an internal context quota/budget struct) rather than field 4. That led to the conclusion in `b2b0af11` that `gen_metadata` only contained context window limits and lacked cache hit / output counters, causing prompt cache tracking to be dropped.

Furthermore, regarding directory layout: TokenTracker's `resolveAntigravityBrainDirs` already scans `antigravity`, `antigravity-ide`, and `antigravity-cli`, and `resolveAntigravityDbPath` automatically resolves `<base>/brain/<id>/.system_generated/logs/transcript.jsonl` to `<base>/conversations/<id>.db`. Both CLI and IDE store their generation info using the same SQLite schema and protobuf format.

In reality, the protobuf blob stored in `gen_metadata.data` contains Antigravity internal usage data in Field 1 -> Field 4:
- `f4.2`: `prompt_token_count` (uncached prompt tokens)
- `f4.5`: `cached_content_token_count` (cached prompt tokens / prompt cache hits)
- `f4.3`: internal output checksum used with the other fields
- `f4.9`: `text_token_count` (text output tokens)
- `f4.10`: `thoughts_token_count` (reasoning tokens)

### Before vs After Comparison (Live Session Data)
Evaluated against a real Antigravity CLI session (`~/.gemini/antigravity-cli/brain/0eef01fa-4715-4de0-a921-c131fb4f563b`, 37 conversation turns):

| Metric | Before (`main` / v0.96.2) | After (This PR) | Change |
| :--- | :--- | :--- | :--- |
| `cached_input_tokens` | `0` | `1,523,485` | **+1,523,485** (recovered prompt cache reads) |
| `input_tokens` (uncached) | `92,628` | `177,839` | **+85,211** (authoritative SQLite vs fallback estimate) |
| `output_tokens` (net text) | `5,500` | `19,001` | **+13,501** |
| `reasoning_output_tokens` | `5,109` | `16,521` | **+11,412** |
| `total_tokens` | `103,237` | `1,736,846` | **+1,633,609** (16.8× increase) |

**Token Invariant Verification:**
`177,839 (input) + 1,523,485 (cached) + 19,001 (text output) + 16,521 (reasoning) = 1,736,846 (total)` — exact match.

### Changes
1. **`src/lib/rollout.js`**:
   - Updated `extractAntigravityGenInfo` to extract Field 4 `UsageMetadata` (`uncachedInput`, `cachedInput`, `outputTokens`, `textOutput`, `reasoningOutput`).
   - Updated `readAntigravityConversationDb` to index turns that contain `UsageMetadata`.
   - Updated `parseAntigravityFile`:
     - When `UsageMetadata` is present, record exact `input_tokens`, `cached_input_tokens`, and `reasoning_output_tokens`.
     - Net output tokens (`output_tokens`) use `textOutput` or `candidates_token_count - thoughts_token_count` so reasoning tokens aren't double-billed by `computeRowCost`.
     - Maintains existing character heuristic as a fallback when SQLite rows are absent.
2. **`test/rollout-parser.test.js`**:
   - Extended `buildAntigravityTestProto` to encode Field 4 test fixtures.
   - Added unit test for `extractAntigravityGenInfo` usage metadata extraction.
   - Added integration test verifying multi-turn prompt caching and reasoning token aggregation.

### Testing
- `node --test test/rollout-parser.test.js` (all 272 tests pass).
- `node --test test/edge-pricing-parity.test.js` (pass).
- `npm run validate:guardrails` (clean).
- Verified idempotency and incremental cursor advances on live sessions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved token usage reporting for cached, uncached, output, text, and reasoning usage when metadata is available.
  - Billing and context-token calculations now use recorded usage data, with estimates retained when metadata is unavailable.
  - Usage totals remain accurate after session updates, database changes, or cursor migrations.
  - Correctly handles zero-context conversations and missing text output.

- **Documentation**
  - Updated usage notices to explain exact metadata and estimation behavior across supported languages.


src/lib/rollout.js
```diff
@@ -18817,7 +18817,52 @@ function extractAntigravityGenInfo(buf) {
...
+    const rTok = f4fields.find((f) => f.num === 10)?.val;
+    if (
+      Number.isFinite(pTok) ||
+      Number.isFinite(cTok) ||
+      Number.isFinite(oTok) ||
+      Number.isFinite(tTok) ||
+      Number.isFinite(rTok)
+    ) {
+      hasUsageMetadata = true;
+      uncachedInput = Number.isFinite(pTok) ? pTok : 0;
```
obsesivegamer: `f4.2` is the fresh uncached prompt, but uncached input is `f4.1 + f4.2`.

`f4.1` is present on 2,350 of 2,352 local turns and is a per-model constant (1016 or 1318 on this machine) — the fixed system/tool prefix. Leaving it out drops ~1.0–1.3k uncached tokens every turn.

```js
const prefix = Number.isFinite(f4fields.find((f) => f.num === 1)?.val)
  ? f4fields.find((f) => f.num === 1).val
  : 0;
uncachedInput = prefix + (Number.isFinite(pTok) ? pTok : 0);
```

Also include `num === 1` in the `hasUsageMetadata` check so a prefix-only row is not dropped.

src/lib/rollout.js
```diff
@@ -18817,7 +18817,52 @@ function extractAntigravityGenInfo(buf) {
...
+  let textOutput;
+  let reasoningOutput;
+  let hasUsageMetadata = false;
+
+  const f4 = inner.find((f) => f.num === 4)?.val;
+  if (f4) {
+    const f4fields = findAntigravityProtoFields(f4);
+    const pTok = f4fields.find((f) => f.num === 2)?.val;
+    const cTok = f4fields.find((f) => f.num === 5)?.val;
+    const oTok = f4fields.find((f) => f.num === 3)?.val;
```
obsesivegamer, line 19086: `f4.3` is not `candidates_token_count`. On local data it is an output checksum: `f4.3 === f4.9 + f4.10` on every complete row (and `f4.3 === f4.10` on the 205 reasoning-only rows that omit `f4.9`).

The fallback at the billing site (`textOutput` else `outputTokens - reasoning`) is the right shape. Worth renaming this so it is not confused with the public Vertex/Gemini `UsageMetadata` proto, which puts thoughts at field 14 and nested modality messages at 9/10. This blob is Antigravity's own usage message (same numbers tokscale reverse-engineered).

obsesivegamer (commented):
Nice catch on Field 4 — that is the usage blob. I independently checked 2,352 `gen_metadata` rows across the local Antigravity conversation DBs; the same two gaps show up.

**1. Include `f4.1` in uncached input**

`f4.1` is present on 2,350 of 2,352 turns. It is a per-model constant (1016 or 1318 here), i.e. the fixed system/tool prefix. Tokscale bills uncached input as `f4.1 + f4.2`; this PR bills `f4.2` only, which drops ~1.0–1.3k uncached tokens every turn.

**2. Treat `f4.3` as an output checksum, not candidates**

Coercing a missing `f4.9`/`f4.10` to 0, `f4.3 === f4.9 + f4.10` holds on 2,344 of 2,352 rows (the other 8 have no `f4.3`). So `f4.9` is text, `f4.10` is reasoning, and `f4.3` is their sum. The `textOutput || (outputTokens - reasoning)` fallback is fine; just don't name `f4.3` as Google `candidates_token_count`.

This layout is Antigravity's own usage message, not the public Vertex/Gemini `UsageMetadata` proto (thoughts is field 14 there, and 9/10 are nested modality messages). Worth correcting in the PR body so the field map does not drift later.

`f4.5` as cache-read is right, and Field 9 really is the context-quota struct, not billing.

xiufengsun (changes requested):
The field-4 extraction fixes fresh scans, but two incremental cases still miss the correction. Reproduced with the PR's SQLite/protobuf fixtures:

1. Parse a transcript with v0.96.2, then run this head using the same cursors and unchanged files. A turn containing 2,000 cached tokens remains at 0; eventsAggregated=0 and bucketsQueued=0.
2. Parse fresh on this head, then update only gen_metadata (cached input 2,000 -> 4,000). The next sync keeps 2,000, again with no queued update.

The fast path checks only transcript inode/size/mtime, and lastLine skips previously counted events. Please add a versioned, replay-safe migration and database-change reconciliation that replaces previous contributions rather than adding them again; verify global/project totals and two consecutive syncs for both upgrades and delayed SQLite metadata.

Also, you are right about discovery: antigravity-cli is already included in resolveAntigravityBrainDirs. My earlier path diagnosis was incorrect. #595 stays open until the complete fix is released.

src/lib/rollout.js
xMinhx: Nice catch! And thanks for pointing it out!

src/lib/rollout.js
xMinhx: Thanks for the review!

xMinhx:
Thanks for the thorough reviews and reproduction cases!

Pushed two commits to address both findings:

1. **System prompt prefix (`f4.1`)** (`61038ca5`):
   - Added `f4.1` into `uncachedInput` (`f4.1 + f4.2`) as pointed out by @obsesivegamer so we don't drop the fixed system/tool prefix tokens each turn.
   - Updated the proto notes in the code to reflect that `f4.3` is Antigravity's internal output checksum (`f4.9 + f4.10`) rather than standard Gemini `UsageMetadata`.

2. **Cursor migration and delayed SQLite reconciliation** (`b44bd635`):
   - Bumped `ANTIGRAVITY_CURSOR_VERSION = 2` and added tracking for the SQLite database's `mtime` and `size` alongside the transcript cursor.
   - Each cursor now stores its bucket-level `contributions`. When the DB changes or an older cursor is detected on unchanged transcripts, we retract the previous contribution before re-aggregating, avoiding double counts.
   - Added tests covering legacy v0.96.2 migration, delayed SQLite metadata updates (`2,000 -> 4,000`), project queue reconciliation, and confirmed consecutive syncs are idempotent no-ops.

xiufengsun (changes requested):
Re-reviewed e534324c3aa306f4cc312366643456c92dcdc56e against current main. The unchanged-transcript migration and SQLite-only refresh cases are now addressed, but migration still fails when the transcript also grows.

Reproduction: parse a session with the v0.96.2/main parser (SQLite usage has 1,000 uncached + 2,000 cached input); append one USER_INPUT line before the first upgraded sync; run this head twice. The queue stays unchanged with cached_input_tokens = 0 instead of 2,000. `needsReconcile` requires `sameTranscript`, so the incremental branch skips the historical planner and nevertheless stamps cursorVersion = 2. The missing history is then treated as migrated. The same condition skips corrections when SQLite and the transcript change together.

Please reconcile old contributions whenever the cursor needs migration or SQLite changes, including an appended transcript, and cover legacy + append, SQLite correction + append, project totals, and a second no-op sync. Issue #595 stays open until a release containing a verified fix.

xMinhx:
Thanks, this identified a real migration bug in PR #599. Partial discovery could reconstruct a legacy file's contribution differently from v0.96.2 SQLite context billing. The fix now defers all legacy migration until the full inventory allows a source-wide rebuild. Regression coverage includes a missing session, an appended live session, and SQLite-backed legacy totals.

xiufengsun:
The migration regression is fixed. I re-ran the scenario from my last comment against this branch, and to make sure the harness could actually detect the old failure I first pointed it at 165438d7, where it reproduced the loss (13 → 7 on my synthetic corpus, the same shape as the 36 → 18 I reported). On the current head the same scenario holds at 13 across two consecutive syncs, and the SQLite-backed and whole-directory-deleted variants hold too. Thanks for taking the source-wide rebuild approach instead of patching the subtraction.

Two problems remain in the same code path.

**1. Partial discovery still runs the character-estimate rebuild, and the drift is permanent.**

The skip guard at `rollout.js:19630` tests `preserveUnknownLegacyAggregate` only:

```js
if (preserveUnknownLegacyAggregate && legacyMigrationPaths.has(key)) { ... continue; }
```

while `needsFullRebuild` requires `inventoryComplete` (19520). So when one brain directory is unreadable but every legacy transcript is still present, neither guard fires: `preserveUnknownLegacyAggregate` is false because nothing is missing, and `needsFullRebuild` is false because the inventory is incomplete. Each legacy file then goes through the `legacyMode: true` branch, which subtracts a character-estimate reconstruction of a contribution that v0.96.2 billed from the SQLite context value.

Measured on a synthetic corpus: baseline 1509, after sync 5536, ground truth 4040, so +37%. It does not self-heal, because the cursor is already stamped v2 and there is no legacy file left to trigger a rebuild. Syncs 3 and 4 after the directory becomes readable again stay at 5536.

This matters for the reporter on #595 specifically: Windows with both a cli and an ide directory is exactly the shape where one path can be unreadable while the other is fine.

No test covers it. I instrumented the branch and ran the suite: `legacyMigration` taken 0 times across 293 passing tests, `needsFullRebuild` taken 8 times. Every legacy migration test leaves `inventoryComplete` at its default of true.

**2. A deleted transcript freezes the whole source, permanently.**

The deferral at 19630 has no exit. A deleted transcript never comes back, and nothing clears an Antigravity cursor (I grepped `sync.js`; only codex and claude-mem have cleanup). So `preserveUnknownLegacyAggregate` stays true forever, and the source stops counting new usage on pre-existing sessions:

```
baseline 13 → delete one historical transcript, append a new turn to the live session
165438d7: 18        (the new turn counts)
this PR:   13, 13   (the new turn is lost, permanently)
```

A newly created session after the upgrade does count, so the symptom is "old sessions frozen, new ones fine", which is hard to notice and harder to report. Anyone who reinstalled Antigravity, cleaned out an old session, or renamed the directory lands here.

A shape that avoids both: instead of skipping the file, stamp a "no retroactive claim" v2 cursor on it, with `lastLine = prev.lastLine`, `contributions = {}`, `cursorVersion = 2`, and `usageSource` carried over. That means "this file's history stays in the old aggregate untouched, and from here on it bills from exact values". It neither retracts history nor freezes the future, and it does not depend on a reconstruction being accurate.

Everything else checks out: cost stays on the five columns, reasoning is not double-counted into output, `f4.1` is in `uncachedInput` now, and all five locales are synced (`validate:copy`, `validate:ui-hardcode` and `validate:guardrails` all exit 0).

One thing worth adding while you are in there. The evidence for the token semantics is one user, one model (gemini-3.8-flash), one platform. A self-consistency check would strengthen it: whether `f4.1 + f4.2 + f4.5` matches field 9's `contextTokens` on the same turn. The code already assumes that relation holds where it computes `contextTokens = uncached + cached`.

obsesivegamer:
## One thing to fix before this lands: on my data, 74% of usage records get dropped

Thanks for this PR — reading real usage from the SQLite metadata is a big improvement. I tested it on my own Antigravity data and found one problem that hides most of that improvement. Details, a repro, and a suggested fix are below.

### The problem

The stricter decoder added in `0736520a` (`decodeAntigravityVarint`) throws whenever a number is bigger than JavaScript can hold exactly (2^53). `extractAntigravityGenInfo` catches that error and returns `null`, so **the whole record is thrown away**: its token counts *and* its model name. Those turns then fall back to the text-length estimate under `antigravity-unknown`.

Antigravity writes one of those big numbers into most records — on my data, 3 out of 4 — and it sits in a field the parser never even reads.

### Short summary

- The value is **2^64 − 1**, stored at `.1.9.2` as the ten bytes `ff ff ff ff ff ff ff ff ff 01`. That's valid protobuf: the standard ten-byte encoding of 2^64 − 1, which is also how a signed 64-bit `-1` is written. Without a schema there's no way to tell which one Antigravity means, but either way it's legal data.
- **`main` doesn't have this problem.** Its decoder never throws. The record loss is new in this PR.
- **The fix:** keep your stricter checks, but return oversized values exactly, as a `BigInt`, instead of throwing. Token reads then skip anything that isn't a safe integer.
- On my data that brings back **all 4,153 lost records**, and the Antigravity total goes from **165.6M to 441.3M** tokens.
- **It also clears up almost all of `antigravity-unknown`:** 1,911,646 → 7,932 tokens. The model name was inside the dropped records all along.
- **The fix is ready to pull:** three commits on top of `bd033eba` that fast-forward (details at the end).
- **Also below:** the `f4.1 + f4.2 + f4.5` vs `contextTokens` check from the last review, run on a second data set. It never matches exactly.

### The three decoders side by side

|  | `main` | This PR (`bd033eba`) | This PR + fix |
|---|---|---|---|
| A value above 2^53 | accepted, but rounded (2^64 − 1 → `18446744073709552000`) | **throws; whole record dropped** | kept exactly, as a `BigInt` |
| A varint wider than 64 bits | accepted | rejected | rejected |
| Reads the usage fields (`.1.4`) | no | yes | yes |
| Records dropped, on my 5,636 | 0 | **4,153 (74%)** | 0 |

Your hardening is right to reject bad data. It's just also rejecting good data.

---

### Example 1 — minimal repro

A single record with model, context tokens, and usage, plus 2^64 − 1 in the unread field `.1.9.2`:

```js
// node repro.js   (run from the repo root)
const { extractAntigravityGenInfo } = require(require("path").resolve("src/lib/rollout.js"));

const varint = (n) => { const o = []; do { let b = n & 0x7f; n >>>= 7; if (n) b |= 0x80; o.push(b); } while (n); return Buffer.from(o); };
const field = (num, bytes) => Buffer.concat([varint((num << 3) | 2), varint(bytes.length), bytes]);
const int = (num, value) => Buffer.concat([varint(num << 3), varint(value)]);

const UINT64_MAX = Buffer.from([0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x01]);

const record = field(1, Buffer.concat([
  field(19, Buffer.from("gemini-3.8-flash")),                        // .1.19     model
  field(9, Buffer.concat([varint((2 << 3) | 0), UINT64_MAX,          // .1.9.2    2^64 - 1 (never read)
                          field(10, int(1, 25000))])),              // .1.9.10.1 context tokens
  field(4, Buffer.concat([int(2, 4583), int(5, 16319)])),            // .1.4      usage
  field(20, Buffer.concat([field(1, Buffer.from("last_step_index")), field(2, Buffer.from("7"))])),
]));

console.log(JSON.stringify(extractAntigravityGenInfo(record)));
```

Output:

```
bd033eba (this PR):  null
with the fix:        {"model":"gemini-3.8-flash","contextTokens":25000,"lastStepIndex":7,"uncachedInput":4583,"cachedInput":16319,"outputTokens":0,"textOutput":0,"reasoningOutput":0,"hasUsageMetadata":true}
```

The only thing wrong with that record is a number in a field nobody reads.

### Example 2 — real data

A frozen copy of my Antigravity data: 117 conversation databases, 5,636 `gen_metadata` records.

|  | This PR (`bd033eba`) | This PR + fix |
|---|---|---|
| Records dropped | 4,153 | 0 |
| Antigravity total tokens | 165,586,843 | 441,337,831 |
| Cached tokens | 149.18M | 386.12M |
| `antigravity-unknown` | 1,911,646 | 7,932 |
| `gemini-3.8-flash-tiered` | 0 | 118.22M |

Two things to notice:

- **Every one** of the 4,153 oversized values is at `.1.9.2`, and that count matches the dropped records exactly. On this data, it's the only cause.
- `gemini-3.8-flash-tiered` was **completely invisible** before: all 118M of its tokens were in dropped records.

### Example 3 — why not just clamp to `MAX_SAFE_INTEGER`?

I tried that first, and it's worse. Put the same bytes in a field that *is* read, such as prompt tokens at `.1.4.2`, and clamping bills **9,007,199,254,740,991 prompt tokens** for a single turn. Keeping the exact `BigInt`, and refusing to bill anything that isn't a safe integer, avoids that.

---

### Suggestions

**1. Return oversized values instead of throwing.** Values up to 2^53 − 1 stay Numbers, exactly as now. Only when the running total stops being a safe integer, re-read the bytes into an exact `BigInt`:

```diff
 function decodeAntigravityVarint(buf, offset) {
   if (!Buffer.isBuffer(buf) || !Number.isSafeInteger(offset) || offset < 0) {
     throw new RangeError("invalid Antigravity protobuf offset");
   }
+  const start = offset;
   let res = 0;
   for (let count = 0; count < 10; count++) {
     if (offset >= buf.length) throw new RangeError("truncated Antigravity protobuf varint");
     const b = buf[offset++];
+    // 64 bits is nine 7-bit groups plus one bit of a tenth byte.
+    if (count === 9 && b > 1) throw new RangeError("Antigravity protobuf varint exceeds 64 bits");
     res += (b & 0x7f) * 2 ** (count * 7);
-    if (!Number.isSafeInteger(res)) {
-      throw new RangeError("unsafe Antigravity protobuf varint");
+    if (!(b & 0x80)) {
+      if (Number.isSafeInteger(res)) return [res, offset];
+      let exact = 0n;
+      for (let i = start; i < offset; i++) {
+        exact |= BigInt(buf[i] & 0x7f) << BigInt((i - start) * 7);
+      }
+      return [exact, offset];
     }
-    if (!(b & 0x80)) return [res, offset];
   }
   throw new RangeError("overlong Antigravity protobuf varint");
 }
```

**2. Keep tags strict.** A tag must still be an ordinary Number, so a corrupt record fails on purpose, not through a BigInt/Number `TypeError` at `tag / 8`. Lengths need no change: a `BigInt` length is always past 2^53, so your existing `len > buf.length - offset` check already rejects it.

```diff
     const [tag, next] = decodeAntigravityVarint(buf, offset);
+    if (typeof tag !== "number") throw new RangeError("unsafe Antigravity protobuf field tag");
```

**3. Read token counts with `Number.isSafeInteger` instead of `Number.isFinite`** in `extractAntigravityGenInfo` — 13 places. Both reject a `BigInt`, but `Number.isFinite(2 ** 53)` is `true` and `Number.isSafeInteger(2 ** 53)` is `false`. With the fix above nothing changes today; this just keeps an out-of-range count from ever being billed if the decoder changes later.

**4. Add a test with the real ten bytes.** None of the current tests use a value above 2^53, which is why they all pass while this bug is present. The three cases worth covering:

- 2^64 − 1 in an unread field → the record keeps its model and usage
- 2^64 − 1 in `.1.4.2` next to `.1.4.1 = 100` → `uncachedInput` is `100`, not a clamped value
- a tenth byte of `0x02` → rejected as wider than 64 bits

**5. Optional: re-read sessions synced before the fix.** Anyone who ran this branch before the fix has saved cursors holding the undercounted totals. Since neither the transcript nor the database changes, nothing triggers a re-read. I added a small `extractorRevision` stamp on the cursor; a stale one sends the file through your existing reconcile path, the same one used when the database changes.

For people upgrading straight from `main` this **shouldn't be needed**: `main` has no cursor version, so your legacy migration should already re-read everything. (I've tested upgrading from this branch, not from `main`.) Happy to leave it out if you'd rather keep the PR small.

If you keep it, one catch: it also has to go in the incomplete-ledger preflight, the `!sameDb` check that sets `needsFullRebuild`. Otherwise a capped ledger (over `ANTIGRAVITY_MAX_CONTRIBUTIONS`) gets reconciled file by file, and the buckets it no longer tracks stay in the totals underneath the fresh numbers.

And a heads-up if you build the "no retroactive claim" cursor from the last review: if that stamp is written by hand instead of through the normal cursor write, give it `extractorRevision: ANTIGRAVITY_EXTRACTOR_REVISION` too. A v2 cursor without the field counts as stale, and a stale cursor is sent back through reconcile on the next sync, which reparses the transcript. That is exactly the retroactive claim the stamp exists to prevent.

---

### The context-token check from the last review

@xiufengsun asked whether `f4.1 + f4.2 + f4.5` matches `.1.9.10.1` (`contextTokens`) on the same turn, since the evidence so far was one user, one model, one platform. Here's a second data set: a different user, on macOS, with 8 model names, over the same 5,636 records.

**It never matches exactly.** It's close, and the gap depends on the model family:

|  | Records | Exactly equal | Median gap (context − usage sum) | Middle half of the gap | Within 10% | Within 25% |
|---|---|---|---|---|---|---|
| Gemini (6 model names) | 5,044 | 0 | +6,787 tokens (+9.1%) | +4,261 to +8,081 | 52.7% | 96.7% |
| Claude (2 model names) | 239 | 0 | −596 tokens (−0.8%) | −6,167 to +1,174 | 95.0% | 99.2% |

I left out the other 353 records: they have no model name and only `f4.1` set.

- **It isn't an off-by-one.** Comparing the next turn's context with this turn's prompt plus output is no closer (within 10%: Gemini 51.0%, Claude 92.2%).
- **On Gemini the context value is usually 4–8K tokens higher**, which looks like a fixed chunk rather than noise. My guess is something counted in one number but not the other, such as tool declarations. The data can't confirm that.

**What it means for the code:** these are two different counters, so one shouldn't be subtracted from the other, and right now that can happen. After a turn with usage, `previousContextTokens` is `uncached + cached`. If the next turn has no `.1.4`, the fallback bills `Math.max(0, contextTokens - previousContextTokens)` with `contextTokens` taken from `.1.9.10.1`, so on Gemini the ~6.8K offset would be billed as input. On my data it can't happen, because every record carries `.1.4`, but it could on data that mixes the two.

---

### Separate follow-ups (not blocking this PR)

The bug above doesn't cause these two, but fixing it makes both easier to see.

**A. The same Claude model shows up under two names.**

`normalizeAntigravityTranscriptModel` keeps dots, and the two places a name can come from write Claude versions differently:

| Where the name comes from | Raw name | Slug today |
|---|---|---|
| Database record (field 19) | `claude-opus-4-6-thinking` | `claude-opus-4-6` |
| Transcript "Model Selection" line | `Claude Opus 4.6` | `claude-opus-4.6` |

So one model becomes two dashboard rows. On my data after the fix: `claude-opus-4-6` has 7,409,538 tokens and `claude-opus-4.6` has 3,078. Both are priced the same ($5 per million input tokens), so it's cosmetic, not a billing error.

Gemini isn't affected, because its database names use dots too (`gemini-3.8-flash`), so both sources agree.

The hyphen form is Anthropic's official model id: it's what Claude Code logs (`claude-haiku-4-5-20251001`) and what Antigravity's own database records. One line merges them:

```js
// in normalizeAntigravityTranscriptModel, right after the loop that trims to the model name
if (slug.startsWith("claude-")) slug = slug.replace(/(\d)\.(\d)/g, "$1-$2");
```

I ran it against every raw name in my data. Only `Claude Opus 4.6` and `Claude Sonnet 4.6` change (to `claude-opus-4-6` and `claude-sonnet-4-6`); every Gemini name and every database Claude name comes out exactly as before. One thing to know: rows already saved under the dot name won't move on their own; they'd need the session re-read.

**B. `gemini-pro-default` and `gemini-default` have no price.**

These are real names Antigravity writes into the database: `gemini-pro-default` in 250 records, `gemini-default` in 1. TokenTracker has no pricing for either, so both show $0.

Before the fix, nearly all of those records were being dropped, so it barely showed. After the fix, on my data, it's **10,379,800 tokens at $0** (`gemini-pro-default` 10,313,441 + `gemini-default` 66,359), up from 188,698.

The data doesn't say which real model "default" means. Two options:

- price them as a specific Gemini model, if you know which one Antigravity's default is; or
- leave them unpriced, but label them "unpriced" in the dashboard so $0 isn't read as free.

---

### How I checked it

- **Record by record:** old and fixed extractors on all 5,636 records. 4,153 recovered, **0 broken**, and **0 changed** among the records that already worked.
- **Your field mapping holds on the recovered records:** `f4.3 == f4.9 + f4.10` on all 4,984 that carry all three. The grok-4.6 review also found `.1.4.2` resets between turns (per-request, not a running total) in 114 of 114 multi-record databases.
- **Upgrade:** old code, then the fixed code over the same saved cursors, lands on exactly the same totals as a fresh sync (70 of 70 buckets), and a second sync changes nothing.
- `npm run ci:local`: 2,875 tests, 0 failed.
- Reviewed independently with gpt-5.6-sol and grok-4.6 (before the cleanup commit, which changes no behavior).

**Not verified:** whether these totals match Google's own usage numbers. I don't have a way to check that from local data.

### Commits you can pull

They're on [`obsesivegamer/TokenTracker` → `pr599-uint64-varint-fix`](https://github.com/obsesivegamer/TokenTracker/tree/pr599-uint64-varint-fix): three commits directly on top of this PR's head `bd033eba`, so they fast-forward ([full diff](https://github.com/obsesivegamer/TokenTracker/compare/bd033eba...pr599-uint64-varint-fix)).

```sh
git pull --ff-only https://github.com/obsesivegamer/TokenTracker.git pr599-uint64-varint-fix
```

| Commit | What it does |
|---|---|
| [`6181461f`](https://github.com/obsesivegamer/TokenTracker/commit/6181461fda71d6653fce989e7468628d8af6cf03) | Decoder fix and tag check (suggestions 1–2), `extractorRevision` (5), and four of the tests (4) |
| [`61f704bd`](https://github.com/obsesivegamer/TokenTracker/commit/61f704bd0877d7ec2f14fcd6f43abda48d556c53) | `Number.isSafeInteger` for token reads (3), stale extractor in the incomplete-ledger preflight, capped-ledger test |
| [`03fb1a3f`](https://github.com/obsesivegamer/TokenTracker/commit/03fb1a3f57d144e344bdf94927318037c4328f89) | Cleanup: drops a length check that could never fire, trims comments |

If you'd rather leave out suggestion 5, say so and I'll split it into its own commit.

xMinhx:
> ## One thing to fix before this lands: on my data, 74% of usage records get dropped
> Thanks for this PR — reading real usage from the SQLite metadata is a big improvement. I tested it on my own Antigravity data and found one problem that hides most of that improvement. Details, a repro, and a suggested fix are below.
> 
> ### The problem
> The stricter decoder added in `0736520a` (`decodeAntigravityVarint`) throws whenever a number is bigger than JavaScript can hold exactly (2^53). `extractAntigravityGenInfo` catches that error and returns `null`, so **the whole record is thrown away**: its token counts _and_ its model name. Those turns then fall back to the text-length estimate under `antigravity-unknown`.
> 
> Antigravity writes one of those big numbers into most records — on my data, 3 out of 4 — and it sits in a field the parser never even reads.
> 
> ### Short summary
> * The value is **2^64 − 1**, stored at `.1.9.2` as the ten bytes `ff ff ff ff ff ff ff ff ff 01`. That's valid protobuf: the standard ten-byte encoding of 2^64 − 1, which is also how a signed 64-bit `-1` is written. Without a schema there's no way to tell which one Antigravity means, but either way it's legal data.
> * **`main` doesn't have this problem.** Its decoder never throws. The record loss is new in this PR.
> * **The fix:** keep your stricter checks, but return oversized values exactly, as a `BigInt`, instead of throwing. Token reads then skip anything that isn't a safe integer.
> * On my data that brings back **all 4,153 lost records**, and the Antigravity total goes from **165.6M to 441.3M** tokens.
> * **It also clears up almost all of `antigravity-unknown`:** 1,911,646 → 7,932 tokens. The model name was inside the dropped records all along.
> * **The fix is ready to pull:** three commits on top of `bd033eba` that fast-forward (details at the end).
> * **Also below:** the `f4.1 + f4.2 + f4.5` vs `contextTokens` check from the last review, run on a second data set. It never matches exactly.
> 
> ### The three decoders side by side
> `main`	This PR (`bd033eba`)	This PR + fix
> A value above 2^53	accepted, but rounded (2^64 − 1 → `18446744073709552000`)	**throws; whole record dropped**	kept exactly, as a `BigInt`
> A varint wider than 64 bits	accepted	rejected	rejected
> Reads the usage fields (`.1.4`)	no	yes	yes
> Records dropped, on my 5,636	0	**4,153 (74%)**	0
> Your hardening is right to reject bad data. It's just also rejecting good data.
> 
> ### Example 1 — minimal repro
> A single record with model, context tokens, and usage, plus 2^64 − 1 in the unread field `.1.9.2`:
> 
> ```js
> // node repro.js   (run from the repo root)
> const { extractAntigravityGenInfo } = require(require("path").resolve("src/lib/rollout.js"));
> 
> const varint = (n) => { const o = []; do { let b = n & 0x7f; n >>>= 7; if (n) b |= 0x80; o.push(b); } while (n); return Buffer.from(o); };
> const field = (num, bytes) => Buffer.concat([varint((num << 3) | 2), varint(bytes.length), bytes]);
> const int = (num, value) => Buffer.concat([varint(num << 3), varint(value)]);
> 
> const UINT64_MAX = Buffer.from([0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x01]);
> 
> const record = field(1, Buffer.concat([
>   field(19, Buffer.from("gemini-3.8-flash")),                        // .1.19     model
>   field(9, Buffer.concat([varint((2 << 3) | 0), UINT64_MAX,          // .1.9.2    2^64 - 1 (never read)
>                           field(10, int(1, 25000))])),              // .1.9.10.1 context tokens
>   field(4, Buffer.concat([int(2, 4583), int(5, 16319)])),            // .1.4      usage
>   field(20, Buffer.concat([field(1, Buffer.from("last_step_index")), field(2, Buffer.from("7"))])),
> ]));
> 
> console.log(JSON.stringify(extractAntigravityGenInfo(record)));
> ```
> 
> Output:
> 
> ```
> bd033eba (this PR):  null
> with the fix:        {"model":"gemini-3.8-flash","contextTokens":25000,"lastStepIndex":7,"uncachedInput":4583,"cachedInput":16319,"outputTokens":0,"textOutput":0,"reasoningOutput":0,"hasUsageMetadata":true}
> ```
> 
> The only thing wrong with that record is a number in a field nobody reads.
> 
> ### Example 2 — real data
> A frozen copy of my Antigravity data: 117 conversation databases, 5,636 `gen_metadata` records.
> 
> This PR (`bd033eba`)	This PR + fix
> Records dropped	4,153	0
> Antigravity total tokens	165,586,843	441,337,831
> Cached tokens	149.18M	386.12M
> `antigravity-unknown`	1,911,646	7,932
> `gemini-3.8-flash-tiered`	0	118.22M
> Two things to notice:
> 
> * **Every one** of the 4,153 oversized values is at `.1.9.2`, and that count matches the dropped records exactly. On this data, it's the only cause.
> * `gemini-3.8-flash-tiered` was **completely invisible** before: all 118M of its tokens were in dropped records.
> 
> ### Example 3 — why not just clamp to `MAX_SAFE_INTEGER`?
> I tried that first, and it's worse. Put the same bytes in a field that _is_ read, such as prompt tokens at `.1.4.2`, and clamping bills **9,007,199,254,740,991 prompt tokens** for a single turn. Keeping the exact `BigInt`, and refusing to bill anything that isn't a safe integer, avoids that.
> 
> ### Suggestions
> **1. Return oversized values instead of throwing.** Values up to 2^53 − 1 stay Numbers, exactly as now. Only when the running total stops being a safe integer, re-read the bytes into an exact `BigInt`:
> 
> ```diff
>  function decodeAntigravityVarint(buf, offset) {
>    if (!Buffer.isBuffer(buf) || !Number.isSafeInteger(offset) || offset < 0) {
>      throw new RangeError("invalid Antigravity protobuf offset");
>    }
> +  const start = offset;
>    let res = 0;
>    for (let count = 0; count < 10; count++) {
>      if (offset >= buf.length) throw new RangeError("truncated Antigravity protobuf varint");
>      const b = buf[offset++];
> +    // 64 bits is nine 7-bit groups plus one bit of a tenth byte.
> +    if (count === 9 && b > 1) throw new RangeError("Antigravity protobuf varint exceeds 64 bits");
>      res += (b & 0x7f) * 2 ** (count * 7);
> -    if (!Number.isSafeInteger(res)) {
> -      throw new RangeError("unsafe Antigravity protobuf varint");
> +    if (!(b & 0x80)) {
> +      if (Number.isSafeInteger(res)) return [res, offset];
> +      let exact = 0n;
> +      for (let i = start; i < offset; i++) {
> +        exact |= BigInt(buf[i] & 0x7f) << BigInt((i - start) * 7);
> +      }
> +      return [exact, offset];
>      }
> -    if (!(b & 0x80)) return [res, offset];
>    }
>    throw new RangeError("overlong Antigravity protobuf varint");
>  }
> ```
> 
> **2. Keep tags strict.** A tag must still be an ordinary Number, so a corrupt record fails on purpose, not through a BigInt/Number `TypeError` at `tag / 8`. Lengths need no change: a `BigInt` length is always past 2^53, so your existing `len > buf.length - offset` check already rejects it.
> 
> ```diff
>      const [tag, next] = decodeAntigravityVarint(buf, offset);
> +    if (typeof tag !== "number") throw new RangeError("unsafe Antigravity protobuf field tag");
> ```
> 
> **3. Read token counts with `Number.isSafeInteger` instead of `Number.isFinite`** in `extractAntigravityGenInfo` — 13 places. Both reject a `BigInt`, but `Number.isFinite(2 ** 53)` is `true` and `Number.isSafeInteger(2 ** 53)` is `false`. With the fix above nothing changes today; this just keeps an out-of-range count from ever being billed if the decoder changes later.
> 
> **4. Add a test with the real ten bytes.** None of the current tests use a value above 2^53, which is why they all pass while this bug is present. The three cases worth covering:
> 
> * 2^64 − 1 in an unread field → the record keeps its model and usage
> * 2^64 − 1 in `.1.4.2` next to `.1.4.1 = 100` → `uncachedInput` is `100`, not a clamped value
> * a tenth byte of `0x02` → rejected as wider than 64 bits
> 
> **5. Optional: re-read sessions synced before the fix.** Anyone who ran this branch before the fix has saved cursors holding the undercounted totals. Since neither the transcript nor the database changes, nothing triggers a re-read. I added a small `extractorRevision` stamp on the cursor; a stale one sends the file through your existing reconcile path, the same one used when the database changes.
> 
> For people upgrading straight from `main` this **shouldn't be needed**: `main` has no cursor version, so your legacy migration should already re-read everything. (I've tested upgrading from this branch, not from `main`.) Happy to leave it out if you'd rather keep the PR small.
> 
> If you keep it, one catch: it also has to go in the incomplete-ledger preflight, the `!sameDb` check that sets `needsFullRebuild`. Otherwise a capped ledger (over `ANTIGRAVITY_MAX_CONTRIBUTIONS`) gets reconciled file by file, and the buckets it no longer tracks stay in the totals underneath the fresh numbers.
> 
> And a heads-up if you build the "no retroactive claim" cursor from the last review: if that stamp is written by hand instead of through the normal cursor write, give it `extractorRevision: ANTIGRAVITY_EXTRACTOR_REVISION` too. A v2 cursor without the field counts as stale, and a stale cursor is sent back through reconcile on the next sync, which reparses the transcript. That is exactly the retroactive claim the stamp exists to prevent.
> 
> ### The context-token check from the last review
> @xiufengsun asked whether `f4.1 + f4.2 + f4.5` matches `.1.9.10.1` (`contextTokens`) on the same turn, since the evidence so far was one user, one model, one platform. Here's a second data set: a different user, on macOS, with 8 model names, over the same 5,636 records.
> 
> **It never matches exactly.** It's close, and the gap depends on the model family:
> 
> Records	Exactly equal	Median gap (context − usage sum)	Middle half of the gap	Within 10%	Within 25%
> Gemini (6 model names)	5,044	0	+6,787 tokens (+9.1%)	+4,261 to +8,081	52.7%	96.7%
> Claude (2 model names)	239	0	−596 tokens (−0.8%)	−6,167 to +1,174	95.0%	99.2%
> I left out the other 353 records: they have no model name and only `f4.1` set.
> 
> * **It isn't an off-by-one.** Comparing the next turn's context with this turn's prompt plus output is no closer (within 10%: Gemini 51.0%, Claude 92.2%).
> * **On Gemini the context value is usually 4–8K tokens higher**, which looks like a fixed chunk rather than noise. My guess is something counted in one number but not the other, such as tool declarations. The data can't confirm that.
> 
> **What it means for the code:** these are two different counters, so one shouldn't be subtracted from the other, and right now that can happen. After a turn with usage, `previousContextTokens` is `uncached + cached`. If the next turn has no `.1.4`, the fallback bills `Math.max(0, contextTokens - previousContextTokens)` with `contextTokens` taken from `.1.9.10.1`, so on Gemini the ~6.8K offset would be billed as input. On my data it can't happen, because every record carries `.1.4`, but it could on data that mixes the two.
> 
> ### Separate follow-ups (not blocking this PR)
> The bug above doesn't cause these two, but fixing it makes both easier to see.
> 
> **A. The same Claude model shows up under two names.**
> 
> `normalizeAntigravityTranscriptModel` keeps dots, and the two places a name can come from write Claude versions differently:
> 
> Where the name comes from	Raw name	Slug today
> Database record (field 19)	`claude-opus-4-6-thinking`	`claude-opus-4-6`
> Transcript "Model Selection" line	`Claude Opus 4.6`	`claude-opus-4.6`
> So one model becomes two dashboard rows. On my data after the fix: `claude-opus-4-6` has 7,409,538 tokens and `claude-opus-4.6` has 3,078. Both are priced the same ($5 per million input tokens), so it's cosmetic, not a billing error.
> 
> Gemini isn't affected, because its database names use dots too (`gemini-3.8-flash`), so both sources agree.
> 
> The hyphen form is Anthropic's official model id: it's what Claude Code logs (`claude-haiku-4-5-20251001`) and what Antigravity's own database records. One line merges them:
> 
> ```js
> // in normalizeAntigravityTranscriptModel, right after the loop that trims to the model name
> if (slug.startsWith("claude-")) slug = slug.replace(/(\d)\.(\d)/g, "$1-$2");
> ```
> 
> I ran it against every raw name in my data. Only `Claude Opus 4.6` and `Claude Sonnet 4.6` change (to `claude-opus-4-6` and `claude-sonnet-4-6`); every Gemini name and every database Claude name comes out exactly as before. One thing to know: rows already saved under the dot name won't move on their own; they'd need the session re-read.
> 
> **B. `gemini-pro-default` and `gemini-default` have no price.**
> 
> These are real names Antigravity writes into the database: `gemini-pro-default` in 250 records, `gemini-default` in 1. TokenTracker has no pricing for either, so both show $0.
> 
> Before the fix, nearly all of those records were being dropped, so it barely showed. After the fix, on my data, it's **10,379,800 tokens at $0** (`gemini-pro-default` 10,313,441 + `gemini-default` 66,359), up from 188,698.
> 
> The data doesn't say which real model "default" means. Two options:
> 
> * price them as a specific Gemini model, if you know which one Antigravity's default is; or
> * leave them unpriced, but label them "unpriced" in the dashboard so $0 isn't read as free.
> 
> ### How I checked it
> * **Record by record:** old and fixed extractors on all 5,636 records. 4,153 recovered, **0 broken**, and **0 changed** among the records that already worked.
> * **Your field mapping holds on the recovered records:** `f4.3 == f4.9 + f4.10` on all 4,984 that carry all three. The grok-4.6 review also found `.1.4.2` resets between turns (per-request, not a running total) in 114 of 114 multi-record databases.
> * **Upgrade:** old code, then the fixed code over the same saved cursors, lands on exactly the same totals as a fresh sync (70 of 70 buckets), and a second sync changes nothing.
> * `npm run ci:local`: 2,875 tests, 0 failed.
> * Reviewed independently with gpt-5.6-sol and grok-4.6 (before the cleanup commit, which changes no behavior).
> 
> **Not verified:** whether these totals match Google's own usage numbers. I don't have a way to check that from local data.
> 
> ### Commits you can pull
> They're on [`obsesivegamer/TokenTracker` → `pr599-uint64-varint-fix`](https://github.com/obsesivegamer/TokenTracker/tree/pr599-uint64-varint-fix?rgh-link-date=2026-09-16T19%3A47%3A36Z): three commits directly on top of this PR's head `bd033eba`, so they fast-forward ([full diff](https://github.com/obsesivegamer/TokenTracker/compare/bd033eba...pr599-uint64-varint-fix)).
> 
> ```shell
> git pull --ff-only https://github.com/obsesivegamer/TokenTracker.git pr599-uint64-varint-fix
> ```
> 
> Commit	What it does
> [`6181461f`](https://github.com/obsesivegamer/TokenTracker/commit/6181461fda71d6653fce989e7468628d8af6cf03)	Decoder fix and tag check (suggestions 1–2), `extractorRevision` (5), and four of the tests (4)
> [`61f704bd`](https://github.com/obsesivegamer/TokenTracker/commit/61f704bd0877d7ec2f14fcd6f43abda48d556c53)	`Number.isSafeInteger` for token reads (3), stale extractor in the incomplete-ledger preflight, capped-ledger test
> [`03fb1a3f`](https://github.com/obsesivegamer/TokenTracker/commit/03fb1a3f57d144e344bdf94927318037c4328f89)	Cleanup: drops a length check that could never fire, trims comments
> If you'd rather leave out suggestion 5, say so and I'll split it into its own commit.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Great find and amazing work! Thank you!!

xMinhx:
Thanks @obsesivegamer and @xiufengsun!

Just updated the PR:

- Pulled in @obsesivegamer's uint64 varint fix for 2^64 - 1.
- Switched to the v2 cutover xiufengsun suggested. Old history stays intact, and we track baselineLine so future SQLite updates don't double-count past turns.
- Fixed the context scale mismatch so fallback turns stay on the dbContextTokens scale.
- Updated the test fixture to use an open file handle, which clears the CodeQL check.

All CI checks and local test suites are passing across macOS, Linux, and Windows.

xiufengsun:
Merged. @obsesivegamer — the uint64 find was the kind of review this repo needs; a hardening change that silently drops 74% of records is exactly the failure that hides behind a green build.

I did not take the numbers on trust. Ran the parser on my own Antigravity data (265 transcripts, 227 conversation DBs, 14,750 `gen_metadata` records) under `main` and under this branch, from a clean cursor:

| | main | this PR |
|---|---|---|
| total tokens | 49,246,260 | 1,387,042,180 |
| of which cache reads | 0 | 1,242,877,963 |
| input | 44,816,832 | 134,870,045 |
| computed cost | $35.40 | $100.39 |

28x the tokens but only 2.8x the cost. That ratio is the reassuring part — it means the growth is cache reads, not a prompt field counted twice.

**The double-count check, since this is where provider parsers here usually go wrong.** Codex's `input` includes cached reads, and copying that shape naively inflated cost 6-7x once already, so `f4.1 + f4.2` and `f4.5` had to be proven disjoint rather than assumed. On 14,453 records carrying usage metadata:

```
context == f4.1+f4.2            :      0
context == f4.1+f4.2+f4.5       :      1
context >  f4.1+f4.2+f4.5       : 11,961
other (context 0 or smaller)    :  2,491
```

They are disjoint. The per-turn samples say the same thing more directly — as soon as caching kicks in, uncached drops and cached rises by about the same amount:

```
u=17,556  c=0       ctx=21,223
u= 5,818  c=12,174  ctx=21,619
u= 6,235  c=12,169  ctx=22,044
```

Which also justifies folding `f4.1` (the fixed system prefix) into uncached input: on an uncached turn the provider bills it, and on a cached turn it moves into `f4.5` on its own.

**The migration, which is the part I cared about most.** Ran the real upgrade path — `main` parses first and leaves its cursor at EOF, then this branch parses on top of it:

```
phase 1 (main)          310 rows   49,246,260
phase 2 (upgrade)       310 rows   49,246,260   +0 rows
phase 3 (run it again)  310 rows   49,246,260   +0 rows
```

Zero rows appended on the upgrade, zero on the rerun. The `baselineLine` cutover does what it says: no retroactive claim, no double-count, idempotent. That is the right trade — a retroactive rebuild here would have been the shape that cost this project real data before.

**One consequence worth stating plainly, for whoever writes the release note.** Because history is deliberately left alone, existing users get a visible step in their Antigravity charts: everything before the upgrade stays on the old text-length estimate, everything after jumps roughly 28x. Nothing is wrong, but it will look wrong to someone who does not know why. New installs see the accurate numbers throughout.

Not folded in here: the usual `sync` double-run stress test against a seeded sandbox before this ships in a release.

PR #601, opened 2026-09-09
refactor(settings): split token number units into a dedicated row

LceAn:
## Summary

Follow-up to #596. As reported after using 0.96.2: the Wan/Yi option sat in the same segmented control as Compact/Full, but **display density** (compact vs full digits) and **unit system** (K/M/B vs 万/亿) are different dimensions. Split them into two settings rows:

| Row | Options | Notes |
|---|---|---|
| Token numbers | 紧凑 / 完整 | same as before #596 |
| Number units | K/M/B / 万/亿 | new row, disabled while full numbers are active |

Compact + Chinese units renders exactly like the previous combined option (`12,345 → 1.2万`, tooltips keep `1.2万 · 12,345`).

## Compatibility

- Unit system persists under a new `tt.tokenUnitSystem` localStorage key; cross-window `storage` sync covers both keys.
- 0.96.2 stored the Wan/Yi display as a third mode value (`tt.tokenFormat=chinese`). The provider migrates that transparently on mount to `compact` + Chinese units, and `normalizeTokenFormatMode` / `formatTokenCount` still map the legacy value defensively, so no user-visible setting is lost after upgrading.
- All existing consumers only use `formatTokens` / `formatTokensTooltip` / `setMode`, so the provider value extension is backward compatible.

## Changes

- `dashboard/src/lib/token-format.js`: `TOKEN_FORMAT_MODES` back to compact/full, new `TOKEN_UNIT_SYSTEMS` (`english`/`chinese`), `tt.tokenUnitSystem` persistence + legacy migration, `formatTokenCount` routes via unit system (and defensively via the legacy `chinese` mode value)
- `dashboard/src/ui/foundation/TokenFormatProvider.jsx`: unit system state, dual-key storage sync, migration on mount, `setUnitSystem` exposed
- `dashboard/src/components/settings/AppearanceSection.jsx`: dedicated Number-units row, disabled in full mode
- `copy.csv` + `zh` / `zh-TW` / `ja` / `ko` / `de`: token-unit labels and hints, hints no longer conflate units with density
- tests: unit-system routing, full-mode ignores units, legacy migration (unit + provider level), provider persistence/switch coverage

## Verification

- targeted suites 54/54 (`token-format`, `format`, `TokenFormatProvider`, settings sections)
- `tsc --noEmit`, `eslint` (changed files), `vite build`: pass
- `validate:copy`, `validate:locale` (100% zh/zh-TW), `validate:ui-hardcode`, `validate:guardrails`: pass
- full vitest: 722 passed; 3 failures are pre-existing on clean `main` (LeaderboardPage 5s-timeout flake under parallel load ×2, `dashboard-preload` route-boundary test fails identically single-run on `main`)

## Summary by CodeRabbit

- **New Features**
  - Added a separate setting for compact token units, allowing users to choose English (K/M/B) or Chinese (万/億) notation.
  - Token display settings now independently control compact versus full formatting and the units used in compact mode.
  - Unit preferences are saved and synchronized across sessions and browser tabs.

- **Bug Fixes**
  - Existing Chinese unit preferences are automatically preserved when upgrading to the new settings structure.

- **Documentation**
  - Updated German, Japanese, Korean, Traditional Chinese, and Simplified Chinese translations for the revised appearance settings.


xiufengsun:
Integrated through #607, preserving your commit. I added regression fixes for interrupted legacy migration, legacy Chinese storage events from another window, and standalone format overrides. Unit tests and browser checks confirm the unit choice survives refresh and full/compact switching. No further author action is needed; this will ship in the next release.

PR #605, opened 2026-09-09
feat(cli): add update command and print help for unknown commands

paveezheng:
## Summary

Headless CLI installs had no way to update TokenTracker in-place, and a mistyped subcommand dumped a Node stack trace. This PR adds `tokentracker update`, which detects the install channel (npm, bun, pnpm, yarn, Homebrew) and delegates to that package manager — matching Codex's approach — and prints help instead of a stack for unknown commands/options.

## Scope

- [x] CLI (`src/`)
- [x] Docs / CI / config

## Checklist

- [x] `npm test` passes

CLI help remains in `src/cli.js`, same as the existing command list. Dashboard copy registry is unchanged.

## Behavior

`tokentracker update` (alias: `upgrade`) inspects the running entrypoint and:

- **npm / bun / pnpm / yarn / Homebrew** — runs that channel's upgrade command
- **npx, desktop EmbeddedServer bundles, git/source checkouts** — refuses and tells the user how to update without switching install methods

Unknown commands print `Unknown command: …` plus help and exit 1, with no stack. Usage errors (`Unknown option:`) also omit the stack unless `--debug` is set.

## Tests

- Channel detection for brew keg vs Homebrew's npm global prefix, npm/bun/pnpm/yarn/npx, desktop bundles, and git checkouts
- End-to-end: a real `node bin/tracker.js update` from an npm-prefix layout invokes `npm install -g tokentracker-cli@latest` (fake `npm` on `PATH`, no registry)
- Unknown commands and `update --dry-run` print help/errors without a stack

## Summary by CodeRabbit

- **New Features**
  - Added `tokentracker update` and `upgrade` commands to update the CLI through its detected installation method.
  - Added support for `help` as an alias for CLI help.
  - Update guidance now covers npm, bun, pnpm, yarn, and Homebrew installations.

- **Bug Fixes**
  - Unknown commands and options now display concise error messages without stack traces.
  - Unsupported or locally installed dependencies no longer trigger inappropriate global updates and instead provide guidance.

- **Documentation**
  - Added update-command examples to English, German, Japanese, Korean, and Chinese quick-start guides.

### DIFF
--- a/README.de.md
+++ b/README.de.md
@@ -70,6 +70,7 @@ tokentracker status       # Hook-Status prüfen
 tokentracker status --json     # Maschinenlesbare Ausgabe (für jq, AI-Agenten)
 tokentracker status --light    # Reine ASCII-Tabelle (CI / SSH, kein Spinner)
 tokentracker doctor       # Health Check
+tokentracker update       # Dieses CLI aktualisieren
 ```
 
 ### 🍺 Homebrew (macOS)
--- a/README.ja.md
+++ b/README.ja.md
@@ -67,6 +67,7 @@ tokentracker              # ダッシュボードを開く
 tokentracker sync         # 手動同期
 tokentracker status       # hook の状態を確認
 tokentracker doctor       # ヘルスチェック
+tokentracker update       # この CLI を更新
 ```
 
 ### 🍺 Homebrew (macOS)
--- a/README.ko.md
+++ b/README.ko.md
@@ -67,6 +67,7 @@ tokentracker              # 대시보드 열기
 tokentracker sync         # 수동 동기화
 tokentracker status       # hook 상태 확인
 tokentracker doctor       # 헬스 체크
+tokentracker update       # 이 CLI 업데이트
 ```
 
 ### 🍺 Homebrew (macOS)
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ tokentracker status       # Check hook status
 tokentracker status --json     # Machine-readable summary (pipe to jq, ingest from AI agents)
 tokentracker status --light    # Plain ASCII table (CI / SSH, no spinner)
 tokentracker doctor       # Health check
+tokentracker update       # Update this CLI
 ```
 
 ### 🍺 Homebrew (macOS)
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -69,6 +69,7 @@ tokentracker status       # 查看 hook 挂接状态
 tokentracker status --json   # 机器可读 JSON（pipe 到 jq、给 AI agent 喂数据）
 tokentracker status --light  # 纯 ASCII 表（CI / SSH 用，无 spinner）
 tokentracker doctor       # 健康检查
+tokentracker update       # 更新本 CLI
 ```
 
 ### 🍺 Homebrew（macOS）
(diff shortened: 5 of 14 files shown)


xiufengsun (changes requested):
The update command currently mistakes project-local installs for global installs. Reproduced on this head with detectInstallChannel:

- /project/node_modules/tokentracker-cli/bin/tracker.js -> npm install -g tokentracker-cli@latest
- /project/node_modules/.pnpm/tokentracker-cli@0.96.2/node_modules/tokentracker-cli/bin/tracker.js -> pnpm add -g tokentracker-cli@latest

Both leave the running project dependency unchanged while modifying a different installation. Please require evidence that the resolved entry belongs to the package manager's active global root/prefix, and refuse project-local/ambiguous installs. Also cover a global copy from a different Node/npm prefix: selecting npm from PATH alone can update a different copy. Add regression tests for these cases before enabling mutation.

paveezheng:
Addressed the install-ownership feedback in a2a72668.\n\n`tokentracker update` now verifies that its resolved entry belongs to the active npm/pnpm/yarn/bun global root before invoking an updater. Project-local installs, mismatched global prefixes, and unverified layouts now refuse without mutation. Added regression coverage for local npm/pnpm installs and mismatched npm prefixes.\n\nCI is green.

xiufengsun:
The ownership verification from a2a72668 does what I asked, and the channel matrix is right on macOS and Linux. I checked brew, the three desktop bundles, npx and a source checkout individually. All of them refuse correctly, and brew wins over npm even though the keg path contains `/node_modules/tokentracker-cli/`.

One blocker: **on Windows the command refuses to update a perfectly normal npm global install.**

`globalRootFor()` shells out through `commandOutput()`:

```js
// src/lib/install-channel.js:66
const output = runCommand(bin, args, {
  encoding: "utf8",
  stdio: ["ignore", "pipe", "ignore"],
});
```

On Windows there is no `npm.exe`, only `npm.cmd` and `npm.ps1`, and Node's `execFileSync` cannot launch a `.cmd` directly. Older Node throws `ENOENT`; Node >= 18.20.2 / 20.12.2 throws `EINVAL` after the CVE-2024-27980 fix. Either way the `catch` returns `""`, `isWithinPath(resolved, "")` returns false, and the channel degrades to `other`. The user is then told:

> This TokenTracker is not inside the active npm global install ...
> It may be a project-local dependency or belong to a different global prefix.

for an install that is neither. npm, pnpm and yarn are all affected; bun ships a real `.exe` and is fine.

The unit tests pass because they inject an `execFileSync` stub that always succeeds, so the real `.cmd` path is never exercised, and `ci.yml`'s Windows job does not run these files.

This repo already solved this. `src/lib/command-runner.js:49` says so in a comment:

> `useShell` is opt-in per call site. It exists for npm's Windows `.cmd` shims, which Node's spawn cannot execute directly

It also does `where.exe` discovery and quotes the command under shell, because `C:\Users\a&b\` splits at the `&` in cmd.exe. Please route the probe through that module rather than a bare `execFileSync`, and add a Windows probe case plus `test/install-channel.test.js` to the Windows CI job.

While you are in there: `update.js:37` passes `shell: true` on win32, and cmd.exe resolves a bare command name from the current directory first. Running `tokentracker update` inside a directory that contains a hostile `npm.cmd` executes that file. Resolving the absolute path with `where.exe` and spawning it without a shell removes both problems at once. The arguments are all constants, so there is no injection beyond this.

Non-blocking: `printHelp` is exported from `src/cli.js:122` but nothing calls it, and a `--dry-run` that prints the detected channel and the command it would run is cheap and useful for anyone with a custom global prefix.

PR #613, opened 2026-09-10
feat(skills): add one-click "Update all" for stale skills

obsesivegamer:
## Problem

Updating a managed skill is per-skill only: an "update" badge appears on each row whose upstream copy has drifted, and clicking it re-installs that one skill. There is no select-all, so with a large managed set that is one click per skill. On my machine 10 of 128 skills were stale.

The obvious implementation — loop the existing install call — does not work. `installSkill()` fetches the repo tree itself, so N skills costs N calls against GitHub's 60/hour unauthenticated limit.

## Approach

`checkUpdates()` already solved this exact problem: it groups skills by repo, fetches one tree per repo, limits concurrency, and caches. `updateSkills()` reuses that shape, so N skills from one repo cost **one** tree call rather than N. File contents come from `raw.githubusercontent.com`, which is not metered against that quota.

`installSkill()` gains an optional third argument for an already-fetched tree. Its signature stays backward-compatible.

Two properties fall out of holding the tree:

- **Idempotent.** Each skill's signature is re-derived from the tree already in hand, and skills that already match are skipped. A stale badge cannot cause a needless re-download, and re-running is free.
- **Honest partial results.** Installs run sequentially, because `installSkill()` does its own read/modify/write of the registry and concurrent installs would drop each other's entries. The run stops at the first `RateLimitError` and reports what completed instead of emitting a cascade of identical failures.

`checkUpdates()` and `updateSkills()` now share the candidate filter, the cache fingerprint and the repo grouping (`managedUpdateCandidates` / `updateFingerprint` / `groupSkillsByRepo`), because all three feed the same two decisions — how a cached verdict is keyed, and which tree a signature is derived from. Drift between them would put the badge and the update on different answers about the same skill.

## Cache behaviour

The first version of this PR claimed the updates cache needed no explicit invalidation, on the grounds that its fingerprint is built from `sourceSignature` values so a successful update misses it by construction. **That was wrong in both directions**, and the review caught it:

- A run that only **skips** changes no signature, so the fingerprint was unchanged and `checkUpdates()` served the cached "stale" verdict straight back. The badge survived, and clicking `Update all` was a permanent no-op until the 1h TTL expired.
- Even on the success path, the follow-up refresh missed the cache and fetched **one tree per repo all over again**, so a click cost 2R calls against a 60/hour budget rather than R.

`updateSkills()` now records a verdict per skill as it goes (skip → not stale, install → not stale, failure → still stale) and writes them through `writeUpdateCache()`, the single writer both paths share. The refresh that follows is a cache hit, and a skip-only run clears the badge.

The same writer fixes a related bug in `checkUpdates()`: it used to omit any repo whose tree call failed, so one unreachable repo returned an empty map, which the UI read as "nothing is stale" and used to clear every badge *and* the `Update all` button — exactly when a rate-limited user needs something left to retry. Verdicts a round cannot derive now carry forward from the previous map, and entries for skills that are no longer installed are dropped.

## Counting

`updates` maps every **checked** skill to a boolean, `false` entries included. On my machine that map has 47 entries of which 10 are `true`, so counting keys would report 47 stale skills where there are 10. The count is `Object.values(updates).filter(Boolean)` intersected with what is actually installed — the intersection matters because `confirmRemove` refreshes the installed list but not the verdict map, so removing a badged skill used to leave the button reading `(2)` with one badge on screen and send a dead id on click.

Removals now also prune the map, `updateSkills()` returns a result row for **every** requested id (including ones that are no longer installed), and the toast's total comes from the rows the backend returned, so `updated + skipped + failed` always reconciles. Previously a dropped id produced `Updated 0 of 2 skills. 1 failed.`

Note that `hasUpdatesLoaded` in `SkillsPage.jsx` legitimately uses `Object.keys(...).length` as a "have we loaded yet" flag — the two look alike and mean different things. The count is deliberately **not** a `useMemo`: the local-only early return sits between it and the hooks above, and the comment on that return asks for hooks to stay above it.

## Other fixes from review

- **No implicit sweep.** An empty or malformed `ids` used to mean "every managed skill", and `local-api` passed `body.ids` straight in — so one bad field triggered a full re-install of the library. It is now a no-op, and the endpoint only forwards an array. No caller used the sweep.
- **Registry race.** `updateSkills()` re-checks that a skill still exists before installing it, and `installSkill()` re-reads the registry immediately before the mutation instead of working from the copy it read before minutes of downloads. A skill uninstalled mid-run is no longer written back from a stale snapshot. This narrows the window to one block with no `await` in it; there is still no lock.
- **Confirmation.** `Update all` replaces N managed skill directories on disk, local edits included. Bulk remove already confirms, so this does too.
- **Per-row Update now shares the backend call.** It went through `installSkill()` with targets from `listInstalledSkills()` — a list derived from disk that holds only targets which currently resolve, falling back to `claude`+`codex` when empty. Updating a skill with a broken symlink therefore re-synced it to two agents the user may never have chosen. Both buttons now take targets from the registry's intent via `updateSkills([id])`.
- **Translations.** zh and zh-TW carry every other `skills.*` string, so the seven new ones are translated rather than left to fall back to English; both locales are back at 1596/1596. `validate:locale` also enforces a product glossary — under `skills.*` the term **Skill** stays in Latin script — so the wording matches the house style already used by `skills.toast.bulk_removed` ("已移除 {{count}} 个 Skill。"). `de`/`ja`/`ko` are untouched at their existing partial coverage; that validator only gates zh/zh-TW.
- **CodeQL.** The test fetch stubs picked a canned response with `String(url).includes("api.github.com")`, which also matches `api.github.com.example.test` (incomplete URL sanitization). They now compare the parsed hostname exactly.

## Behaviour changes worth a second look

1. Per-row `Update` resolves targets from the registry, so a skill whose targets are empty stays unsynced instead of being re-synced to `claude`+`codex`.
2. `updateSkills(ids)` no longer sweeps when given nothing. If a sweep is ever wanted it should be an explicit flag.
3. `checkUpdates()` returns a merged map, carrying forward verdicts it could not re-derive. Bounded by the same 1h TTL, but a verdict can now outlive the check that produced it.

## Scope

One button in the My-tab header, shown only when something is stale. Selection-based bulk update in `BatchToolbar` is deliberately left out to keep this reviewable; it would reuse the same backend call.

Not included: a locale-coverage check in `validate:copy` (nothing in CI catches an English-only key until `validate:locale`, and only for zh/zh-TW), and topping up `de`/`ja`/`ko`.

## Testing

- **11 tests** in `test/skills-manager-extras.test.js` for this path: per-repo call batching, skip-when-current, id filtering, rate-limit partial progress, cache priming on both the skip and success paths, badge survival when a repo is unreachable, result-row reconciliation, no-sweep id scope, and no resurrection of a skill uninstalled mid-download. Each of the post-review ones fails on the pre-fix code and passes after — verified by running them against the previous implementation (7 failures).
- Full suite **2739 tests, 2737 pass, 2 skipped, 0 fail**. `npm run ci:local` clean, typecheck and dashboard build clean. All 8 CI checks green.
- Exercised in the running app against a sandbox `HOME` (3 skills, 2 marked stale): removing a badged skill moves the button `(2) → (1)` (it used to stick at 2); a failing run leaves both badges and the button in place (they used to vanish, with nothing left to retry); and the failure toast reads `Updated 0 of 2 skills. 2 failed.` instead of the incoherent `0 of 2 … 1 failed`.

## Note

Still a draft — description updated to match the tip (`e13fd37c`) after the review fixes landed.

## Summary by CodeRabbit

- **New Features**
  - Added “Update all” support for installed skills with per-skill success and failure results.
  - Added confirmation prompts and rate-limit notifications for bulk updates.
  - Added Devin provider labels and setup guidance, including login reuse and usage refresh instructions.
  - Added Simplified and Traditional Chinese translations for skill update messages.

- **Bug Fixes**
  - Improved update status tracking when checks fail or skills are removed.
  - Prevented stale updates from overwriting newer skill changes.
  - Disabled skill actions while another update or operation is in progress.


src/lib/skills-manager.js
```diff
@@ -1303,6 +1305,96 @@ async function checkUpdates({ force = false } = {}) {
...
+// Re-install managed skills from upstream, grouped by repo like checkUpdates()
+// so N skills from one repo cost one tree call. Sequential by necessity:
+// installSkill() read/modify/writes the registry, so parallel installs would
+// drop each other's entries. A successful update rewrites the skill's
+// sourceSignature, so the updates cache — fingerprinted on those signatures —
+// misses on the next check without explicit invalidation; a run that only skips
+// changes no signature, so the cached verdict (and its badge) survives.
+async function updateSkills(ids = []) {
+  const registry = readRegistry();
+  const wanted = new Set(Array.isArray(ids) ? ids : []);
```
obsesivegamer: **The guard and the sentinel point in opposite directions.** `Array.isArray(ids) ? ids : []` reads as safety, but `wanted.size === 0` below means *every managed skill*, so any malformed input becomes the maximum-blast-radius action. `local-api.js` then hands in `body.ids || []`, which turns `null`/missing/wrong-type into exactly that.

Measured through the real HTTP handler with two managed skills in the registry:

```
{"action":"update_all","ids":["o/r:a"]}  -> updated 1   (correct)
{"action":"update_all"}                  -> updated 2   (all)
{"action":"update_all","ids":"o/r:a"}    -> updated 2   (all)
{"action":"update_all","ids":[]}         -> updated 2   (all)
```

On the 128-skill box from your description that is 128 directory replacements plus a burst of tree calls from one malformed field. The sweep mode also has no caller today — `handleUpdateAll` early-returns on an empty list — so it is currently unreachable-by-design and dangerous-by-accident.

Suggest: non-array or empty → no-op, and add an explicit `{ all: true }` later if you ever want the sweep.

Related: in sweep mode the `skill.sourceSignature &&` precondition on the skip check means signature-less legacy entries are never skipped, and one whose upstream path is nested fails with `SKILL.md not found in selected directory`. Both only reachable through this.

src/lib/skills-manager.js
```diff
@@ -1303,6 +1305,96 @@ async function checkUpdates({ force = false } = {}) {
...
+    }
+    if (rateLimited) break;
+  }
+
+  return {
+    results,
+    updated: results.filter((entry) => entry.ok && !entry.skipped).length,
+    skipped: results.filter((entry) => entry.skipped).length,
+    failed: results.filter((entry) => !entry.ok).length,
+    rateLimited: rateLimited ? rateLimited.message : null,
```
obsesivegamer, line 1497: **You already hold everything `checkUpdates` is about to re-fetch.** Writing `updates-cache.json` here — fingerprint from the post-install registry, verdicts from the signatures computed in the loop above — fixes two things at once.

**a) Cost.** `handleUpdateAll` calls `loadUpdates()` → `checkSkillUpdates()` un-forced → the fingerprint changed, so the cache misses and `checkUpdates` fetches **one tree per repo all over again**. Measured over 2 repos:

```
tree calls: updateSkills = 2 | follow-up checkUpdates = 2
follow-up check cached? false
```

So a click costs 2R against the 60/hour unauthenticated budget, spent right after the run has been drawing on it. The description's *N skills from one repo cost one tree call* is true of `updateSkills` in isolation but not of the click.

**b) Correctness.** The description says the cache *needs no explicit invalidation … by construction*. That holds for the success path only. A run that **skips** changes no `sourceSignature`, so the fingerprint is identical and `checkUpdates` re-serves the cached verdict:

```
checkUpdates (cached, says stale): {"o/r:a":true}
updateSkills -> {"updated":0,"skipped":1,"failed":0}
checkUpdates after update:        {"o/r:a":true}   cached: true
>>> badge still stale after clicking Update all? true
```

The button keeps reading `Update all (1)`, clicking is a permanent no-op, and it only clears when the 1h TTL expires. The skip path is the feature you introduced here, so it's the one case the cache reasoning doesn't cover.

src/lib/skills-manager.js
```diff
@@ -1303,6 +1305,96 @@ async function checkUpdates({ force = false } = {}) {
...
 
+// Re-install managed skills from upstream, grouped by repo like checkUpdates()
+// so N skills from one repo cost one tree call. Sequential by necessity:
+// installSkill() read/modify/writes the registry, so parallel installs would
+// drop each other's entries. A successful update rewrites the skill's
+// sourceSignature, so the updates cache — fingerprinted on those signatures —
+// misses on the next check without explicit invalidation; a run that only skips
+// changes no signature, so the cached verdict (and its badge) survives.
+async function updateSkills(ids = []) {
+  const registry = readRegistry();
```
obsesivegamer: **This snapshot outlives the writes that follow it.** `updateSkills` reads the registry once, then `installSkill` re-reads and rewrites it per skill. Anything that lands in between is overwritten from the stale snapshot.

Probe: delete an entry mid-run (simulating an uninstall from another surface) and the run writes it back.

```
results: ["o/r:a","o/r:c"]
registry ids after run: [ 'o/r:a', 'o/r:c' ]  <- 'o/r:c' resurrected? true
```

The window matters more than it used to. A single install is milliseconds; a sequential run over a large stale set is minutes, and `saveRegistry` is a plain `writeFileSync` — no tmp+rename — so a concurrent reader can also catch a truncated file. The endpoint has no lock and the menu-bar app plus a second dashboard tab both reach it; `busyKey` only serializes within one page.

Cheapest fix: re-read and re-check membership immediately before each `installSkill`, and skip entries that have disappeared.

src/lib/skills-manager.js
```diff
@@ -1303,6 +1305,96 @@ async function checkUpdates({ force = false } = {}) {
...
+  const wanted = new Set(Array.isArray(ids) ? ids : []);
+  const managed = registry.skills.filter(
+    (skill) =>
+      !skill.trashedAt &&
+      skill.repoOwner &&
+      skill.repoName &&
+      (wanted.size === 0 || wanted.has(skill.id)),
+  );
+
+  const byRepo = new Map();
```
obsesivegamer: These 9 lines are byte-identical to `checkUpdates` (1278-1286) — I diffed them. Worth one `groupManagedByRepo(skills)` helper, not for brevity but because the two **must** agree: the grouping key decides which tree a skill's signature is derived from, so any future drift makes the badge and the update disagree about the same skill.

src/lib/skills-manager.js
```diff
@@ -1303,6 +1305,96 @@ async function checkUpdates({ force = false } = {}) {
...
+            name: skill.name,
+            description: skill.description,
+            // installSkill() wants the repo path, not the flat install name.
+            directory: sourceDir,
+            readmeUrl: skill.readmeUrl,
+            repoOwner: skill.repoOwner,
+            repoName: skill.repoName,
+            repoBranch: branch,
+          },
+          skill.targets || [],
```
obsesivegamer, line 1457: Reading intended targets from the registry is the right call — better than the per-row path, which is worth aligning while this is fresh.

`handleUpdate` receives its `skill` from `listInstalledSkills()`, where `targets` is **disk-derived** (only entries whose `classifyTargetSkill` is `"synced"`), and then falls back to `DEFAULT_TARGETS` when that list is empty. So for a skill whose symlink is broken, the row button silently re-syncs it to claude+codex, while Update all restores its real targets from the registry.

Two buttons labelled "update" that resolve targets differently is the kind of thing that gets discovered as a bug report six months out. Confirmed preserved here: `["claude"]` stays `["claude"]`, `[]` stays `[]`, and only `alpha` appears under `.claude/skills`.

src/lib/local-api.js
```diff
@@ -3160,6 +3160,11 @@ function createLocalApiHandler({ queuePath }) {
             json(res, { ok: true, skill: await skills.installSkill(body.skill, body.targets || ["claude", "codex"]) });
             return true;
           }
+          if (action === "update_all") {
```
obsesivegamer, line 3163: `body.ids || []` is the other half of the `Array.isArray` issue at `skills-manager.js:1317` — it maps a missing or null `ids` onto the sweep-everything path. Worth validating here too (array of strings, non-empty) so the manager isn't the only thing standing between a malformed request and 128 re-installs.

The comment is accurate and the auth gate covers this action — unauthenticated POST returns 401, confirmed.

dashboard/src/pages/SkillsPage.jsx
```diff
@@ -1232,6 +1254,40 @@ export function SkillsPage() {
...
+    if (!ids.length) return;
+    runMutation("update-all", async () => {
+      const result = await updateSkills(ids);
+      await loadUpdates();
+      const { updated = 0, failed = 0, rateLimited = null, results = [] } = result || {};
+      if (rateLimited) {
+        showToast({ title: copy("skills.update.rate_limited", { count: updated }), timeout: 6000 });
+      } else if (failed > 0) {
+        showToast({
+          title: copy("skills.toast.updated_partial", { count: updated, total: ids.length, failed }),
```
obsesivegamer: `total: ids.length` counts ids the backend silently dropped, so the toast can print arithmetic that doesn't add up.

`confirmRemove` refreshes `installed` but not `updates`, so removing a badged skill leaves its `true` in the map. In the running app I removed `alpha` from a 3-skill set with 2 stale: the list dropped to `2 of 2 skills` with one badge left, and the button still read **`Update all (2)`**. Clicking it sent the dead id, which `updateSkills` drops with no result row and no failure (confirmed through HTTP: `{"results":[],"updated":0,"skipped":0,"failed":0}`), and the toast rendered:

> **Updated 0 of 2 skills. 1 failed.**

0 + 1 ≠ 2. Two fixes, both cheap: call `loadUpdates()` after a removal so the count can't over-report, and report from `results` rather than `ids.length` so the numbers always reconcile — ideally with `updateSkills` returning a row for ids it couldn't match.

obsesivegamer (commented):
## Review: `feat/skills-update-all` — verdict **ISSUES**

The core design holds up. Grouping by repo really does cost one tree call per repo, the skip path is idempotent, the rate-limit stop keeps completed work, and the `Object.values(updates).filter(Boolean)` subtlety you flagged is handled correctly — I confirmed the button reads `Update all (2)` with 3 skills checked and 2 stale, matching the badges exactly.

What I found sits at the edges this feature newly exposes: an `ids` guard that fails toward *update everything*, an updates cache that isn't primed from trees already in hand, and an `updates` map the UI now **counts** rather than only indexes.

### How I verified

- `npm test` → **2730 pass, 2 skipped, 0 fail**. `tsc --noEmit` clean, `dashboard:build` clean, `validate:copy` / `validate:ui-hardcode` / `validate:guardrails` all pass. Every testing claim in the description checks out.
- A runtime probe driving `updateSkills()` against a stubbed `global.fetch` with call counters: legacy entries, malformed `ids`, mid-repo failure, nested `sourceDirectory`, updates-cache interaction, concurrent registry writes.
- The same actions through the real `/functions/tokentracker-skills` handler (auth gate included).
- The real app: `node bin/tracker.js serve --port 7699` against a sandbox `HOME` seeded with 3 managed skills, 2 marked stale.

### Findings, ranked

| # | Finding | Where |
|---|---|---|
| 1 | Malformed or empty `ids` re-installs **every** managed skill | `skills-manager.js:1317`, `local-api.js:3163` |
| 2 | Updates cache not primed from the trees in hand → 2× tree calls per click, and a skip-only run leaves the stale badge up | `skills-manager.js:1394` |
| 3 | Stale `updates` map → count over-reports and the toast math goes incoherent | `SkillsPage.jsx:1274` |
| 4 | A failed run wipes every update signal, leaving no retry affordance | `SkillsPage.jsx:1268` |
| 5 | Registry read/modify/write race resurrects a concurrently-uninstalled skill | `skills-manager.js:1316` |
| 6 | Bulk destructive overwrite with no confirmation (bulk *remove* has one) | `SkillsPage.jsx:562` |
| 7 | Repo-grouping block duplicated byte-for-byte from `checkUpdates` | `skills-manager.js:1326` |
| 8 | Two update paths, two different target semantics | `skills-manager.js:1374` |
| 9 | 5 new strings are English-only; zh/zh-TW were otherwise ~complete | `copy.csv:1224` |

1–4 are reachable on ordinary paths and three of them are visible to the user. Details inline.

### What I confirmed works

- **Batching**: one tree call per repo regardless of skill count — 2 repos / 3 skills → 2 calls; 1 repo / 1 skill through HTTP → 1 call.
- **Skip-when-current**: second run does zero raw downloads (`updated: 0, skipped: 1, raw fetches: 0`). Re-running really is free.
- **Rate-limit partial**: stops at the first `RateLimitError`, keeps completed work, `failed` correctly stays 0.
- **Mid-repo failure** doesn't abort the rest of the repo — one skill fails, the next still updates.
- **Targets** are preserved from the registry, including the empty case, and nested `sourceDirectory` resolves to the right flat install name with no duplicate registry entry.
- **Auth**: the new action is behind `isAuthorizedLocalMutation` — unauthenticated POST → 401.

### Note on the extra commit

I pushed one comment-only commit (`687068ee`). The three comments added here were the only ` -- ` dashes in their files (each uses `—` everywhere else: 8, 9 and 1 occurrences respectively), and the `updateSkills` header's cache claim is wrong for the skip path — see finding 2. No behaviour changed; I deliberately left every bug above for you rather than fixing them under a cleanup commit. Beyond that the diff is clean — the comments earn their place and I found no boilerplate or defensive padding worth stripping.

Not merging — draft, and 1–4 want decisions from you first.

dashboard/src/pages/SkillsPage.jsx
```diff
@@ -1232,6 +1254,40 @@ export function SkillsPage() {
...
+  const updateCount = useMemo(() => Object.values(updates).filter(Boolean).length, [updates]);
+
+  const handleUpdateAll = () => {
+    const ids = Object.entries(updates)
+      .filter(([, stale]) => stale)
+      .map(([id]) => id);
+    if (!ids.length) return;
+    runMutation("update-all", async () => {
+      const result = await updateSkills(ids);
+      await loadUpdates();
```
obsesivegamer, line 1302: **A failed refresh here destroys the state the user needs to retry.** `loadUpdates` catches and does `setUpdates({})`, and `checkUpdates` drops a whole repo on any tree error (rethrowing on `RateLimitError`). So one failure after the run clears every badge *and* the button.

Observed in the app: after a run where the update failed, the toast appeared and then both the button and all row badges vanished. The rate-limit copy says *Try again in an hour* — but there is nothing left to click, and `hasUpdatesLoaded` is false so the mount effect won't restore it either; only **Refresh** will.

Keeping the previous map on failure (`setUpdates(prev => prev)` in the catch, or only clearing on an explicit refresh) preserves the affordance. Priming the cache in `updateSkills` (see the `skills-manager.js` comment) would mean this call is a cache hit and usually can't fail at all.

dashboard/src/pages/SkillsPage.jsx
```diff
@@ -543,18 +546,37 @@ function MySkillsView({
...
+            onAgentFilter={onAgentFilter}
+            filteredCount={items.length}
+            totalCount={totalCount}
+            anyFilter={anyFilter}
+            onClearFilters={onClearFilters}
+            searchQuery={searchQuery}
+            onSearchQuery={onSearchQuery}
+            searchPlaceholder={searchPlaceholder}
+          />
+          {updateCount ? (
```
obsesivegamer, line 562: One click here replaces N managed skill directories on disk — `installSkill` does `removePath(dest)` then rename, so any local edit to a managed skill's SSOT copy is gone. `handleBulkRemove` guards a comparable action with a `ConfirmModal`; this doesn't. `contentHash` is written at install but never read anywhere, so there's no drift detection to soften it either.

Given the count can over-report (see the toast comment), the user may also be confirming a number that isn't real.

Two smaller notes on this block, neither blocking:
- The button lives in the `else` of `selectionCount > 0`, so it disappears as soon as anything is checked. Consistent with the scope note, just worth knowing.
- `updateCount` is global while the `FilterToolbar` beside it reads *X of Y*. Defensible for a button named "Update all", but under a filter the number won't match the badges on screen.
- No progress signal for a run that can take minutes over a large set. No client-side timeout, so it won't abort — just a long spinner.

dashboard/src/content/copy.csv
```diff
@@ -1214,10 +1214,15 @@ skills.toast.bulk_removed,ui,SkillsPage,SkillsPage,toast_bulk_removed,"Removed {
...
 skills.toast.unsynced_one,ui,SkillsPage,SkillsPage,toast_unsynced_one,"Unsynced {{name}} from {{agent}}.",,active
 skills.toast.updated,ui,SkillsPage,SkillsPage,toast_updated,"Updated {{name}}.",,active
+skills.toast.updated_many,ui,SkillsPage,SkillsPage,toast_updated_many,"Updated {{count}} skills.",,active
+skills.toast.updated_partial,ui,SkillsPage,SkillsPage,toast_updated_partial,"Updated {{count}} of {{total}} skills. {{failed}} failed.",,active
+skills.toast.updated_none,ui,SkillsPage,SkillsPage,toast_updated_none,"All skills are already up to date.",,active
 skills.confirm.bulk_remove_title,ui,SkillsPage,SkillsPage,confirm_bulk_remove_title,"Remove {{count}} skills?",,active
 skills.confirm.bulk_remove_desc,ui,SkillsPage,SkillsPage,confirm_bulk_remove_desc,"This unsyncs each skill from every agent and deletes its local copy.",,active
 skills.update.available,ui,SkillsPage,SkillDetailPanel,update_available,"A newer version is available upstream.",,active
 skills.update.action,ui,SkillsPage,SkillDetailPanel,update_action,"Update",,active
+skills.update.all_action,ui,SkillsPage,SkillsPage,update_all_action,"Update all ({{count}})",,active
```
obsesivegamer, line 1228: These 5 new strings are English-only. `zh` and `zh-TW` carry 98 of the 103 `skills.*` keys in `copy.csv` — the 5 added here are exactly the gap, so a Chinese user gets a `Update all (10)` button and English toasts on an otherwise translated page. (`de`/`ja`/`ko` sit at 56, so they're already partial and this changes little there.)

`validate:copy` only checks key usage, not locale coverage, so nothing in CI catches this.

src/lib/skills-manager.js
obsesivegamer: Fixed in `997f8082`. The sweep is gone: `ids` is now normalised to a set of non-empty strings and an empty result short-circuits to a no-op before any network call. No caller used the sweep, so nothing needed an opt-in replacement.

`test/skills-manager-extras.test.js` covers `[]`, `undefined`, `null`, a bare string, an object, `[""]` and `[null]` — each must report zero rows and make zero GitHub calls with a stale skill sitting in the registry.

The knock-on you flagged goes with it: signature-less legacy entries and nested-path entries were only reachable through the sweep.

src/lib/skills-manager.js
obsesivegamer: Fixed in `997f8082` — both halves.

`updateSkills()` now records a verdict per skill as it goes (skip → false, install → false, failure → true) and writes them to the same cache `checkUpdates()` reads. The follow-up refresh is a cache hit, so a click over R repos costs R tree calls instead of 2R.

The skip-only hole is closed by the same change: a run that only skips still writes `false` for those skills, so the badge clears even though no signature moved. Test asserts exactly your reproduction — cached verdict says stale, the run skips, the next check must return `false` and spend zero tree calls.

Both paths now go through one `writeUpdateCache()` so the check and the update cannot write the cache differently.

src/lib/skills-manager.js
obsesivegamer: Fixed in `997f8082`, at both levels.

`updateSkills()` re-checks that a skill is still in the registry immediately before installing it. That alone only narrows the window, so `installSkill()` now re-reads the registry right before the mutation instead of working from the copy it read before the downloads — the read/modify/write is down to one block with no `await` in it. That helps the single-install path too, which had the same shape.

Test drops a skill from the registry mid-download (the same shape as your probe) and asserts it stays uninstalled while its sibling still updates.

Not claiming the window is mathematically zero — there is no lock — but it is no longer minutes wide.

src/lib/skills-manager.js
obsesivegamer: Done in `997f8082`. `checkUpdates()` and `updateSkills()` now share `managedUpdateCandidates()`, `updateFingerprint()` and `groupSkillsByRepo()`, for the reason you gave rather than for the line count: all three feed the cache key and the tree a signature is derived from, so drift between them would put the badge and the update on different answers.

src/lib/skills-manager.js
obsesivegamer: Aligned in `8985bec6` — by moving the per-row button onto this path rather than the reverse.

`handleUpdate` now calls `updateSkills([skill.id])`, so both buttons take targets from the registry's intent. The disk-derived list plus the `DEFAULT_TARGETS` fallback is gone, which removes the case you described: updating a skill with a broken symlink silently re-synced it to claude+codex.

Worth flagging as a deliberate behaviour change — a skill whose registry targets are empty now stays unsynced through an update instead of being re-synced to the two defaults.

src/lib/local-api.js
obsesivegamer: Fixed in `997f8082`. The endpoint now forwards `Array.isArray(body.ids) ? body.ids : []`, and the manager treats an empty list as a no-op, so neither side alone can turn a malformed field into a full re-install.

dashboard/src/pages/SkillsPage.jsx
obsesivegamer: Fixed in `8985bec6`.

`total` now comes from `results.length` — the rows the backend actually returned — and `updateSkills()` emits a row for every requested id including ones that are no longer installed, so `updated + skipped + failed` always reconciles with the total.

The root cause is fixed too: removals now prune the verdict map, and the count and the id list are the intersection of the map with what is installed, so no single removal path has to remember to do it.

Verified live on a sandbox `HOME`: 3 skills, 2 stale, removed one → the button went from `Update all (2)` to `Update all (1)` (it stuck at 2 before), and a failing run reported `Updated 0 of 2 skills. 2 failed.` instead of `0 of 2 … 1 failed`.

dashboard/src/pages/SkillsPage.jsx
obsesivegamer: Fixed in `997f8082` + `8985bec6` — and the sandbox probe showed my first attempt was not enough, which was worth the run.

Preserving the map in the `catch` only covers a thrown error. The common case is `checkUpdates()` *succeeding* with an empty map: it omitted any repo whose tree call failed, so one unreachable repo returned `{}` and the UI read that as "nothing is stale". Both fixed — `checkUpdates()` now carries forward the previous verdict for anything it could not derive, and the `catch` keeps what it has.

Live check: a failed Update all now leaves both badges and the button in place. Test asserts a check where every tree call 404s still reports the previously known update rather than clearing it.

dashboard/src/pages/SkillsPage.jsx
obsesivegamer: Added in `8985bec6`. `Update all` now opens a `ConfirmModal` mirroring bulk remove — "Update N skills?" / "Each skill is re-downloaded from its source repository, replacing the local copy." Not styled `destructive`, since that red is the removal affordance.

The count it confirms is now trustworthy (see the toast thread), which was the other half of the problem.

The three smaller notes are deliberately left as they are: the button hiding behind `BatchToolbar` during a selection matches the scope note, the global-count-under-a-filter reading is defensible for a button called "Update all", and progress reporting for a long run is a bigger change than this PR wants.

dashboard/src/content/copy.csv
obsesivegamer: Fixed in `8985bec6`. `zh` and `zh-TW` now carry all 105 `skills.*` keys — the five from this PR plus the two new confirm strings — so those locales are back at full parity.

`de`/`ja`/`ko` are left at their existing partial coverage; they sit around half the skills keys already, and topping them up is a separate job from this PR.

You are right that nothing in CI catches this. Worth a locale-coverage check in `validate:copy` as a follow-up, but I have not added one here.

obsesivegamer:
## Review fixes pushed — `997f8082` (backend) + `8985bec6` (dashboard)

All nine findings addressed, replies on each thread. Comment cleanup stayed in its own earlier commit (`687068ee`); these two are behaviour only.

### Verification

Every fix has a test that fails on the pre-fix code and passes after. I checked that directly rather than assuming it: running the new tests against `HEAD`'s `skills-manager.js` gives **7 failures**, against the fixed one **0**.

Suite is **2739 tests, 2737 pass, 2 skipped, 0 fail** (was 2732/2730). `tsc --noEmit`, `dashboard:build`, `validate:copy` / `validate:ui-hardcode` / `validate:guardrails` all clean.

The two user-visible bugs were re-checked in the running app — sandbox `HOME`, 3 skills, 2 marked stale:

| | before | after |
|---|---|---|
| Remove a badged skill | button stuck at `Update all (2)` with one badge on screen | `Update all (1)` |
| A run that fails | every badge and the button vanish; nothing to retry | both survive |
| Failed-run toast | `Updated 0 of 2 skills. 1 failed.` | `Updated 0 of 2 skills. 2 failed.` |

The live probe earned its keep on the wipe bug: preserving the map in `loadUpdates`'s `catch` was **not** sufficient, because the common path is `checkUpdates()` *succeeding* with an empty map when a repo's tree call fails. Reasoning alone would have shipped a half-fix.

### One thing I missed the first time

`updateCount` was a `useMemo` sitting **after** the local-only early return — below the comment that says hooks are placed above it so the order stays stable. Harmless today (both guards are session-constants), a crash the day either becomes dynamic. It is now a plain computation, no hook.

### Behaviour changes worth a second look

- **Per-row Update resolves targets differently.** It now goes through `updateSkills([id])`, so a skill whose registry targets are empty stays unsynced instead of being re-synced to claude+codex. That is the fix for the broken-symlink case, but it is a real change to an existing button.
- **`updateSkills(ids)` no longer sweeps.** An empty or malformed list is a no-op rather than "every managed skill". No caller used the sweep; if one is ever wanted it should be an explicit flag.
- **`checkUpdates()` returns a merged map**, carrying forward verdicts it could not re-derive. Bounded by the same 1h TTL as before, but it does mean a verdict can outlive the check that produced it.

### Left undone

- `de`/`ja`/`ko` stay at their existing partial skills coverage; `zh`/`zh-TW` are back to full parity.
- No locale-coverage check added to `validate:copy` — worth doing, but it belongs outside this PR.
- The registry has no lock. The RMW window is down from minutes to one block with no `await` in it, which is a real fix, not a proof of exclusion.

Still a draft, not merged.

src/lib/skills-manager.js
```diff
@@ -992,9 +994,13 @@ async function installSkill(skillInput, targetIds = ["claude", "codex"]) {
...
-  registry.skills = registry.skills.filter((entry) => entry.id !== installed.id && entry.directory.toLowerCase() !== installName.toLowerCase());
-  registry.skills.push(installed);
-  saveRegistry(registry);
+  // Re-read: the downloads above can take minutes and the registry is shared
+  // with every other mutation path, so the copy fetched before them is stale by
+  // now. Reading here keeps the read/modify/write window down to this block.
+  const latest = readRegistry();
+  latest.skills = latest.skills.filter((entry) => entry.id !== installed.id && entry.directory.toLowerCase() !== installName.toLowerCase());
+  latest.skills.push(installed);
+  saveRegistry(latest);
```
obsesivegamer, line 1019: **The re-read still writes the skill back.** `updateSkills()` checks membership *before* `installSkill()`, then this block filters-by-id and `push`es unconditionally. The review test drops a *sibling* during the first skill's download — that path still works. Dropping the skill that is actually downloading does not.

Probe against `e13fd37c`, stubbed `fetch`, skill uninstalled on the first raw GET:

```
updateSkills(["o/r:a"]) → {updated:1, skipped:0, failed:0}
registry ids after: ["o/r:a"]   resurrected? true
```

Same write also clobbers a replacement that landed on the same `directory` during the download (`other/x:a` gone, `o/r:a` back). The minutes-wide snapshot bug is gone. The skill under the `await` is not.

src/lib/skills-manager.js
```diff
@@ -1299,8 +1347,109 @@ async function checkUpdates({ force = false } = {}) {
...
-  return { updates, checkedAt, cached: false };
+  return { updates: writeUpdateCache(updates), checkedAt, cached: false };
+}
+
+// Re-install managed skills from upstream, grouped by repo like checkUpdates()
+// so N skills from one repo cost one tree call. Sequential by necessity:
+// installSkill() read/modify/writes the registry, so parallel installs would
+// drop each other's entries. Every requested id gets a result row — including
+// ones that are no longer installed — so updated + skipped + failed always
+// reconciles with what the caller asked for.
```
obsesivegamer, line 1391: This sentence is not true of a rate-limit stop. Both `break`s (tree fetch ~1386, raw fetch ~1430) leave the current skill and every later id without a row. `failed` stays 0.

Same stubbed `fetch` against the tip:

```
first tree 403, ids [a, c]  → results=[]           updated=0 failed=0 rateLimited=set
first raw  403, ids [a, c]  → results=[]           updated=0 failed=0 rateLimited=set
repo1 ok, repo2 403         → results=[a]          updated=1 failed=0  // c omitted
```

Ghost-id rows (the test you added) still work. The rate-limit path is the one the toast and the per-row button actually hit. Either emit a row for every remaining id, or stop claiming the counts reconcile with what was asked for.

dashboard/src/pages/SkillsPage.jsx
```diff
@@ -1207,31 +1247,72 @@ export function SkillsPage() {
...
-        skill.targets && skill.targets.length ? skill.targets : DEFAULT_TARGETS,
-      );
+      const result = await updateSkills([skill.id]);
       await loadUpdates();
+      const row = (result?.results || [])[0];
+      if (row && !row.ok) throw new Error(row.error || copy("skills.error.generic"));
       showToast({
         title: copy("skills.toast.updated", { name: skill.name || skill.directory }),
         timeout: 4000,
       });
```
obsesivegamer, line 1276: `updateSkills` reports rate-limit as a return field, not a throw, and not as a result row (see the other comment). This only errors when `results[0]` exists and `!ok`.

Composing the two: a single-skill Update whose tree or first raw GET is 403 returns `{results:[], rateLimited: "..."}`. `row` is `undefined`, no throw, toast is **Updated {{name}}**. The badge then stays (cache not primed, `loadUpdates` cache-hits the old `true`) so the user sees a success toast and a still-stale row.

`handleUpdateAll` already branches on `rateLimited` first. This path needs the same check, and should not toast success on an empty `results`.

obsesivegamer (commented):
## Recheck `e13fd37c` — verdict **ISSUES**

The nine original findings are not all still fixed. Cache priming, the empty-`ids` no-op, removed-id sweep, confirm modal, per-row targets, zh/zh-TW glossary, and the CodeQL host check hold. I re-ran the extras file (24/24) and probed the claimed edges against the tip instead of trusting the write-up.

### Still holds (re-proven)

| Claim | Evidence |
|---|---|
| Empty / malformed `ids` is a no-op | extras + HTTP: `[]` / missing / `"o/r:a"` → `{updated:0, rows:0}`, zero GitHub calls |
| `update_all` auth | unauthenticated POST → 401 |
| Skip-only clears the cached stale badge | extras + probe: next `checkUpdates` is `false` and a cache hit |
| Success path primes the cache (1 tree / repo, not 2) | extras |
| Unreachable-repo badge survives *within the 1h TTL* | extras |
| Ghost / removed ids dropped from the cache and from `updateCount` | probe: `o/r:ghost` gone, sibling `true` carried; UI intersects with installed ids |
| Confirm modal present; per-row Update goes through `updateSkills([id])` | read |
| zh / zh-TW 1596/1596, `Skill` stays Latin | `validate:locale` |
| Fetch stubs match hostname, not substring | `isGitHubApi`; no `includes("api.github.com")` left |
| CI on this tip | 7/7 green |

### Does not hold

1. **Registry race — wrong skill was tested.** Sibling-drop during the *other* download is fixed. The skill under the `await` is written back by `installSkill()`'s re-read+`push`. Same write deletes a same-directory replacement that landed mid-download. Details + numbers on `skills-manager.js`.
2. **"Every requested id gets a result row" is false on a rate-limit stop.** Both `break`s omit the current skill and everything after it (`failed` stays 0). Ghost-id rows still work; that is the only case the new test covers.
3. **Per-row Update toasts success on that empty result.** `handleUpdateAll` already branches on `rateLimited`. `handleUpdate` does not, so a 403 looks like "Updated {{name}}" while the badge stays.

### Light deslop (not blocking)

- `updateSkills` still says `local-api` "feeds `body.ids` straight in". The endpoint now wraps with `Array.isArray`. The comment is leftover from the bug it describes.
- `entry` / `stub` / `TREE` are copied three times in the extras file. Fine for a first pass; not load-bearing.
- `ConfirmModal busy={busyKey === "update-all"}` never shows: `confirmUpdateAll` nulls `pendingUpdateAll` before `runMutation`. Same dead pattern as bulk-remove; do not copy it further.

`runMutation` still has no mutex, so a per-row Update can overlap an in-flight Update all. Pre-existing shape; the new minutes-long run makes it easier to hit. Not probed at runtime — read only.

Not merging, not undrafting.

src/lib/skills-manager.js
obsesivegamer: Fixed in `503b7a31`. You were right that the check was in the wrong place, not just testing the wrong skill — a re-check before `installSkill()` cannot see a drop that happens during it.

`installSkill()` now takes an `expectExistingId` option and verifies it **after the last `fetchText` and before `removePath(dest)`**. Everything from that check through `saveRegistry()` is synchronous — rename, hash, re-read, filter, push, write — so nothing can interleave between deciding to commit and committing. Default is `null`, so the plain install path is unchanged.

Both halves of your repro are now regression tests, and both fail on `e13fd37c`:

```
uninstall on first raw GET   → {updated:0}     registry: o/r:a absent
other/x:a lands on same dir  → other/x:a kept, o/r:a not resurrected
```

Caveat worth stating: this closes the in-process window, not a cross-process one. There is still no lock on `registry.json`, so a second Token Tracker process could interleave between that `readRegistry()` and `saveRegistry()`.

src/lib/skills-manager.js
obsesivegamer: Fixed in `503b7a31`, and you are right that the sentence was the problem as much as the code — it claimed an invariant the rate-limit path broke.

Both `break`s now fall through to a backfill that emits a row for every requested id that did not get one, carrying the limit message. All three of your shapes are table-driven tests, each failing on `e13fd37c`:

```
first tree 403        → rows [a, c]   updated 0
first raw 403         → rows [a, c]   updated 0
repo1 ok, repo2 403   → rows [a, c]   updated 1
```

with `updated + skipped + failed === results.length === ids.length` asserted in each.

That also meant changing an existing test rather than adding to it: `stops at a rate limit and reports partial progress` asserted `failed === 0` on the grounds that "a rate limit is not a per-skill failure". That encoded the bug. An id we never got to is a failure to update, not a silent drop, so it now asserts `results.length === 2` and `failed === 1`.

Separately, the stale `local-api` "feeds `body.ids` straight in" comment you flagged is gone in `1aca5e2f` — the endpoint wraps with `Array.isArray` now, so the sentence described the bug rather than the code under it.

dashboard/src/pages/SkillsPage.jsx
obsesivegamer: Fixed in `503b7a31`. Composing the two failures was the part I had missed — `handleUpdate` looked defensible in isolation and only misreports because `updateSkills` returned no rows.

It now branches on `rateLimited` first, exactly like `handleUpdateAll`, and treats an empty `results` as an error rather than falling through to the success toast:

```js
if (rateLimited) { showToast({ title: copy("skills.update.rate_limited", { count: updated }) }); return; }
const row = results[0];
if (!row || !row.ok) throw new Error(row?.error || copy("skills.error.generic"));
```

Covered by three tests in `dashboard/src/pages/SkillsPage.test.jsx` — rate limit, empty result, and a normal update as the control. The first two fail on `e13fd37c` with the toast reading `Updated Alpha Skill.`, which is your finding verbatim. Writing them surfaced a second thing: the shared fixture's skills carry no `repoOwner`/`repoName`, so `handleUpdate` returned early and even the control failed. The fixture override is scoped to the new `describe` so the existing eight tests are untouched.

Two things I did not do. The `ConfirmModal busy` dead prop is removed in `1aca5e2f` (bulk-remove's copy left alone as pre-existing, per your note). And `SkillsPage.test.jsx` is **not** in CI's curated vitest list, so these tests do not run there — I had a workflow step for it, but the push was rejected for lack of `workflow` OAuth scope and I dropped it rather than leave the branch unpushable. Worth adding separately.

On `runMutation` having no mutex: agreed, and not addressed here. A per-row Update can still overlap an in-flight Update all; fixing it properly is a real lock, not a patch.

src/lib/skills-manager.js
```diff
@@ -1299,8 +1360,122 @@ async function checkUpdates({ force = false } = {}) {
...
+    for (const id of wanted) {
+      if (reported.has(id)) continue;
+      results.push({ id, name: nameById.get(id) || null, ok: false, error: rateLimited.message });
+    }
+  }
+
+  // Prime the cache from the trees this run already holds, so the refresh that
+  // follows is a cache hit rather than a second tree call per repo. Skipped when
+  // the run derived nothing — there is no fresh verdict to record.
+  if (Object.keys(verdicts).length) writeUpdateCache(verdicts);
```
obsesivegamer, line 1490: **Partial `writeUpdateCache(verdicts)` claims a complete check.** `verdicts` is only the skills this run actually compared (skip / install / failed-with-`fresh`). Carry of everyone else is gated on `Date.now() - cached.checkedAt < 1h` at `writeUpdateCache` (`:1292–:1298`). Failed tree fetches (`:1403–:1405`) and the rate-limit backfill (`:1458–:1464`) never set `verdicts[id] = true`. This write then stamps `fingerprint: updateFingerprint(managed)` (the **whole** live library) and `checkedAt: Date.now()`, so the `loadUpdates()` that follows is a cache hit on a map that never mentioned the unexamined ids. The UI treats a missing key as not stale (`hasUpdate={Boolean(updates?.[id])}`, `staleUpdateIds` filters `stale && installed`).

The in-TTL extras test hides this because it seeds `checkedAt: Date.now()`. Independent probe, two stale skills A/C:

```
expired cache, updateSkills([A])     → checkUpdates cached {A:false}          C dropped
no cache, update-all, C tree 500     → checkUpdates cached {A:false}          C dropped
expired cache, 2nd repo 403          → checkUpdates cached {A:false}          C dropped
in-TTL cache, same three shapes      → C:true carried                         control
```

e13’s “failed run wipes the retry affordance” is back for the expired/missing-cache path: rate-limit toast says try in an hour, remaining badge and Update all are gone. `handleRefresh` calls un-forced `checkUpdates`, which hits this prime. `{A:false}` still has keys, so `hasUpdatesLoaded` prevents a tab-switch refetch. Recovery is wait out the **new** 1h TTL.

Treat “not in this `verdicts` map” as unknown, never as current. Either carry previous booleans for still-live ids even when the on-disk cache is older than TTL, or do not stamp a full-library fingerprint/`checkedAt` until every live id has a derived verdict. For requested ids that failed or were skipped by a rate-limit stop, set `verdicts[id] = true` explicitly. Add a test with an expired (or missing) cache covering “update A, C remains stale.”

obsesivegamer (commented):
## Recheck `1aca5e2f` — verdict **ISSUES**

The three findings from the `e13fd37c` recheck hold on this tip. Cache priming still has a hole that the in-TTL tests do not cover: a partial `writeUpdateCache()` stamps a **full-library** fingerprint and a fresh `checkedAt`, so the follow-up `loadUpdates()` cache-hits a map that never mentioned the skills this run did not re-check, and the UI reads a missing key as “not stale”.

Not undrafting.

### Prior three — re-proved (PASS)

Compared the tip against `e13fd37c` with a standalone probe (not the extras file). e13 still fails each case; the tip does not.

| Claim | Tip | e13 still does |
|---|---|---|
| Uninstall under the skill’s own download | skill stays gone (`updated: 0`) | resurrects `o/r:a` |
| Same-directory replacement mid-download | replacement id + dest `MARKER.txt` survive | both clobbered |
| Rate-limit reports every requested id | 3 shapes, `dropped: []`, counts reconcile, `failed` includes the stop | first-tree/first-raw drop both ids; second-repo drops `o/r2:c`; `failed` stays 0 |
| Per-row toast on 403 | empty-results and backfilled-row both `rate_limited` | empty-results toasts `updated` |

Tests on this SHA: `node --test test/skills-manager-extras.test.js` **29/29**; `npm --prefix dashboard test -- src/pages/SkillsPage.test.jsx` **11/11** (includes the three per-row toast tests); `node --test test/local-api-skills.test.js` **14/14**. HTTP probe: unauthenticated `update_all` → **401**; missing / malformed / `[]` ids → `{updated:0, rows:0}`, **zero** GitHub calls.

CI on `1aca5e2f`: 7/7 green.

### Does not hold

**Partial cache prime wipes badges this run never re-checked.** Carry of the previous map is gated on `checkedAt` still being inside the 1h TTL. Failed trees and rate-limit backfill write result rows but **not** `verdicts[id]=true`. `writeUpdateCache(verdicts)` then publishes `{fingerprint: entire live library, checkedAt: now, updates: only this run}`. `checkUpdates()` hits. `loadUpdates()` replaces React state. Missing keys are not stale.

Independent probe (two stale skills, A then C):

| Setup | `checkUpdates` after the run | `o/r2:c` still stale? |
|---|---|---|
| In-TTL cache, per-row Update of A | `{a:false, c:true}`, cached | yes (control) |
| **Expired cache (2h), per-row Update of A** | `{a:false}` only, cached | **no** |
| **No cache, Update-all, C’s tree HTTP 500** | `{a:false}` only, cached | **no** |
| In-TTL cache, Update-all, C’s tree 500 | `{a:false, c:true}`, cached | yes (control) |
| **Expired cache, Update-all, second repo 403** | `{a:false}` only, cached | **no** |

That last one is the retry-affordance bug again: the toast says try in an hour, and the remaining badge / Update-all button are gone. Refresh does not save it — `handleRefresh` → `loadUpdates()` → un-forced `checkUpdates()`, which cache-hits the prime. `hasUpdatesLoaded` is `Object.keys(updates).length > 0`, and `{a:false}` has keys, so tab switches do not refetch either. Recovery is wait out the new 1h TTL, then a load that actually misses.

The extras test `keeps the badge of a skill whose repo could not be reached` seeds `checkedAt: Date.now()`, so it only covers the control column.

### Previously accepted, still holds (spot-check)

Empty/malformed `ids` no-op, `update_all` auth, skip-only clears the badge in-TTL, success path primes in-TTL, confirm modal present, per-row goes through `updateSkills([id])`, zh/zh-TW 1596/1596 with `Skill` in Latin, fetch stubs match hostname.

### Residual, not re-opened

No registry lock (same-process window is now sync; cross-process is install/uninstall’s shape). `runMutation` still has no mutex. `getRepoTree` still retries `master` after a 403 on `main`. None of those are new on this tip.

No new deslop. `1aca5e2f` already dropped the stale `body.ids` sentence and the dead `ConfirmModal busy` prop.

Not merging, not undrafting.

obsesivegamer:
## Re: recheck of `1aca5e2f` — fixed in `9f14501c`

You were right, and the framing was the useful part: the entry is read back as a verdict for the **whole library**, so only a round that attempted the whole library may publish one. `updateSkills()` never does — it attempts the ids it was handed — yet it stamped the full fingerprint and a fresh `checkedAt` regardless. Whenever the carried map was empty, that published silence about skills nobody looked at.

Two changes in `writeUpdateCache()`:

- A round that did not attempt the library may only publish when the carried map fills in the rest. Otherwise the cache is left untouched and the next check does the work. `checkUpdates()` passes `full: true`; it is the only caller that sweeps everything.
- A prime that does publish inherits the `checkedAt` of the check it built on, so re-deriving a few ids cannot restart the hour for the rest.

Your probe table is now the test, and I ran it red against `1aca5e2f` first:

| Row | `1aca5e2f` | `9f14501c` |
|---|---|---|
| In-TTL, per-row update of A | `{a:false, c:true}` | unchanged (control) |
| **Expired cache, per-row update of A** | `{a:false}` only, cached | `{a:false, c:true}` from a real check |
| In-TTL, Update-all, C's tree 500 | `{a:false, c:true}` | unchanged (control) |
| **Expired cache, Update-all, second repo 403** | `{a:false}` only, cached | cache untouched; the check raises the limit, so the UI keeps its map and the badge |
| **No cache, Update-all, C's tree 500** | `{a:false}` only, cached | cache untouched; next check is a real miss |
| Partial prime off a 40-min-old check | `checkedAt` reset to now | inherits the older `checkedAt` |

Four of those fail on the previous tip, two are controls. A sixth test pins the optimisation the guard wraps: after a fresh check, Update all still primes and the refresh still spends **no** tree call — the guard was not allowed to buy correctness with the thing it was added for.

### One row I did not fully solve, stated plainly

**No cache, Update-all, C's tree 500.** After the fix the run no longer manufactures a snapshot, and the follow-up is a genuine miss — but `checkUpdates()` then reaches C, fails, omits it, and writes `{a:false}` under a full fingerprint because it *did* attempt the library. So C's badge is still absent.

The difference is that nothing is now hiding a verdict the system ever held: with no prior check there is no verdict for C to preserve. The test for that row asserts the contract I can actually defend — the cache is left alone and the next check is real (`cached: false`) — rather than asserting a badge I cannot honestly produce.

Making `checkUpdates()` withhold the full fingerprint whenever any repo failed would fix it, at the cost of no caching at all for anyone with one unreachable repo. That is a bigger behaviour change than this thread, so I left it. Happy to take it if you think the trade is right.

### Residuals, unchanged

No registry lock, no `runMutation` mutex, `getRepoTree` still retrying `master` after a 403 on `main` — none touched here, all as you listed them.

Suites on this tip: `test/skills-manager-extras.test.js` **36/36**, node **2751** (2749 pass / 2 skipped / 0 fail), dashboard vitest **734/734**, all validators green, CI **8/8**.

Still a draft. Not merging, not undrafting.

https://claude.ai/code/session_01TxuZak4RcjPp7WBHLSP28W

PR #614, opened 2026-09-11
fix(dsh): support DeepSeek Harness v3 session format

### LINKED ISSUE #611: [bug]: 不支持deepseek harness新的v3会话格式
### Preflight checklist

- [x] I searched [existing issues](https://github.com/xiufengsun/TokenTracker/issues) and didn't find a duplicate.
- [x] I'm on the [latest version](https://github.com/xiufengsun/TokenTracker/releases).

### What happened?

不支持deepseek harness新的v3会话格式，漏记dshv0.1.5-rc.1版本产生的会话。

Image

### Steps to reproduce

支持deepseek harness新的v3会话格式

### Expected behavior

_No response_

### Is this a regression?

None

### TokenTracker version

0.96.2

### Node.js version

_No response_

### How did you install it?

Windows installer / zip

### OS / version

win11

### Output of `tokentracker doctor` (if applicable)

```shell

```

### Anything else?

_No response_

xMinhx:
## Summary

Fixes #611. DeepSeek Harness `v0.1.5-rc.1` writes session logs using the v3 format (`session.v3.jsonl` / `session.v3.jsonl.zstd`) and updated event types (`session/start`, `message/assistant`). Because `isDshSessionLogName` checked strictly for `session.jsonl`, all v3 sessions were being skipped during sync.

## Scope

- [x] CLI (`src/`)
- [ ] Docs / CI / config

## Checklist

- [ ] Full `npm test` passes (13 unrelated baseline failures remain)
- [x] Focused rollout/DeepSeek tests pass (305/305)

## Changes

- `src/lib/rollout.js`:
  - `isDshSessionLogName`: match versioned session logs (`session.v<N>.jsonl[.zstd]`) alongside unversioned ones.
  - `parseDshVersion`: parse version suffix for tie-breaking on tied file mtimes.
  - `resolveDshSessionFiles`: sort by mtime, then higher format version, then `.zstd`.
  - `extractDshSessionUsage`: accept both `session` / `session/start` and `assistant/message` / `message/assistant`, plus fallback to `data.model` and `timestamp`.
- `test/deepseek-harness.test.js`:
  - added test cases for filename regex and version extraction.
  - added tie-breaker verification (prefers v3 over legacy when timestamps match).
  - added end-to-end integration test parsing synthetic v3 session logs.

## Verification

- `node --test test/*deepseek*.test.js test/*rollout*.test.js`: 296/296 pass.
- `npm run validate:copy && npm run validate:locale && npm run validate:guardrails && npm run validate:versions`: all pass with 0 errors.
- End-to-end simulated run on both uncompressed `session.v3.jsonl` and compressed `session.v3.jsonl.zstd`: tokens and models correctly parsed and queued.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session logs now support versioned filenames and compressed variants.
  * Newer log formats are preferred when selecting equally recent artifacts.
  * Session and assistant events are recognized across supported event types and field formats.
  * Session parsing now handles version 3 logs correctly.
  * Log migrations avoid duplicate records and preserve unchanged data.
  * Session data is rebuilt accurately when artifacts change.
  * Archived sessions remain available when log files are removed.
  * Unreadable or incomplete logs no longer block healthy sessions.
  * Sync now warns when artifact migrations are deferred and can be retried.

## Migration accounting hardening

DeepSeek Harness v2→v3 can preserve a session while rewriting artifact names and sequence semantics. The parser now reconciles migrations with per-session contribution ledgers, validates complete replacement coverage before subtraction, preserves archived/deleted history, and defers missing, unreadable, partial, or unsafe replacements instead of guessing. Incomplete JSONL records cannot advance the watermark, and queue/cursor updates remain retryable. Deferred migration reasons are surfaced for recovery.

The conservative limitation is intentional: if a pre-ledger artifact disappears before its replacement can be verified, accounting is deferred rather than risking double-counting or data loss.


xiufengsun (changes requested):
Re-reviewed 86bc97981a100dbb888da6c92d2d004f5bb25c63 against current main. Filename discovery handles v3, but switching an already-counted session to the new artifact doubles its usage.

Minimal reproduction: parse session.jsonl with session id `same-session` and one assistant usage event (100 input + 20 output); rename it to session.v3.jsonl with the same header/event; run parseDshIncremental again. The latest queue bucket changes from 120 to 240 tokens and conversation_count from 1 to 2. `fileState` is keyed by pathname, so the v3 artifact has no prior watermark and the old cursor is deleted afterward. This is an upgrade path introduced by accepting the versioned artifact.

Please preserve accounting across legacy/v3 and plain/zstd artifact transitions, including coexistence and repeated sync. Do not blindly transfer a sequence watermark: the upstream v2-to-v3 migration rewrites sequence numbers (see `packages/session/session-format-v2-to-v3/src/migration.ts`, upstream c291e7961a515f6d7af9304e7fd1d257929aef26). Use stable event identity or verified contribution reconciliation and add regression coverage. Issue #611 stays open until the fixed version is released.

xMinhx:
## DeepSeek Harness v3 migration accounting hardening

The original v3 support handled filename discovery and parsing, but an already-counted legacy session could be switched to `session.v3.jsonl`/`.zstd` with different sequence numbers. Treating the replacement as a normal append could double-count usage; rebuilding all DSH buckets could also erase archived sessions or unrelated sessions sharing the same bucket. Missing, partial, or header-only replacement files could likewise retract valid history, while incomplete trailing JSONL records could advance the cursor and lose usage permanently.

The pushed fix is deliberately fail-closed and per-session:

- Tracks each session's bucket contribution ledger and reconciles only the affected session.
- Reconstructs legacy baselines only from the previously accounted prefix; it never assumes the current legacy file was already counted.
- Accepts replacement subtraction only after stable session identity, complete parsing, prior-contribution coverage, and bucket-coverage checks.
- Defers missing, unreadable, incomplete, empty, or unsafe migrations without deleting the old cursor identity.
- Detects same-path resets/sequence rewrites and prevents malformed or partial records from advancing `lastSeq`.
- Keeps queue/cursor updates transactional and preserves archived/deleted-session totals.
- Exposes deferred migration reasons for retry and bounds legacy tombstones to prevent unbounded cursor growth.

Validation on head `7d99abeb`:

- Focused rollout/DeepSeek tests: **305/305 passed**
- CI, CodeQL, Linux/macOS/Windows checks: **passed**
- Guardrails, locale, version, and copy validators: **passed**
- Full suite: **2720 passed, 13 unrelated existing failures, 8 skipped**

A ledger-less migration whose old artifact disappeared before the replacement can be verified remains deferred rather than guessed; this is intentional to prevent silent double-counting or data loss.

PR #628, opened 2026-09-13
docs: record the skills registry mutation lessons from #613

obsesivegamer:
## What

Adds a `### Skills registry mutations` entry to the "Lessons learned" section, recording why the two follow-up commits on #613 were needed.

## Why

Both bugs you fixed on top of that PR came from one assumption: that a snapshot taken *before* a multi-minute download is still current when the write happens.

`updateSkills()` hands `installSkill()` a registry entry read before the downloads start. #613 added a guard that re-reads the registry at the last synchronous point and checks the skill still exists — but kept taking `targets` from the stale copy. So toggling an agent while a skill downloaded silently reverted the toggle. `9c732818` fixed it by reading both from the same fresh `readRegistry()`.

The second bullet covers the temp-dir naming (`${name}-${Date.now()}` collides at millisecond resolution once mutations overlap, and the old `removePath(temp)` then deleted the other run's in-flight download) and the `operationInFlight` ref in `SkillsPage.jsx` — `busyKey` state commits too late to block a fast second click. Both only became reachable once bulk updates made mutations overlap, which is the part worth writing down: "pre-existing" stopped being true the moment the window went from milliseconds to minutes.

## Checks

Docs only — no version bump or release per the release-scope table. Every claim verified against the merged code on `main` (`mkdtempSync` in `skills-manager.js`, `operationInFlight` in `SkillsPage.jsx`, `current.targets` at `skills-manager.js:982`) rather than written from memory.

Branched off `main` at `7f3d9bc4` (0.96.4).

## Note

Opened as a draft — your call whether this belongs in the shared file at all, or whether it is too PR-specific for a doc that loads on every turn. Happy to trim it to one bullet or drop it.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for safely updating the skills registry.
  - Documented requirements for fresh registry reads during installation.
  - Clarified temporary directory creation and operation-serialization considerations.


xiufengsun:
Taking it. To answer your question: yes, it belongs in the file. The first bullet is the general lesson (a snapshot taken before a multi-minute await is not current at the write), and it is not specific to skills; the second one is concrete enough to act on.

I pushed a commit that tightens both to about half the length and adds the line anchors, since that is the form the rest of that section is in. Content unchanged.

One note on your verification method, because it matters for anything you check in this repo again: `src/lib/skills-manager.js` is classified as binary by `file(1)`, so a plain `grep` silently skips it and reports nothing. I nearly concluded your references were wrong before remembering the repo has this quirk (`rollout.js` has it too). With `grep -a` everything you cited is there, `current.targets` at 982 exactly as you said. Use `grep -a` on those two files.

Thanks for writing this up rather than letting the reasoning evaporate after the fix.

PR #636, opened 2026-09-16
docs: record why grep silently misses skills-manager.js

obsesivegamer:
## What

One bullet under `### Skills registry mutations` recording that `grep` skips `src/lib/skills-manager.js`.

## Why

You flagged this on #628 — plain `grep` on that file finds nothing and you nearly concluded my line references were wrong. I went looking for the cause rather than just writing down the workaround.

`hashDirectory()` writes three literal NUL bytes as field separators:

- `:493` — the two separators inside `hash.update(...)` that look like spaces are NUL
- `:499` — the trailing `hash.update(" ")` is a NUL, not a space

That is enough for `grep` to classify the file as binary. It prints nothing and **exits 1**, which is indistinguishable from "no match" — it never reports that it skipped anything. Hence the wrong conclusion.

Two corrections to what we both assumed:

- **`rollout.js` does not have this.** It reads as plain UTF-8. I scanned every tracked file: `skills-manager.js` is the only text file in the repo that `grep` treats as binary, so the caveat is scoped to it rather than written as a repo-wide search rule.
- It is a property of those three bytes, not of the repo or of `file(1)`'s opinion.

## The option I did not take

Writing those three bytes as JS escapes for U+0000 instead of raw bytes is byte-identical at runtime (verified — every hash stays the same) and would remove the trap for grep, ripgrep, editors and diff viewers alike. I left it alone because a zero-behaviour change under `src/` triggers the full npm + DMG + Windows + Linux release per the scope table, which seemed a poor trade for a source-encoding tidy. Worth doing as a passenger on the next `src/` change if you agree — say the word and I will open it.

## Checks

Docs only — no version bump or release per the release-scope table. Line numbers verified against `main`. Branched off the #628 merge commit (`779454a1`) rather than `main`'s head so the push carries no workflow-file changes.


xiufengsun:
Good find on the cause, and thanks for chasing it rather than just writing down the workaround. But one of the two "corrections" is itself wrong, and it is the load-bearing one.

`src/lib/rollout.js` **does** have this property:

```
$ grep -c "function " src/lib/rollout.js
                     # (empty, exit 1)
$ grep -a -c "function " src/lib/rollout.js
515
$ python3 -c "print(open('src/lib/rollout.js','rb').read().count(b'\x00'))"
1
```

One NUL, at line 3468 — the `.join(...)` in the dedup-key builder uses a raw NUL byte, the same way `hashDirectory()` does. `file(1)` reports `Unicode text, UTF-8 text` for that file, which is presumably what you checked; it only sniffs a prefix and the NUL is 122KB in. grep reads the whole thing and skips it.

I scanned every tracked file by byte rather than by `file(1)`. Exactly two are affected:

```
   1 NUL     819184B  src/lib/rollout.js
   3 NUL      59306B  src/lib/skills-manager.js
```

Which matters here, because `rollout.js` is the file people grep constantly — it is the one CLAUDE.md sends you to for every provider parser. A bullet that says "it is the only tracked text file in the repo with this property", filed under **Skills registry mutations**, would leave the more common trap undocumented while actively asserting it does not exist.

Could you rescope it? What I would take:

- move it out of `### Skills registry mutations`, since it is not about skills
- name both files
- drop the "only file" claim, or state it as "these two"

On the option you did not take: yes, do it. Replacing the raw bytes with escape sequences for U+0000 in both files is the right fix — a caveat in CLAUDE.md is a workaround, and this one has already cost us each a wrong conclusion. The release-scope argument does not really apply: a `src/` change ships with the next release either way, and there is a `src/` PR open right now.

One thing to verify before you do: the hashes must be byte-identical before and after. You already checked that for `skills-manager.js`. `rollout.js` needs the same check against real queue data, because that join feeds a dedup key — if the key changes, every reader re-counts history it already had.

Separate PR is fine, and if it lands this doc bullet may not be needed at all.

xiufengsun:
Fixed at the source in 21dee115: the NUL bytes in `rollout.js` and `skills-manager.js` are now `\u0000` escapes, so plain `grep` reads both files as text and this caveat is no longer needed. The escaped literals produce the same runtime strings; I checked `hashDirectory` on a real skills tree before and after and the hashes match.

Closing this one. Thanks for tracking down the cause instead of just documenting the workaround.

